PR #23, opened 2022-03-22
Api comparision

toddbaert:
Adding overview of some major APIs/SDKs. See the file in the branch for the markdown preview,:

https://github.com/open-feature/spec/blob/api-comparision/research/api-comparision.md

### DIFF
--- a/research/api-comparision.md
+++ b/research/api-comparision.md
@@ -1,0 +1,298 @@
+# API Comparison
+
+The goal of this research is to make it easy to quickly compare feature flag
+vendor API's in order to help define the OpenFeature spec.
+
+## Boolean
+
+<details>
+  <summary>Java</summary>
+
+```java
+/**
+* Unleash
+*
+* SDK Repo: https://github.com/Unleash/unleash-client-java
+*/
+boolean isEnabled(String toggleName)
+boolean isEnabled(String toggleName, UnleashContext context)
+boolean isEnabled(String toggleName, UnleashContext context, boolean defaultSetting)
+boolean isEnabled(String toggleName, BiFunction<String, UnleashContext, Boolean> fallbackAction)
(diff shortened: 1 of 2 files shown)


research/api-comparision.md
```diff
@@ -0,0 +1,619 @@
...
+
+</details>
+
+### Integers and Floats
+
+<details>
+  <summary>Java</summary>
+
+```java
+// TODO: confirm this.
```
beeme1mr, line 346: Resolve TODO

research/api-comparision.md
```diff
@@ -0,0 +1,619 @@
...
+  titleSize: new RoxNumber(12, [12, 14, 18, 24])
+};
+
+Rox.register('test-namespace', flags);
+flags.titleSize.value();
+
+export class RoxNumber {
+  getValue(context?: unknown): number;
+}
+
```
beeme1mr, line 479: I couldn't find any reference to dynamic APIs in Node. Perhaps it's worth confirming and adding a note here since it's available in Java.

research/api-comparision.md
```diff
@@ -0,0 +1,618 @@
...
+*/
+const flags = {
+  videoChat: new Rox.Flag()
+};
+
+Rox.register('test-namespace', flags);
+flags.videoChat.isEnabled()
+flags.videoChat.isEnabled(context)
+
+export class Flag {
```
beeme1mr, line 158: How is this class used?

toddbaert: That's just the instance that's created when you define a boolean flag object. It's just the class that houses the `getValue()` method.

toddbaert: I've removed the class context, it's not needed, the Java sections only have method signatures... and if the classes aren't needed in the Java samples( the "classiest" language there is) we don't need them in TS.

beeme1mr (approved):
Looks good! Thanks for putting this together.

PR #26, opened 2022-03-24
Clarify API vs SDK

justinabrahms:
I'm not sure if others will agree that this makes it clearer, but this is my mental model of how this works.

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -45,13 +45,13 @@ The maintainer of a shared library which is a dependency of many applications or
 
 ### Feature Flag SDK
 
-The libraries used by Application Author to implement feature flags in their application or service. The interface of these libraries adhere to the Feature Flag API.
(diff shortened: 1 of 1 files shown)


toddbaert (approved):

toddbaert:
Makes sense to me.

glossary.md
```diff
@@ -45,11 +45,14 @@ The maintainer of a shared library which is a dependency of many applications or
 
 ### Feature Flag SDK
 
-The interfaces and abstractions used by Application Author to implement feature flags in their application or service.
+The libraries used by Application Author to implement feature flags in their application or service. The interface of these libraries adhere to the Feature Flag API.
```
toddbaert, line 48: I like this point. I think it hits on the fact that that SDK is really a library that contains interfaces and abstractions, whereas the

justinabrahms:
Applied both suggestions.

toddbaert (approved):

PR #28, opened 2022-03-25
Add Application Integrator to User Roles in glossary

moredip:
I'm not a fan of "Application Integrator" as a label, but couldn't think of a better term. Suggestions welcome!

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -31,7 +31,7 @@ A developer of an application or service which utilizes the feature flags SDK. T
 
 ### Application Integrator
 
-A developer who is setting up or configuring an application or service to use the feature flags SDK.
+A developer who is setting up or configuring an application or service to use the feature flags SDK. They would write code like "We should speak to the open source flagging service, not $vendor" or "The way the system should handle telemetry is through $library".
 
 ### Provider Author
 
--- a/glossary.md
+++ b/glossary.md
@@ -27,7 +27,11 @@ This document defines some terms that are used across this specification.
 
 ### Application Author
 
-The maintainer of an application or service which utilizes the feature flags SDK to manage functionality.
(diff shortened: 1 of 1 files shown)


toddbaert:
Thanks @moredip ! I'm having trouble differentiating between the "Application Author" and "Application Integrator". Can you clarify for me the differences?

toddbaert:
I see. I think that's a worthwhile differentiation, and the differentiation might even become more important as our API/SDK becomes more well-defined.

toddbaert (approved):

glossary.md
```diff
@@ -27,7 +27,11 @@ This document defines some terms that are used across this specification.
 
 ### Application Author
 
-The maintainer of an application or service which utilizes the feature flags SDK to manage functionality.
+A developer of an application or service which utilizes the feature flags SDK. This person writes code which calls into the SDK to make flagging decisions.
+
+### Application Integrator
+
+A developer who is setting up or configuring an application or service to use the feature flags SDK.
```
justinabrahms, line 34:
```suggestion
A developer who is setting up or configuring an application or service to use the feature flags SDK. They would write code like "We should speak to the open source flagging service, not $vendor" or "The way the system should handle telemetry is through $library".
```

justinabrahms (approved):
This is just a rephrasing of the thing you said in the main convo thread, but would be nice to have in docs too. :)

beeme1mr:
What do you think about "OpenFeature Integrator" instead of "Application Integrator"? It seems like we're focusing on the persona of the user that actually configures the provider and any global settings.

toddbaert:
I'm going to merge this as it is for now, since it seems we agree it's a useful distinction. We can tweak language whenever.

Issue #32, opened 2022-03-31
"reasoning" concept for a flag evaluation


justinabrahms:
When we evaluate a flag, we hand over a pile of context and a key name. In response, we get a value. For many reasons (logging, debugging, etc), we might like to know _why_ we got that value. The spec should provide an API that will allow us to introspect this stuff.

Key info, in my mind, is:
- was there an error? What [kind](https://launchdarkly.github.io/java-server-sdk/com/launchdarkly/sdk/EvaluationReason.ErrorKind.html)?
- is this the default I provided in my API call or something from the evaluator implementation?

Maybe also:
- which rule lead to this value being chosen?

toddbaert:
I think we can agree this is a good idea. It's been brought up in a number of threads and in community meetings. Of course, it seems like our ability to surface this data is highly dependent on what is exposed by the underlying provider, but I think defining another set of methods that wrap the flag-value along with some additional evaluation metadata is probably a good approach. This isn't unlike the pattern LD uses, and something like what split does to surface the additional configuration data associated with a flag.

The trouble is that I don't think all the providers have this class of functionality, and the majority also don't seem to throw exceptions/errors (a perfectly understandable choice in the case of feature flags) further obscuring things.

I'll take a research action item from this, and I'd also appreciate input from anyone familiar with how we might get such information from the various provider SDKs.

beeme1mr:
This may be a situation where we have to determine what information would be the most valuable and leave it up to the providers to handle. We would then have to handle the situation where this information is not provided but I think that's a reasonable compromise.

toddbaert:
Relevant PR to this discussion: https://github.com/open-feature/sdk-research/pull/22

toddbaert:
We've had some interesting discussion about this, see thread [here](https://github.com/open-feature/sdk-research/pull/22#discussion_r849000070).

In summary, I think we need to decide whether or not we want OpenFeature to rigorously define a set of reasons and force providers to map their own reasons onto that, or allow reasons to be more free-form strings. The clincher here, as @beeme1mr points out, is really about whether we expect Application Authors or Integrators to write logic based on these values, or otherwise, they're only important for diagnostics/telemetry/troubleshooting... If we think code will be written around these reason values, I think we should define them well within OpenFeature, and accept that there will be some loss of fidelity between the provider's "reason" and what's exposed in OpenFeature.

@justinabrahms since you opened this discussion, I'm particularly interested in your take.

justinabrahms:
[...]

PR #35, opened 2022-04-21
Add flag-evaluation spec, Makefile, initial spec structure

toddbaert:
First pass at a partial spec, based on cumulative findings from our [research repo](https://github.com/open-feature/sdk-research). Only the `Evaluation API` is proposed (`specification/flag-evaluation/flag-evaluation.md`, also note the associated assertion JSON file: `specification/flag-evaluation/flag-evaluation.json`). The best way to view the actual spec is probably to view the branch directly, since it's mostly markdown: https://github.com/open-feature/spec/tree/spec/flag-evaluation-api/specification

One of the goals was to design the structure of the spec, so I've included stubs for some of the other parts.

Please also checkout the `specification/README.md`, which contains an overview, conformance clause, and document statuses. 

For ease of generating the JSON assertion files, I've also added a very simple Makefile. If you're interested in looking at spec only, please make use of the Github's file-type filter: 
image

UPDATE: 

I believe I've addressed all the concerns. I've also opened: https://github.com/open-feature/spec/issues/36 in response to comments about CI and various helpful automation suggestions. Please remember, everything in the spec at the moment is still `status: experimental`, so don't think that anything here is final!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# OpenFeature specification (draft)
+# OpenFeature Specification (Draft)
 
 [![Roadmap](https://img.shields.io/static/v1?label=Roadmap&message=public&color=green)](https://github.com/orgs/open-feature/projects/1)
 [![Contributing](https://img.shields.io/static/v1?label=Contributing&message=guide&color=blue)](https://github.com/open-feature/.github/blob/main/CONTRIBUTING.md)
@@ -13,7 +13,7 @@ It also contains research that was considered while defining the spec.
 > Also see the ongoing research on SDK/APIs [here](./research/existing-landscape.md)
 > See the [glossary](./glossary.md) for definitions of key terminology.
 
-## Design principles
+## Design Principles
 
 - Compatibility with existing open source and commercial feature flag offerings
 - Simple, understandable API
@@ -21,7 +21,7 @@ It also contains research that was considered while defining the spec.
 - First class Kubernetes support
 - Native support for OpenTelemetry
 
-### SDKs and client libraries
+### SDKs and Client Libraries
 
 The project aims to provide a unified API and SDK for feature flag management in
 various technology stacks. The flag evaluation logic will **not** be handled in
@@ -32,6 +32,25 @@ external evaluation engine in a vendor agnostic way.
 
 The OpenFeature project will include client libraries for common technology stacks including, but not limited to:
 
-* Golang
-* Java
-* JavaScript/TypeScript
+- Golang
+- Java
+- JavaScript/TypeScript (Node.js)
+
+### Tooling
+
+This specification complies with [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) and seeks to conform to the [W3C QA Framework Guildlines](https://www.w3.org/TR/qaframe-spec/).
+
+In accordance with this, some basic tooling (donated graciously by [Diego Hurtado](https://github.com/ocelotl)) has been employed to parse the specification and output a JSON structure of concise requirements, highlighting the particular `RFC 2119` verb in question.
+
+To parse the specification, simply type `make`. Please review the generated JSON files, which will appear as siblings to any of the markdown files in the `/specification` folder.
+
+### Style Guide
+
+- Use code blocks for examples.
+  - Code blocks should be pseudocode, not any particular language, but should be vaguely "Java-esque".
+- Use conditional requirements for requirements that only apply in particular situations, such as particular languages or runtimes.
+- Use "sentence case" enclosed in ticks (\`) when identifying entities outside of code blocks (ie: `evaluation details` instead of `EvaluationDetails`).
+- Do not place line breaks into sentences, keep sentences to a single line for easier review.
+- String literals appearing outside of code blocks should be enclosed in both ticks (\`) and double-quotes (") (ie: `"PARSE_ERROR"`).
+- Use "Title Case" for all titles.
+- Use the imperative mood and passive voice.
--- a/specification/README.md
+++ b/specification/README.md
@@ -1,0 +1,34 @@
+OpenFeature Specification
+
+## Contents
+
+- [Glossary](./glossary.md)
+- [Types](./types.md)
+- [Design Principles](design-principles.md)
+- [Evaluation API](./flag-evaluation/flag-evaluation.md)
+- [Hooks](./flag-evaluation/hooks.md)
+- [Providers](./provider/providers.md)
+
+## Conformance
+
+### Normative Sections
+
+The following parts of this document are normative:
+
+- Statements under markdown H5 headings, appearing in markdown block quotes, and containing an uppercase keyword from RFC 2119.
+- This conformance clause.
+
+### Conformance Requirements and Test Assertions
+
+Each [normative section](#normative-sections) defines a single requirement. By enumerating these normative sections, a list of test assertions can be derived.
+
+### Compliance
+
+An implementation is not compliant if it fails to satisfy one or more of the "MUST", "MUST NOT", "REQUIRED", "SHALL", or "SHALL NOT" requirements defined in the [normative sections](#normative-sections) of the specification. Conversely, an implementation of the specification is compliant if it satisfies all the "MUST", "MUST NOT", "REQUIRED", "SHALL", and "SHALL NOT" requirements defined in the [normative sections](#normative-sections) of the specification.
+
+## Document Statuses
+
+| Status               | Explanation                   |
+| -------------------- | ----------------------------- |
+| No Explicit "Status" | Equivalent to Experimental.   |
+| Experimental         | Breaking changes are allowed. |
--- a/specification/design-principles.md
+++ b/specification/design-principles.md
@@ -1,0 +1,5 @@
+# Design Principles
+
+## Overview
+
+// TODO: this is a stub.
--- a/specification/flag-evaluation/flag-evaluation.json
+++ b/specification/flag-evaluation/flag-evaluation.json
@@ -1,0 +1,146 @@
+[
+    {
+        "id": "1.1",
+        "clean id": "1_1",
+        "content": "The `API`, and any state it maintains SHOULD exist as a global singleton, even in cases wherein multiple versions of the `API` are present at runtime.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.2",
+        "clean id": "1_2",
+        "content": "The `API` MUST provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.3",
+        "clean id": "1_3",
+        "content": "The `API` MUST provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.4",
+        "clean id": "1_4",
+        "content": "The API MUST provide a function for retrieving the `provider` implementation.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.5",
+        "clean id": "1_5",
+        "content": "The `API` MUST provide a function for creating a `client` which accepts the following options: - name (optional): A logical string identifier for the client.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.6",
+        "clean id": "1_6",
+        "content": "The client MUST provide a method to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.7",
+        "clean id": "1_7",
+        "content": "The `client` MUST provide methods for flag evaluation, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.9",
+        "clean id": "1_9",
+        "content": "The `client` MUST provide methods for detailed flag value evaluation with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns an `evaluation details` structure.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.10",
+        "clean id": "1_10",
+        "content": "The `evaluation details` structure's `value` field MUST contain the evaluated flag value.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.12",
+        "clean id": "1_12",
+        "content": "The `evaluation details` structure's `flag key` field MUST contain the `flag key` argument passed to the detailed flag evaluation method.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.13",
+        "clean id": "1_13",
+        "content": "The `evaluation details` structure's `variant` field SHOULD contain a identifier logically corresponding to the return flag value.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.14",
+        "clean id": "1_14",
+        "content": "The `evaluation details` structure's `reason` field SHOULD indicate the semantic reason for the returned flag value. Possible values vary by provider, but may include `\"TARGETING_MATCH\"`, `\"SPLIT\"`, `\"DISABLED\"`, `\"DEFAULT\"`, `\"UNKNOWN\"` and `\"ERROR\"`.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.15",
+        "clean id": "1_15",
+        "content": "In cases of normal execution, the `evaluation details` structure's `error code` field MUST not be set, or otherwise must contain a null or falsy value.",
+        "RFC 2119 keyword": "MUST"
+    },
+    {
+        "id": "1.16",
+        "clean id": "1_16",
+        "content": "In cases of abnormal execution, the `evaluation details` structure's `error code` field SHOULD identify an error occurred during flag evaluation, with possible values `\"PROVIDER_NOT_READY\"`, `\"FLAG_NOT_FOUND\"`, `\"PARSE_ERROR\"`, `\"TYPE_MISMATCH\"`, or `\"GENERAL\"`.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.17",
+        "clean id": "1_17",
+        "content": "In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` SHOULD indicate an error.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.18",
+        "clean id": "1_18",
+        "content": "The `evaluation options` structure's `hooks` field denotes a collection of hooks to be executed for the respective flag evaluation.",
+        "RFC 2119 keyword": null
+    },
+    {
+        "id": "1.19",
+        "clean id": "1_19",
+        "content": "No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.",
+        "RFC 2119 keyword": null
+    },
+    {
+        "id": "1.20",
+        "clean id": "1_20",
+        "content": "In the case of abnormal execution, the client SHOULD log an informative error message.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.21",
+        "clean id": "1_21",
+        "content": "The `client` SHOULD provide asynchronous or non-blocking mechanisms for flag evaluation.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "Condition 1.8",
+        "content": "The language type system differentiates between strings, numbers, booleans and structures.",
+        "children": [
+            {
+                "id": "1.8.1",
+                "clean id": "1_8_1",
+                "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure.```// example boolean flag evaluationboolean myBool =  client.getBooleanValue('bool-flag', false);// example overloaded string flag evaluation with optional paramsstring myString = client.getStringValue('string-flag', 'N/A', evaluationContext, options);// example number flag evaluationnumber myNumber = client.getNumberValue('number-flag', 75);// example overloaded structure flag evaluation with optional paramsMyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);```See [evaluation context](../evaluation-context/evaluation-context.md) for details.",
+                "RFC 2119 keyword": "MUST"
+            },
+            {
+                "id": "1.8.2",
+                "clean id": "1_8_2",
+                "content": "The `client` SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.",
+                "RFC 2119 keyword": "SHOULD"
+            }
+        ]
+    },
+    {
+        "id": "Condition 1.11",
+        "content": "The language supports generics (or an equivalent feature).",
+        "children": [
+            {
+                "id": "1.11.1",
+                "clean id": "1_11_1",
+                "content": "The `evaluation details` structure SHOULD accept a generic argument (or use an equivalent language feature) which indicates the type of the wrapped `value` field.",
+                "RFC 2119 keyword": "SHOULD"
+            }
+        ]
+    }
+]
--- a/specification/flag-evaluation/flag-evaluation.md
+++ b/specification/flag-evaluation/flag-evaluation.md
@@ -1,0 +1,194 @@
+# Flag Evaluation API
+
+**Status**: [Experimental](../README.md#document-statuses)
+
+## Overview
+
+The `evaluation API` allows for the evaluation of feature flag values, independent of any flag control plane or vendor. In the absence of a [provider](../provider/providers.md) the `evaluation API` uses the "No-op provider", which simply returns the supplied default flag value.
+
+#### API Initialization and Configuration
+
+##### Requirement 1.1
+
+> The `API`, and any state it maintains **SHOULD** exist as a global singleton, even in cases wherein multiple versions of the `API` are present at runtime.
+
+It's important that multiple instances of the `API` not be active, so that state stored therein, such as the registered `provider`, static global `evaluation context`, and globally configured `hooks` allow the `API` to behave predictably. This can be difficult in some runtimes or languages, but implementors should make their best effort to ensure that only a single instance of the `API` is used.
+
+##### Requirement 1.2
+
+> The `API` **MUST** provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
+
+```
+// example provider mutator
+OpenFeature.setProvider(new MyProvider());
+```
+
+See [provider](../provider//providers.md) for details.
+
+##### Requirement 1.3
+
+> The `API` **MUST** provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.
+
+```
+// example hook attachment
+OpenFeature.addHooks([new MyHook()]);
+```
+
+See [hooks](./hooks.md) for details.
+
+##### Requirement 1.4
+
+> The API **MUST** provide a function for retrieving the `provider` implementation.
+
+```
+// example provider accessor
+OpenFeature.getProvider();
+```
+
+See [provider](../provider/provider.md) for details.
+
+##### Requirement 1.5
+
+> The `API` **MUST** provide a function for creating a `client` which accepts the following options:
+>
+> - name (optional): A logical string identifier for the client.
+
+```
+// example client creation and retrieval
+OpenFeature.getClient({
+  name: 'my-openfeature-client'
+});
+```
+
+The name is a logical identifier for the client.
+
+#### Client Usage
+
+##### Requirement 1.6
+
+> The client **MUST** provide a method to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.
+
+```
+// example hook attachment
+client.addHooks([new MyHook()]);
+```
+
+See [hooks](./hooks.md) for details.
+
+#### Flag Evaluation
+
+##### Requirement 1.7
+
+> The `client` **MUST** provide methods for flag evaluation, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.
+
+```
+// example flag evaluation
+var myValue = client.getValue('my-flag', false);
+```
+
+##### Condition 1.8
+
+> The language type system differentiates between strings, numbers, booleans and structures.
+>
+> ##### Conditional Requirement 1.8.1
+>
+> > The `client` **MUST** provide methods for typed flag evaluation, including boolean, numeric, string, and structure.
+>
+> > ```
+> > // example boolean flag evaluation
+> > boolean myBool =  client.getBooleanValue('bool-flag', false);
+> >
+> > // example overloaded string flag evaluation with optional params
+> > string myString = client.getStringValue('string-flag', 'N/A', evaluationContext, options);
+> >
+> > // example number flag evaluation
+> > number myNumber = client.getNumberValue('number-flag', 75);
+> >
+> > // example overloaded structure flag evaluation with optional params
+> > MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);
+> > ```
+>
+> > See [evaluation context](../evaluation-context/evaluation-context.md) for details.
+>
+> ##### Conditional Requirement 1.8.2
+>
+> > The `client` **SHOULD** guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.
+
+##### Requirement 1.9
+
+> The `client` **MUST** provide methods for detailed flag value evaluation with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns an `evaluation details` structure.
+
+```
+// example detailed boolean flag evaluation
+FlagEvaluationDetails<boolean> myBoolDetails = client.getBooleanDetails('bool-flag', false);
+
+// example detailed string flag evaluation
+FlagEvaluationDetails<string> myStringDetails = client.getStringDetails('string-flag', 'N/A', evaluationContext, options);
+
+// example detailed number flag evaluation
+FlagEvaluationDetails<number> myNumberDetails = client.getNumberDetails('number-flag', 75);
+
+// example detailed structure flag evaluation
+FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);
+
+```
+
+##### Requirement 1.10
+
+> The `evaluation details` structure's `value` field **MUST** contain the evaluated flag value.
+
+##### Condition 1.11
+
+> The language supports generics (or an equivalent feature).
+>
+> ##### Conditional Requirement 1.11.1
+>
+> > The `evaluation details` structure **SHOULD** accept a generic argument (or use an equivalent language feature) which indicates the type of the wrapped `value` field.
+
+##### Requirement 1.12
+
+> The `evaluation details` structure's `flag key` field **MUST** contain the `flag key` argument passed to the detailed flag evaluation method.
+
+##### Requirement 1.13
+
+> The `evaluation details` structure's `variant` field **SHOULD** contain a identifier logically corresponding to the return flag value.
+
+For example, the flag value might be `3.14159265359`, and the variant field's value might be `"pi"`.
+
+##### Requirement 1.14
+
+> The `evaluation details` structure's `reason` field **SHOULD** indicate the semantic reason for the returned flag value. Possible values vary by provider, but may include `"TARGETING_MATCH"`, `"SPLIT"`, `"DISABLED"`, `"DEFAULT"`, `"UNKNOWN"` and `"ERROR"`.
+
+##### Requirement 1.15
+
+> In cases of normal execution, the `evaluation details` structure's `error code` field **MUST** not be set, or otherwise must contain a null or falsy value.
+
+##### Requirement 1.16
+
+> In cases of abnormal execution, the `evaluation details` structure's `error code` field **SHOULD** identify an error occurred during flag evaluation, with possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+
+##### Requirement 1.17
+
+> In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` **SHOULD** indicate an error.
+
+##### Requirement 1.18
+
+> The `evaluation options` structure's `hooks` field denotes a collection of hooks to be executed for the respective flag evaluation.
+
+See [hooks](./hooks.md) for details.
+
+##### Requirement 1.19
+
+> No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.
+
+##### Requirement 1.20
+
+> In the case of abnormal execution, the client **SHOULD** log an informative error message.
+
+Implementations may define a standard logging interface that can be supplied as an optional argument to the client creation function, which may wrap standard logging functionality of the implementation language.
+
+##### Requirement 1.21
+
+> The `client` **SHOULD** provide asynchronous or non-blocking mechanisms for flag evaluation.
+
+It's recommended to provide non-blocking mechanisms for flag evaluation, particularly in languages or environments wherein there's a single thread of execution.
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -58,6 +58,10 @@ The interfaces and abstractions used by authors (Application, Integration, Provi
 Provider & Integration authors adhere to the API to add support for their feature flag implementation or integration.
 Application authors use it via the Feature Flag SDK.
 
+### Evaluation API
+
+The subset of the [Feature Flag API](#feature-flag-api) that the Application Author uses to evaluate flags.
+
 ### Provider
 
 An SDK-compliant feature flag implementation which adheres to the Feature Flag API. Implementations may include Saas feature flag vendors, custom "in-house" feature flag infrastructure, or open-source implementations.
@@ -86,7 +90,7 @@ Flags represent a single pivot point of logic. Flags have a type, like `string`,
 
 ### Variant
 
-A variant is a space-constrained identifier for values. This allows us to reference these values without having to include all of the data (which may be quite large). We may use some hash of the values themselves or a string name. So for the `header-order` flag, we may have variants like `reverse`, `wonky` or `standard`.
+A variant is a semantic identifier for a value. This allows for referral to particular values without necessarily including the value itself, which may be quite prohibitively large or otherwise unsuitable in some cases.
 
 ### Values
 
--- a/specification/types.md
+++ b/specification/types.md
@@ -1,0 +1,37 @@
+# Types and Data Structures
+
+## Overview
+
+This document outlines some of the common types and data structures defined by OpenFeature and referenced elsewhere in this specification.
+
+### boolean
+
+A logical true or false, as represented idiomatically in the implementation languages.
+
+### string
+
+A UTF-8 encoded string.
+
+### number
+
+A numeric value of unspecified type or size. Implementation languages may further differentiate between integers, floating point numbers, and other specific numeric types and provide functionality as idioms dictate.
+
+### structure
+
+Structured data, presented however is idiomatic in the implementation language, such as JSON or YAML.
+
+### evaluation details
+
+A structure containing the following fields:
+
+- flag key (string, required)
+- value (boolean | string | number | structure, required)
+- error code (string, optional)
+- reason (string, optional)
+- variant (string, optional)
+
+### evaluation options
+
+A structure containing the following fields:
+
+- hooks (one or more [hooks](./flag-evaluation/hooks.md), optional)
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,15 @@
+IS_PYTHON_INSTALLED = $(shell which python >> /dev/null 2>&1; echo $$?)
+
+parse: clean _check_python
+	@python ./tools/specification_parser/specification_parser.py
+
+clean:
+	@find ./specification -name '*.json' -delete
+
+_check_python:
+	@if [ $(IS_PYTHON_INSTALLED) -eq 1 ]; \
+		then echo "" \
+		&& echo "ERROR: python must be available on PATH." \
+		&& echo "" \
+		&& exit 1; \
+		fi;
--- a/specification/evaluation-context/evaluation-context.md
+++ b/specification/evaluation-context/evaluation-context.md
@@ -1,0 +1,5 @@
+# Evaluation Context
+
+## Overview
+
+// TODO: this is a stub.
--- a/specification/flag-evaluation/hooks.md
+++ b/specification/flag-evaluation/hooks.md
@@ -1,0 +1,5 @@
+# Hooks
+
+## Overview
+
+// TODO: this is a stub.
--- a/specification/provider/providers.md
+++ b/specification/provider/providers.md
@@ -1,0 +1,5 @@
+# Providers
+
+## Overview
+
+// TODO: this is a stub.
--- a/tools/specification_parser/specification_parser.py
+++ b/tools/specification_parser/specification_parser.py
@@ -1,0 +1,205 @@
+from re import (
+    finditer, findall, compile as compile_, DOTALL, sub, match, search
+)
+from json import dumps
+from os.path import curdir, abspath, join, splitext
+from os import walk
+
+rfc_2119_keywords_regexes = [
+    r"MUST",
+    r"REQUIRED",
+    r"SHALL",
+    r"MUST NOT",
+    r"SHALL NOT",
+    r"SHOULD",
+    r"RECOMMENDED",
+    r"SHOULD NOT",
+    r"NOT RECOMMENDED",
+    r"MAY",
+    r"OPTIONAL",
+]
+
+
+def find_markdown_file_paths(root):
+    markdown_file_paths = []
+
+    for root_path, _, file_paths, in walk(root):
+        for file_path in file_paths:
+
+            absolute_file_path = join(root_path, file_path)
+
+            _, file_extension = splitext(absolute_file_path)
+
+            if file_extension == ".md":
+                markdown_file_paths.append(absolute_file_path)
+
+    return markdown_file_paths
+
+
+def clean_content(content):
+
+    for rfc_2119_keyword_regex in rfc_2119_keywords_regexes:
+
+        content = sub(
+            f"\\*\\*{rfc_2119_keyword_regex}\\*\\*",
+            rfc_2119_keyword_regex,
+            content
+        )
+
+    return sub(r"\n>", "", content)
+
+
+def find_rfc_2119_keyword(content):
+
+    for rfc_2119_keyword_regex in rfc_2119_keywords_regexes:
+
+        if search(
+            f"\\*\\*{rfc_2119_keyword_regex}\\*\\*", content
+        ) is not None:
+            return rfc_2119_keyword_regex
+
+
+def parse_requirements(markdown_file_path):
+
+    requirements = []
+
+    with open(markdown_file_path, "r") as markdown_file:
+
+        for requirement in [
+            requirement_match.groupdict() for requirement_match in (
+                finditer(
+                    r"##### Requirement (?P<id>(.*?)+)\n\n"
+                    r"> (?P<content>(.*?))\n\n",
+                    markdown_file.read() + "\n\n",
+                    DOTALL
+                )
+            )
+        ]:
+
+            content = requirement["content"]
+
+            requirements.append(
+                {
+                    "id": requirement["id"],
+                    "clean id": sub(r"[^\w]", "_", requirement["id"].lower()),
+                    "content": clean_content(content),
+                    "RFC 2119 keyword": find_rfc_2119_keyword(content)
+                }
+            )
+
+    return requirements
+
+
+regex = compile_(
+    r"(?P<level>(> ?)*)(?P<pounds>##### )?(?P<content>.*)"
+)
+
+
+def parse_conditions(markdown_file_path):
+
+    conditions = []
+
+    with open(markdown_file_path, "r") as markdown_file:
+
+        for condition in findall(
+            r"##### Condition [0-9.]+\n\n.*?\n\n",
+            markdown_file.read() + "\n\n",
+            DOTALL
+        ):
+
+            stack = []
+
+            text = ""
+
+            for line in condition.split("\n"):
+                regex_dict = regex.match(line).groupdict()
+
+                level = len(regex_dict["level"].split())
+                pounds = regex_dict["pounds"]
+                content = regex_dict["content"]
+
+                if not level and not content:
+                    continue
+
+                if not pounds:
+                    text = "".join([text, content])
+                    continue
+
+                if match(
+                    r"(> ?)*##### Condition [0-9.]+", line
+                ) is not None:
+
+                    node = {
+                        "id": content,
+                        "content": "",
+                        "children": []
+                    }
+                else:
+
+                    content = sub(r"Conditional Requirement ", "", content)
+
+                    node = {
+                        "id": content,
+                        "clean id": sub(r"[^\w]", "_", content.lower()),
+                        "content": "",
+                        "RFC 2119 keyword": None
+                    }
+
+                if not stack:
+                    stack.append(node)
+                    continue
+
+                stack[-1]["content"] = clean_content(text)
+
+                if level == len(stack) - 1:
+
+                    stack[-1]["RFC 2119 keyword"] = find_rfc_2119_keyword(
+                        text
+                    )
+                    stack.pop()
+
+                elif level < len(stack) - 1:
+                    stack[-1]["RFC 2119 keyword"] = find_rfc_2119_keyword(
+                        text
+                    )
+                    for _ in range(len(stack) - level):
+                        stack.pop()
+
+                text = ""
+                stack[-1]["children"].append(node)
+                stack.append(node)
+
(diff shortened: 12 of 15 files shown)


specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+```
+// example provider accessor
+OpenFeature.getProvider();
+```
+
+see [provider](../provider/provider.md)
+
+##### Requirement 1.5
+
+> The `API` **MUST** provide a function for creating a `client`, with an options object containing an optional field for the client `name`.
```
toddbaert, line 50: Is an "options object" too prescriptive? One one hand, it seems less likely to cause breaking changes in many languages, since new properties can be added, on the other hand, it might not be idiomatic in some cases.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+OpenFeature.getClient({
+  name: 'my-openfeature-client'
+});
+```
+
+The name is a logical identifier for the client.
+
+##### Requirement 1.6
+
+> The client **MUST** provide a method for `hook` addition, which accepts an array of API-conformant `hooks`. This adds the specified hooks to any hooks already attached.
```
toddbaert, line 63: I've intentionally not gone into detail here, since `hooks` will have their own section.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+```
+// example provider mutator
+OpenFeature.setProvider(new MyProvider());
+```
+
+see [provider](../provider//providers.md)
+
+##### Requirement 1.3
+
+> The `API` **MUST** provide a function for `hook` addition, which accepts an array of API-conformant `hooks`. This adds the specified hooks to any hooks already attached.
```
toddbaert, line 28: I've intentionally not gone into detail here, since `hooks` will have their own section.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+client.addHooks([new MyHook()]);
+```
+
+see [hooks](./hooks.md)
+
+#### Flag Evaluation
+
+##### Requirement 1.7
+
+> The `client` **MUST** provide methods for typed flag value evaluation, including boolean, numeric, string, and map (object), with parameters `flag key` (string, required), `default value` (boolean | number | string | object, required), `evaluation context` (optional), and `evaluation options` (optional).
```
toddbaert, line 76: Is the `boolean | number | string | object` syntax clear?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+> In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` **MUST** be set to `"ERROR"`.
+
+##### Requirement 1.17
+
+> The `evaluation details` wrapper **MUST** have an optional `error code` field of type string, which, in cases of abnormal execution, identifies an error occurred during flag evaluation, with possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+
+##### Requirement 1.18
+
+> If the `reason` in the `evaluation details` is set to `"ERROR"`, the sibling `error code` field **MUST** be set.
```
toddbaert, line 152: This is dependent on the outcome of the discussion here: https://github.com/open-feature/spec/issues/32#issuecomment-1101612389

toddbaert:
We also may want a section that specifically mentions that we encourage idiomatic use of language features, and that prescriptions that imply certain unavailable language features should be tolerated.

specification/design-principles.md
```diff
@@ -0,0 +1,5 @@
+# Design Principles
+
+## Overview
+
+// TODO: this is a stub.
```
toddbaert, line 5: I think it's a good idea to mention some high-level design philosophies, for instance:

- no/low dependency evaluation API
- vendor neutrality
- extensibility concepts (hooks, etc)
- first class integrations (OTel, for example)

README.md
```diff
@@ -32,6 +32,21 @@ external evaluation engine in a vendor agnostic way.
 
 The OpenFeature project will include client libraries for common technology stacks including, but not limited to:
 
-* Golang
-* Java
-* JavaScript/TypeScript
+- Golang
+- Java
+- JavaScript/TypeScript
```
InTheCloudDan, line 37: Given we haven't approached the client vs server side yet, do we want to be more constrained here and limit it to NodeJS?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+##### Requirement 1.19
+
+> The `evaluation details` wrapper **MUST** have a `executed hooks` field, containing an object which itself contains 4 arrays, identified with keys (`before`, `after`, `error` and `finally`) indicating the hooks run for the respective life-cycle stages.
+
+```
+executedHooks: {
+  before: [],
+  after: [],
+  error: [],
+  finally: []
```
justinabrahms, line 163: Not sure how we want to handle it, but 'finally' is a java reserved word. I've chosen "afterAll" instead.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
+# Flag Evaluation API
+
+## Overview
+
+The `evaluation API` allows for the retrieval of feature flag values, independent of any flag control plane or vendor. In the absence of a feature flag system implementation (a [provider](../provider/providers.md)) the `evaluation API` uses the "No-op provider", which simply returns the supplied default flag value.
```
InTheCloudDan, line 5: minor nit, should we rephrase this as `return of feature flag values` in place of `retrieval`. 

Is it worth calling out optional reason/metadata of why it was returned/error info?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+> No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally end. Flag evaluation calls must always return the `default value` in the event of abnormal execution.
+
+##### Condition 1.23
+
+> The language type system differentiates between floating and integer type numeric values.
+>
+> ##### Conditional Requirement 1.23.1
+>
+> > The `client` **MUST** provide functions for the resolution of both floating point and integer numeric values.
```
justinabrahms, line 187: Do you see this as `getFloatDetails` and `getIntegerDetails` instead of `getNumberDetails`?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+The `evaluation API` allows for the retrieval of feature flag values, independent of any flag control plane or vendor. In the absence of a feature flag system implementation (a [provider](../provider/providers.md)) the `evaluation API` uses the "No-op provider", which simply returns the supplied default flag value.
+
+#### Initialization and Configuration
+
+##### Requirement 1.1
+
+> The `API`, and any state it maintains **SHOULD** exist as a global singleton, even in cases wherein multiple versions of the `API` are present at runtime.
+
+It's important that multiple instances of the `API` not be active, so that state stored therein, such as the registered `provider`, static global `evaluation context`, and globally configured `hooks` allow the `API` to behave predictably. This can be difficult in some runtimes or languages, but implementors should make their best effort that only a single instance of the `API` is used.
```
InTheCloudDan, line 13: I did not realize this was a consideration. It is not what we've seen in practice. It is generally how users work but not all of them. We have customers that instantiate multiple clients in an application.

specification/flag-evaluation/flag-evaluation.md
InTheCloudDan: if the `name` is optional, it does seem like the configuration itself should be an optional field. But if at a minimum it's required to pass a `name` then the options object makes sense being required.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+```
+// example flag activity evaluation
+boolean enabled = client.isEnabled('my-flag', false);
+```
+
+##### Requirement 1.9
+
+> The `client` method for flag activity evaluation **MAY** be semantically different from the boolean flag evaluation method.
+
+Some providers may draw a distinction between boolean flag value evaluation, and flag activity. Put another way, a flag may be "disabled", but simultaneously have an associated auxillary boolean payload of `true`.
```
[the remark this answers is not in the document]
InTheCloudDan: Is there additional information around this? I'm having some trouble parsing the use case that isn't handled internally in the provider.

specification/flag-evaluation/flag-evaluation.md
toddbaert: I'm OK with that. I think the only thing to consider is that in some languages it's less of a  breaking change to add new optional fields to an object, so things are slightly less brittle. I dont even know if that's true in Java though, I don't think so.

I will make this phrase less opinionated.

specification/flag-evaluation/flag-evaluation.md
toddbaert: Oh, right. I like `finally` because really these sort of hooks behave just like a `finally` block, so it's intuitive. But if it's a reserved word in Java, it may be in other languages as well (frankly I'm a bit surprised the TS parser has no issue with it).

I'm OK with making it "afterAll" in general, or we could say it's `finally` OR `afterAll`.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+> The `evaluation details` wrapper **MUST** have a `reason` field of type string indicating the semantic reason for the returned flag value. Possible values vary by provider, but may include `"TARGETING_MATCH"`, `"SPLIT"`, `"DISABLED"`, `"DEFAULT"`, `"UNKNOWN"` and `"ERROR"`.
+
+##### Requirement 1.16
+
+> In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` **MUST** be set to `"ERROR"`.
+
+##### Requirement 1.17
+
+> The `evaluation details` wrapper **MUST** have an optional `error code` field of type string, which, in cases of abnormal execution, identifies an error occurred during flag evaluation, with possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
```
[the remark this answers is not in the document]
toddbaert: It all goes back to https://github.com/open-feature/spec/issues/32#issuecomment-1101612389 and the linked discussion. I think we should decide if we want this to be free-form or not. This is an open question as I mentioned here: https://github.com/open-feature/spec/pull/35/files/a782d675f1aef6924092f7b9421a53c2f2ce449f#r855313094

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.11
+
+> The `evaluation details` wrapper **MUST** have a `value` field matching the type of the flag value, and containing to the flag value.
+
+##### Condition 1.12
+
+> The language supports generics (or an equivalent feature).
+>
+> ##### Conditional Requirement 1.12.1
```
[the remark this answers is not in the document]
toddbaert: Ah, I see that. It didn't happen on my local preview. I think the spec parser might require this, but I can tweak it. It looks somewhat ugly this way.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: >This single API will tell you "Is this enabled in the UI" and also "is the flag evaluating to true"?

Not quite (unless I'm misunderstanding you)

The issue is  that some providers allow logically boolean flags (enabled/disabled) WITH separated associated boolean payloads (every flag has a bool and an optional number/string/boolean payload). Because some providers work that way, I think it forces us to have this kind of resolution - we need a method for the boolean flag, and a separate method to get it's associated boolean payload.

In providers that DONT have this concept, `isEnabled()` and `getBooleanValue()` do the same thing

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+// example overloaded string flag evaluation with optional params
+string myString = client.getStringValue('string-flag', 'N/A', evaluationContext, options);
+
+```
+
+see [evaluation context](../evaluation-context/evaluation-context.md)
+
+##### Requirement 1.8
+
+> The `client` **MUST** provide a method for flag activity evaluation, denoting if the flag is "enabled" or not, with parameters `flag key` (string, required), `default value` (boolean, required), `evaluation context` (optional), and `evaluation options` (optional).
```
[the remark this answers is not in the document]
toddbaert: see: https://github.com/open-feature/spec/pull/35/files/a782d675f1aef6924092f7b9421a53c2f2ce449f#r855378994

specification/flag-evaluation/flag-evaluation.md
toddbaert: There's no problem with multiple clients (see 1.5). This is just a requirement for the API. A global singleton is necessary for advanced use cases, like "injecting" implementation into 3rd party libraries. If a third party implements the flag API, feature flagging can be used therein because it will be referencing the same global singleton, configured with a particular provider.

specification/flag-evaluation/flag-evaluation.md
toddbaert: > if the name is optional, it does seem like the configuration itself should be an optional field. But if at a minimum it's required to pass a name then the options object makes sense being required.

Ya, you're right, I guess this whole options field can be optional in languages that use an object.

specification/flag-evaluation/flag-evaluation.md
toddbaert: > Do you see this as getFloatDetails and getIntegerDetails instead of getNumberDetails?

That's pretty much what I had in mind, ya. Cloudbees Feature Management has something similar.

README.md
toddbaert: That's probably a good idea for now.

specification/flag-evaluation/flag-evaluation.md
toddbaert: Thinking about it more, I think "evaluation" is probably the best word. What do you think?

I also agree it's worth mentioning the metadata, yes.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: @InTheCloudDan 

>Is there additional information around this? I'm having some trouble parsing the use case that isn't handled internally in the provider.

See: https://github.com/open-feature/spec/pull/35/files#r855378994. It is really internally handled by the provider, but I wanted to call it out here because it's relevant to the expected behavior of the evaluation API.

Maybe this proposal isn't the best solution, but @dabeeeenster has brought the issue up a few times, and this is the best I can come up with. Open to more suggestions though.

specification/README.md
```diff
@@ -0,0 +1,33 @@
+OpenFeature Specification
+
+## Contents
+
+- [Glossary](./glossary.md)
```
dyladan, line 5: Is there any tooling to keep the table of content synchronized and correct? Could easily become out of date and be confusing.

specification/flag-evaluation/flag-evaluation.json
```diff
@@ -0,0 +1,164 @@
+[
```
dyladan, line 1: Accidentally checked in JSON? .gitignore?

README.md
```diff
@@ -32,6 +32,21 @@ external evaluation engine in a vendor agnostic way.
...
+
+### Specification style guide and tooling
+
+This specification complies with [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) and seeks to conform to the [W3C QA Framework Guildlines](https://www.w3.org/TR/qaframe-spec/).
+
+In accordance with this, some basic tooling (donated graciously by [Diego Hurtado](https://github.com/ocelotl)) has been employed to parse the specification and output a JSON structure of concise requirements, highlighting the particular `RFC 2119` verb in question.
+
+To parse the specification, simply type `make`. Please review the generated JSON files, which will appear as siblings to any of the markdown files in the `/specification` folder.
+
+- Use code blocks for examples.
```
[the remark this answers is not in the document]
dyladan: I would add "each sentence on its own line" which makes it MUCH easier to review in pull request form because I can make suggestions, comments, and corrections on a single sentence rather than a multi-sentence line.

specification/flag-evaluation/flag-evaluation.md
dyladan, line 5: I would either use the term "feature flag system implementation" or "provider" and not both. If you insist on using both I would put it the other way around as "provider" is the official name and "feature flag system implementation" is a description of what it is. IMO the link to the provider spec explains it sufficiently though.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.1
+
+> The `API`, and any state it maintains **SHOULD** exist as a global singleton, even in cases wherein multiple versions of the `API` are present at runtime.
+
+It's important that multiple instances of the `API` not be active, so that state stored therein, such as the registered `provider`, static global `evaluation context`, and globally configured `hooks` allow the `API` to behave predictably. This can be difficult in some runtimes or languages, but implementors should make their best effort that only a single instance of the `API` is used.
+
+##### Requirement 1.2
+
+> The `API` **MUST** provide a function to set the `provider`, which accepts an API-conformant `provider` implementation.
```
dyladan, line 17:
```suggestion
> The `API` **MUST** provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
```

specification/flag-evaluation/flag-evaluation.md
dyladan: Specifying it must be an array is an implementation detail IMO and depends on the idioms of the implementation language. In some languages it is idiomatic to have variadic arguments for example.

```suggestion
> The `API` **MUST** provide a function to register one or more `hooks`. This adds the specified hooks to any hooks already attached.
```

"This adds the specified hooks to any hooks already attached" reads weird to me. I sort of get what you're going for (that it doesn't replace existing hooks, but includes new ones) but it seems like a truism "addHooks adds hooks"

```suggestion
> The `API` **MUST** provide a function for `hook` addition, which accepts an array of API-conformant `hooks`. When new hooks are added, previously added hooks are not removed.
```

This also makes execution order ambiguous. I would either specify that it is not order guaranteed, or use a word like "append" or add a sentence like "hooks are executed in the order they are added".

```suggestion
> The `API` **MUST** provide a function for `hook` addition, which accepts an array of API-conformant `hooks`. This adds the specified hooks to any hooks already attached. Hooks are executed in the order they are added.
```

And all together:

```suggestion
> The `API` **MUST** provide a function to register one or more `hooks` which accepts an array of API-conformant `hooks`. When a new hook is added, previously added hooks are not removed. Hooks are executed in the order they are added.
```

specification/flag-evaluation/flag-evaluation.md
dyladan: The wording seems too oriented around the languages you're familiar with and similar to the above comment about the array of hooks seems to me an implementation detail which could have different idioms in different languages. For example optional arguments are often keyword arguments in python.

```suggestion
> The `API` **MUST** provide a function for creating a `client` which accepts the following options:
>
> - name (optional): A logical string identifier for the client.
```

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+The name is a logical identifier for the client.
+
+##### Requirement 1.6
+
+> The client **MUST** provide a method for `hook` addition, which accepts an array of API-conformant `hooks`. This adds the specified hooks to any hooks already attached.
+
+```
+// example hook attachment
+client.addHooks([new MyHook()]);
```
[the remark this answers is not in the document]
dyladan: API hooks _and_ client hooks? What is the difference? All my previous comments about that function apply to this one.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+```
+// example client creation and retrieval
+OpenFeature.getClient({
+  name: 'my-openfeature-client'
+});
+```
+
+The name is a logical identifier for the client.
+
+##### Requirement 1.6
```
dyladan, line 61: I was very confused by the sudden change from API to client. Maybe make a heading or even a new file?

specification/flag-evaluation/flag-evaluation.md
dyladan: Not at all clear to the uninitiated what "typed flag value evaluation" is

dyladan: See my comments on addHooks about language agnosticism. Map/object may not apply to all languages you want to target, and many languages differentiate between Map and object.  I would name and define this concept, and let any language use whatever mechanism is appropriate. Is it a map from string keys to (number | string | boolean)? Can the object have nested objects? Can the object have non-string keys? Can the values be arrays? If so, must those arrays be homogeneous?

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
dyladan: "flag activity evaluation" means nothing to people who don't already know what it means. What does it mean for a flag to be enabled? Does this mean it has a value of true, or that it exists in the backend?

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
dyladan: I'm also confused by this but it seems others are on top of clarifications

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.9
+
+> The `client` method for flag activity evaluation **MAY** be semantically different from the boolean flag evaluation method.
+
+Some providers may draw a distinction between boolean flag value evaluation, and flag activity. Put another way, a flag may be "disabled", but simultaneously have an associated auxillary boolean payload of `true`.
+
+##### Requirement 1.10
+
+> The `client` **MUST** provide methods for flag value evaluation with the value wrapped in a `evaluation details` object, with parameters `flag key` (string, required), `default value` (boolean | number | string | object, required), `evaluation context` (optional), and `evaluation options` (optional).
```
dyladan, line 107: You use language-leaning wording again here, but it is less egregious so i'll let you decide if you want to change it.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+FlagEvaluationDetails<boolean> myBoolDetails = client.getBooleanDetails('bool-flag', false);
+
+// example detailed string flag evaluation
+FlagEvaluationDetails<string> myStringDetails = client.getStringDetails('string-flag', 'N/A', evaluationContext, options);
+
+```
+
+##### Requirement 1.11
+
+> The `evaluation details` wrapper **MUST** have a `value` field matching the type of the flag value, and containing to the flag value.
```
dyladan, line 120:
```suggestion
> The `evaluation details` wrapper **MUST** have a `value` field matching the type of the flag value, and containing the flag value.
```

What does "matching the type of the flag value" mean?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Condition 1.12
+
+> The language supports generics (or an equivalent feature).
+>
+> ##### Conditional Requirement 1.12.1
+>
+> > The `evaluation details` wrapper **SHOULD** accept a generic argument (or utilize an equivalent language feature) which indicates the type of the wrapped `value` field.
```
dyladan, line 128: What happens if there is a type mismatch?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.13
+
+> The `evaluation details` wrapper **MUST** have a `flag key` field field of type string, representing the flag key.
```
dyladan, line 132:
```suggestion
> The `evaluation details` wrapper **MUST** have a `flag key` field of type string, representing the flag key.
```

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.14
+
+> The `evaluation details` wrapper **MUST** have an optional `variant` field of type string which represents a space-constrained identifier for a particular flag value.
```
dyladan, line 136: What do you mean space-constrained?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.15
+
+> The `evaluation details` wrapper **MUST** have a `reason` field of type string indicating the semantic reason for the returned flag value. Possible values vary by provider, but may include `"TARGETING_MATCH"`, `"SPLIT"`, `"DISABLED"`, `"DEFAULT"`, `"UNKNOWN"` and `"ERROR"`.
```
dyladan, line 140: Seems like this could be an optional field

specification/flag-evaluation/flag-evaluation.md
dyladan: Error code was quite a contentious topic in opentelemetry which we went back and forth on for a long time. I would think about what usecase you're trying to solve by having an error code and see if you can solve it in a more explicit way. Maybe a "retryable" boolean?

specification/flag-evaluation/flag-evaluation.md
dyladan: Avoiding using well known reserved words is a good idea, but in general it may be impossible to completely avoid reserved words. Maybe add some guidance to the spec that if a reserved word is used then choose an appropriate substitute.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+  before: [],
+  after: [],
+  error: [],
+  finally: []
+}
+```
+
+##### Requirement 1.20
+
+> The value of the elements in the arrays contained in the `executed hooks` object **SHOULD** allow authors to uniquely identify the hooks executed during the flag evaluation lifecycle. This could be done with a mandatory string field defined on the `hook` interface, or by using the fully qualified class name in languages where such concepts exist.
```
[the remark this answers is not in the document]
dyladan: This seems like a feature that I would not include in version 1. If there is need for it then it can always be added later, but for now it seems like trying to cover all the bases before you know if you are even playing baseball.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+
+##### Requirement 1.21
+
+> The `evaluation options` object **MUST** have a `hooks` field which can contain hooks to be executed for the respective flag evaluation.
+
+see [hooks](./hooks.md)
+
+##### Requirement 1.22
+
+> No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally end. Flag evaluation calls must always return the `default value` in the event of abnormal execution.
```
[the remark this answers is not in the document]
dyladan: I would carve out an exception for configuration and setup. If something is incorrectly setup, then failing on startup is a reasonable choice and swallowing that error might be confusing.

specification/flag-evaluation/flag-evaluation.md
dyladan: Maybe an example would help. Also, there could be a language which differentiates between these types but allows for unified handling of them (like python) or languages with more number types than just these. I would leave it up to the language authors to do whatever is idiomatic in their language.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,195 @@
...
+>
+> > The `client` **MUST** provide functions for the resolution of both floating point and integer numeric values.
+
+##### Condition 1.24
+
+> The language features async/await semantics.
+>
+> ##### Conditional Requirement 1.24.1
+>
+> > The `client` **SHOULD** provide asynchronous functions for flag evaluation.
```
dyladan, line 195: This is extremely specific. I would generalize it to something like "The client MUST provide a function for flag evaluation which does not block the thread of execution". I would also not gate this on async/await specifically because that is only one of many ways to achieve this goal.

dyladan (commented):
I did not review the parser at all, but I looked pretty thoroughly at the spec. In general, I would characterize it as a bit too prescriptive and doesn't leave room for the idioms of different languages to be used in some places. This may lead to an API implemented in one language which _feels_ like it doesn't really belong. Take all my suggestions as merely suggestions as I am an outsider here.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
justinabrahms: @toddbaert Which providers do that? A flag evaluation that's "off" seems like it should return the default value and a reason for it.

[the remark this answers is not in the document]
toddbaert: @justinabrahms  Flagsmith, particularly behaves this way. It can return false and a payload.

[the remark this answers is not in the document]
dabeeeenster: Just to chime in here - we see the notion of a flag being `false` and consequently disabled or providing some form of convention in some way as something of an anti-pattern. There are use cases where flags are used in a manner that they travel from true > false for good reason - e.g. retiring a feature. 

We've designed Flagsmith in a manner that `false` does not impute anything in that regard - it's just something that is `false` and up to the developer to interpret that however they see fit.

specification/flag-evaluation/flag-evaluation.md
toddbaert: @dyladan  Good feedback here, I've amalgamated a few of these and updated.

We will have an entire hook section, and I wanted to talk about ordering in that document in some more detail. To that end, we'd discussed `before` hooks running in the order they are added, with all other hooks (`after`, `error`, `finally`, which run after evaluation) running in reverse order. Do you think that's too much to add here?

specification/flag-evaluation/flag-evaluation.md
toddbaert: Yep, this resolves @justinabrahms point too.

specification/flag-evaluation/flag-evaluation.md
beeme1mr: The main use cases we're focusing on is telemetry and debugging.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: So, in Flagsmith, I can setup a single flag, let's call it `toddsFlag`. Like all flags in Flagsmith, it has a boolean and an optional `boolean | string | number` auxiliary value, depending on how you've configuration this it in this case. Let's consider the auxiliary flag is boolean for the rest of this discussion.

This same flag can used by the flagsmith SDK 2 ways:

`flagsmith.hasFeature('toddsFlag')` which returns a boolean corresponding to the big toggle switch on the Flagsmith UI. 

AND 

`flagsmith.getValue('toddsFlag')` which can return a boolean completely independent of the returned value of `hasFeature` above, true or false.

It was supporting this that I was attempting to do, but after this discussion, I'm wondering if it's not better to completely remove the `isEnabled()` method entirely, and let the Flagsmith provider determine how to handle resolving `getBooleanValue` in their case.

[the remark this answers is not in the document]
dyladan: Hmm thanks for the clarification. The wording here makes more sense with this in mind.

> It was supporting this that I was attempting to do, **but after this discussion, I'm wondering if it's not better to completely remove the `isEnabled()` method entirely**, and let the Flagsmith provider determine how to handle resolving `getBooleanValue` in their case.

I'm supportive of this tbh. Seems very provider-specific

specification/flag-evaluation/flag-evaluation.md
toddbaert: Largely agreed, but that would be changing a "SHOULD" to a "MUST". Is that what you recommend? In multithreaded environments like Java, blocking probably isn't that uncommon.

specification/flag-evaluation/flag-evaluation.json
toddbaert: I wasn't sure if I should commit this or not.

README.md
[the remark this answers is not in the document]
toddbaert: That's and interesting suggestion.

specification/flag-evaluation/flag-evaluation.md
toddbaert: I can change it to a "SHOULD" pointing out that implementors should simply consider this.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
beeme1mr: I'm in favor of removing `isEnabled()`. In my opinion, it seems very unlikely that an application author would expect `isEnabled()` and `getBooleanValue()` to behave differently. Removing it would eliminate that confusion.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: I specifically mentioned "client" here, because most configuration is done by the point you have an instance of one, with the exception of adding hooks to the client... I think it's a fair point though, to say this rule really applies to flag evaluation only.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: @beeme1mr I guess I agree with Dan's point here.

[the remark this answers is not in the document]
beeme1mr: It doesn't have to be in the spec but it was a valuable feature in our experimental project. It allows you to understand what hooks ran and in what order.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
dabeeeenster: > flagsmith.getValue('toddsFlag') which can return a boolean completely independent of the returned value of hasFeature above, true or false. @dabeeeenster please correct me here, if I'm wrong.

Correct. 

This is something of a hangover from pure feature flags and a more complete flag and remote config solution. We wanted to give people a simple boolean state for a flag but then be able to enhance it with essentially a general String value, but then what if they store `True` in the string value, and how does that typing get supported through all the SDKs and their various type handling.

Hard to say what the best general solution is here. I think if the OpenFeature SDK can provide its best interpretation of the intention from the provider's interface but let developers have access to the base data is probably the optimal solution. 

I believe most vendors have a similar approach of booleans and additional config/multi-variate values but I cant (and don't want to !) speak for them.

specification/flag-evaluation/flag-evaluation.json
dyladan: Up to you. Maybe create a CI step that runs the generator and fails if there are any files in `git diff` which would ensure that the JSON is always up to date.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
beeme1mr: After some more thought, I agree. It's a good idea but not necessary in the initial spec.

specification/flag-evaluation/flag-evaluation.md
dyladan: Whether you want SHOULD or MUST, the important part is not the async/await syntax but the execution blocking

specification/flag-evaluation/flag-evaluation.md
toddbaert: I think it might be best to say something suggesting that the language use whatever construct is usually used to interrogate and manipulate JSON objects, that's really what these are in most backends.

@justinabrahms @dyladan what do you think?

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: > API hooks and client hooks? What is the difference? All my previous comments about that function apply to this one.

There's only one API, but n number of clients. Clients might have a much smaller scope than the API, and attaching hooks to it only applies to flags evaluated using that client.

I was going to mention more about this in the `hooks` spec.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: You're right. I think we might be removing this completely based on the [this](https://github.com/open-feature/spec/pull/35#discussion_r855574278) discussion.

specification/flag-evaluation/flag-evaluation.md
toddbaert: Agreed. I will add a heading. If this file starts to get too big, it might make sense to break it up, but I'm not convinced it's justified yet.

specification/flag-evaluation/flag-evaluation.md
toddbaert: I will change `object` to `interface`.

specification/flag-evaluation/flag-evaluation.md
toddbaert: Agreed.

specification/flag-evaluation/flag-evaluation.md
dyladan: JSON or arbitrary structured data? Could easily have a system which transmits these in YAML for example. Would that provider be expected to convert to a JSON specific structure?

```suggestion
> The `client` **MUST** provide methods for typed flag value evaluation, including boolean, numeric, string, and structured data, with parameters `flag key` (string, required), `default value` (boolean | number | string | object, required), `evaluation context` (optional), and `evaluation options` (optional). Structured data **SHOULD** be presented however is idiomatic in the implementation language for arbitrarily structured data such as JSON or YAML.
```

specification/flag-evaluation/flag-evaluation.md
toddbaert: I was going to mention more about this in this provider spec, but I should stay in this spec that all the flag evaluation methods should default when there's a type mismatch (for the provider spec, the provider should throw or otherwise indicate abnormal execution).

specification/flag-evaluation/flag-evaluation.md
toddbaert: We're going to remove this for now anyway, as discussed in other comments, unless anyone disagrees. @justinabrahms @dyladan

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: @justinabrahms @InTheCloudDan @beeme1mr  How do you think we should proceed here? At this point, I dont mind which path we go down (enum or string) at this point, I'm sure if we make the wrong choice it will become apparent in alpha. Please take a look at the linked discussion. We can also take a vote if that's preferable.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
beeme1mr: If this functionality is there due to historical reasons, I propose that we do not add this to the spec. I think the expected behavior for OpenFeature application developers would be that a disabled feature flag does not have any meaningful auxiliary value. In this case, it would simply return the default value provided during the evaluation. However, the provider pattern is flexible enough to allow each vendor to decide how best to implement the interface. In this case, the `getBooleanValue()` could get the value from the state of the flag or from the `getValue()` method.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
justinabrahms: No strong opinions yet. Happy to do it in strings in javascript and enums in java and see which ones we like. :)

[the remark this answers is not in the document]
toddbaert: OK. I think we'll keep the related discussion open so we remember to come back to it, and I'll go forward with strings for now in the spec. I'm totally open to discuss it more, nothing here is set in stone.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: For now, I've removed the `isEnabled()` and `isEnabledDetails()` methods. We can always open up a new issue for this if we think it's something that's absolutely necessary, but I agree that the `provider` interface (spec to come) gives enough flexibility to address this in numerous ways.

specification/flag-evaluation/flag-evaluation.json
toddbaert: I've opened: https://github.com/open-feature/spec/issues/36

specification/README.md
toddbaert: I've opened: https://github.com/open-feature/spec/issues/36

README.md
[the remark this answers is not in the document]
toddbaert: I added this to the style guide, an opened: https://github.com/open-feature/spec/issues/36

specification/types.md
```diff
@@ -0,0 +1,37 @@
...
+- value (boolean | string | number | structure, required)
+- error code (string, optional)
+- reason (string, optional)
+- variant (string, optional)
+
+### evaluation options
+
+A structure containing the following fields:
+
+- hooks (one or more [hooks](./flag-evaluation/hooks.md), optional)
```
toddbaert, line 37: At the advice of @dyladan , I've added this "types" section. I agree with his assertion that it will avoid repetition, and allow flexibility and idiomatic implementations across languages.

specification/types.md
```diff
@@ -0,0 +1,37 @@
...
+
+### string
+
+A UTF-8 encoded string.
+
+### number
+
+A numeric value of unspecified type or size. Implementation languages may further differentiate between integers, floating point numbers, and other specific numeric types and provide functionality as idioms dictate.
+
+### structure
```
toddbaert, line 19: @justinabrahms

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
justinabrahms: One thing came up for me:  This is still true for the case where something in the `finally()` hook errors out, right? If that's the case, I think it will make the logic quite odd for handling those exceptions, making it not much different than `after`

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,188 @@
...
+```
+// example flag evaluation
+var myValue = client.getValue('my-flag', false);
+```
+
+##### Condition 1.8
+
+> The language type system differentiates between strings, numbers, booleans and structures.
+>
+> ##### Conditional Requirement 1.8.1
```
justinabrahms, line 93: I still don't think we want these in blockquotes.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: I think the `finally` or `afterAll` hook is pretty similar functionally to the `after` hook. In my mind, the big difference is that the `after` hook would only be executed if flag evaluation proceeds normally, while the `finally` hook executes in both normal, and abnormal `error` execution. What do you think?

justinabrahms (approved):

specification/flag-evaluation/flag-evaluation.md
toddbaert: I agree, but for now our fancy little W3C-QA-compliant spec parser requires it. I will make an issue to resolve that. I agree it looks ugly.

Added chore issue: https://github.com/open-feature/spec/issues/37

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,188 @@
...
+
+> The `evaluation options` structure's `hooks` field denotes a collection of hooks to be executed for the respective flag evaluation.
+
+See [hooks](./hooks.md) for details.
+
+##### Requirement 1.19
+
+> No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.
+
+##### Requirement 1.20
```
moredip, line 184: I think we should take a harder stance here. Either the SDK supports async evaluation (and the Application Author has to plan for that) or the SDK does not support it and our spec should have explicitly defined semantics for what happens when you ask for a flag evaluation but it can't be done synchronously.

The specific scenario I'm thinking about is flagging for clientside SPAs,  where you often need to evaluate a flag at the point a page is loading, where they may well not be any locally cached flag configuration loaded yet.

specification/glossary.md
```diff
@@ -86,7 +90,7 @@ Flags represent a single pivot point of logic. Flags have a type, like `string`,
 
 ### Variant
 
-A variant is a space-constrained identifier for values. This allows us to reference these values without having to include all of the data (which may be quite large). We may use some hash of the values themselves or a string name. So for the `header-order` flag, we may have variants like `reverse`, `wonky` or `standard`.
+A variant is a identifier for a value. This allows for the referral to particular values without necessarily to including the value itself, which may be quite prohibitively large or otherwise unsuitable in some cases.
```
moredip, line 93: there's some iffy grammar here

```suggestion
A variant is a identifier for a value. This allows for referral to particular values without necessarily including the value itself, which may be quite prohibitively large or otherwise unsuitable in some cases.
```

Also maybe "A variant is a semantic identifier for a value" might be a bit clearer?

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,188 @@
...
+>
+> > See [evaluation context](../evaluation-context/evaluation-context.md) for details.
+>
+> ##### Conditional Requirement 1.8.2
+>
+> > The `client` **SHOULD** guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.
+
+##### Requirement 1.9
+
+> The `client` **MUST** provide methods for detailed flag value evaluation with the value wrapped in a `evaluation details` structure, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional).
```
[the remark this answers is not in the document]
toddbaert: These are the parameters to the method, not the fields in the structure. I will rephrase to make that more clear.

specification/flag-evaluation/flag-evaluation.md
toddbaert: I think we need to consider languages and runtimes where asynchronous methods are not available or less common. For example, in many Java frameworks, thread-per-request is quite standard, so blocking while the flag is evaluated seems pretty reasonable there. Only a single request is blocked.

I see your concern, I just dont want to make it a hard requirement for all languages, especially ones in which blocking is common, or even idiomatic.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: Again, these are the parameters to the method, not the fields in the structure. I will rephrase to make that more clear. The structure is detailed in `types.md`, as well as below.

specification/glossary.md
toddbaert: I like both these suggestions.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: rephrased, let me know if that's clearer.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,188 @@
...
+
+> The `evaluation details` structure's `reason` field **SHOULD** indicate the semantic reason for the returned flag value. Possible values vary by provider, but may include `"TARGETING_MATCH"`, `"SPLIT"`, `"DISABLED"`, `"DEFAULT"`, `"UNKNOWN"` and `"ERROR"`.
+
+##### Requirement 1.15
+
+> In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` **SHOULD** be set to `"ERROR"`.
+
+##### Requirement 1.16
+
+> In cases of abnormal execution, the `evaluation details` structure's `error code` field **SHOULD** identify an error occurred during flag evaluation, with possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
```
agentgonzo, line 168: Is it worth stating as a requirement that in the case of normal execution, the `error code` field should not be set?

specification/flag-evaluation/flag-evaluation.md
agentgonzo: Why not say that the client **should** provide both async and sync mechanisms?
One consumer will probably want to almost-always call async methods, yet another consumer will probably want to almost-always call sync methods. Forcing only async to the consumer will likely result in just a lot of unnecessary `async` (JS) or `.get()' (Java future) calls, which they will probably write a simple wrapper around - so why not make that wrapper part of the spec?

It's pretty trivial for a provider to turn an async method into a sync one by using `await` `.get()` if their vendor-specific implementation only provides async methods, or turn a sync method into an async one by wrapping with `new Promise` etc.

specification/types.md
```diff
@@ -0,0 +1,37 @@
...
+
+Structured data, presented however is idiomatic in the implementation language, such as JSON or YAML.
+
```
agentgonzo, line 22: `evaluation context` is missing from this section
```suggestion
### evaluation context

Structured data, presented however is idiomatic in the implementation language, such as JSON or YAML.

```

specification/flag-evaluation/flag-evaluation.json
```diff
@@ -0,0 +1,140 @@
...
+    {
+        "id": "1.15",
+        "clean id": "1_15",
+        "content": "In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` SHOULD be set to `\"ERROR\"`.",
+        "RFC 2119 keyword": "SHOULD"
+    },
+    {
+        "id": "1.16",
+        "clean id": "1_16",
+        "content": "In cases of abnormal execution, the `evaluation details` structure's `error code` field SHOULD identify an error occurred during flag evaluation, with possible values `\"PROVIDER_NOT_READY\"`, `\"FLAG_NOT_FOUND\"`, `\"PARSE_ERROR\"`, `\"TYPE_MISMATCH\"`, or `\"GENERAL\"`.",
```
agentgonzo, line 83: are you literally having the structure have a field called `error code`, or is "error code" a generic name for the field that will have a different string literal in the code?

If the former, then I would consider renaming it so that the field name doesn't have whitespace in it. If the latter, then that's fine (albeit slightly confusing to me as it stands)

toddbaert: No, this is just an effort to stay language neutral, since there's differing casing-conventions. I certainly don't recommend whitespace in fields. See the style guide (and feel free to disagree with it: https://github.com/open-feature/spec/blob/spec/flag-evaluation-api/README.md#style-guide)

> Use "sentence case" enclosed in ticks (`) when identifying entities outside of code blocks (ie: evaluation details instead of EvaluationDetails).

specification/types.md
toddbaert: This is not just the `evaluation context`. "Structure" is a very generic, language neutral way to say arbitrarily structured data. For example, the thing returned from the `getObjectValue` flag call.

specification/flag-evaluation/flag-evaluation.md
toddbaert: Yes, I think that's a good point.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,188 @@
...
+>
+> > See [evaluation context](../evaluation-context/evaluation-context.md) for details.
+>
+> ##### Conditional Requirement 1.8.2
+>
+> > The `client` **SHOULD** guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.
+
+##### Requirement 1.9
+
+> The `client` **MUST** provide methods for detailed flag value evaluation with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns an `evaluation details` structure.
```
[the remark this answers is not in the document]
toddbaert: Basically. All the additional fields besides value are optional if you check the `.types.md` file.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -0,0 +1,188 @@
...
+> >
+> > // example overloaded structure flag evaluation with optional params
+> > MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);
+> > ```
+>
+> > See [evaluation context](../evaluation-context/evaluation-context.md) for details.
+>
+> ##### Conditional Requirement 1.8.2
+>
+> > The `client` **SHOULD** guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.
```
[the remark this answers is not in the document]
toddbaert: Yes, this is a good point. I could see how that would be an issue. We can say something about logging.

README.md
```diff
@@ -32,6 +32,25 @@ external evaluation engine in a vendor agnostic way.
 
 The OpenFeature project will include client libraries for common technology stacks including, but not limited to:
 
-* Golang
-* Java
-* JavaScript/TypeScript
+- Golang
+- Java
+- JavaScript/TypeScript (Node.js)
```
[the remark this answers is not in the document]
toddbaert: It certainly is, and we don't want to forget about it. We just want to focus on the backend for now, since there's more considerations to be made for client side flags in general.

specification/flag-evaluation/flag-evaluation.md
toddbaert: > Why not say that the client should provide both async and sync mechanisms?

I don't think this is reasonable, or possible in Javascript especially, given that we have vendors SDKs with ONLY async APIs already. If a method returns a promise, there's no way to make that method synchronous and also behave predictably without blocking the thread of execution (which means blocking your entire webapp in the case of node).

You're right that it's certainly possible to make a synchronous call async by wrapping it in a promise, but you cant go the other way. I also didn't want to force languages/frameworks where blocking is reasonable NOT to block (Java). Of course, I'm not against Futures or callbacks, but its certainly not unusual for Java calls to block.

agentgonzo: yeah, you're right. I was thinking it is trivial to just `await` it, but of course, that makes your calling function async, then it's turtles all the way down.

so I wonder about saying "the client **should**" provide both a sync and an async method" (for those who want to call it in either form". If a vendor only provides an async method (and the language can't sync-ify it) then one option _could_ be to throw a `NotImplementedException` or similar. Though I do agree with the concept of 'our evaluation APIs shouldn't throw errors'. Hmmm.... I'll have a think. What are your thoughts in general to providing both sync and async evaluation methods?

agentgonzo: Ooooh, thought.
one pattern used elsewhere would be `OpenFeature.getSyncClient(...)`  and `OpenFeature.getAsyncClient(...)`. That way, if needed, the getSyncClient could throw an error (legal in the current spec) without the `client.getStringValue()` itself needing to throw the exception to go "this isn't supported". Thoughts?

toddbaert: > What are your thoughts in general to providing both sync and async evaluation methods?

I think it makes perfect sense in some implementation languages, but not others. It's perfectly reasonable for Java, but not JS. Like I mentioned before, it's pretty common for JS libs not to have any sync methods at all, and for good reason.

We could change it to a **SHOULD** that would just force the implementer to consider this question, not sure how to word it, but I'll think on it and respond.

toddbaert: > Ooooh, thought.
one pattern used elsewhere would be OpenFeature.getSyncClient(...) and OpenFeature.getAsyncClient(...). That way, if needed, the getSyncClient could throw an error (legal in the current spec) without the client.getStringValue() itself needing to throw the exception to go "this isn't supported". Thoughts?

I think this is a reasonable idea, and might be a good path forward. I suppose my only concern is whether it's possible that there's languages where this concept itself wouldn't be idiomatic. I don't think Golang, with it lightweight goroutine concept and channels, typically sees explicitly async APIs. In Go, it's pretty easy to start a goroutine and wait for it to write to a channel without needing to "mark" the entire stack as async, like you would for async/await in c# or JS.

justinabrahms: My Java implementation is blocking. 

I like the notion of a nonblocking version but I'm unsure if it should be the same repo or a different one. There are a *bunch* of methods we're already providing in the Java sdk. I'm not the most excited about doubling the count, but could do it. I'll be focusing on thread safety first.

agentgonzo: @toddbaert @moredip , for the languages where it doesn't make sense to have one or other of the sync or async client/methods, we could just omit those `get[A]SyncClient(...)` methods from the API.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
agentgonzo: That feels right to me. This is the requirement afterall, so just saying "it should log the error" is good enough for now

specification/flag-evaluation/flag-evaluation.md
toddbaert: OK, so to make sure I understand:

In implementation languages where there's both async and sync methods, we'd recommend (require?) the API to define 2 methods:

`openfeature.getAsyncClient()` and `openfeature.getSyncClient()` (these can abnormally terminate with `NotImplemented`, since there are "configuration")

And in languages where that dichotomy doesn't really exist, it's just:

`openfeature.getClient()`

If that's what you're suggesting, I think we can reflect that in the spec, and I think it's a pretty solid plan, so it has my vote. Let's wait to see if anyone objects, and if not, I'll modify the spec accordingly.

Let me know if I've misunderstood.

agentgonzo: That's _very almost_ what I was suggesting 👍 .

And in languages where that dichotomy doesn't really exist, it's just:

`openfeature.getSyncClient()` or `openfeature.getAsyncClient()` (as appropriate)

IMO this keeps consistency across the different languages (and makes documentation easier for us) and is _ever so slightly_ more informative to the end-user... but TBH this slight modification is a tiny point and I'm happy with either option.

beeme1mr (approved):
It looks like the major issues have been addressed now. I think it's time to merge it as is since nothing here is final. Any issues can be handled with a tickets and/or a PR.

specification/flag-evaluation/flag-evaluation.md
moredip: It feels pretty heavy to have two distinct clients (sync vs async), with two distinct interfaces, **AND** different semantics about how certain error scenarios are handled.

I suspect we're "only" really talking about this because of JS support, where you can't have a blocking call. I wonder if it would be better to add a special case for that scenario - and specify that "clients which cannot support a blocking method call should instead provide an equivalent asynchronous variant" or something like that?

toddbaert: > AND different semantics about how certain error scenarios are handled.

I don't think different error semantics were being proposed, though maybe I missed something. There's just a stipulation that no errors are thrown with flag evaluation, but an exception to that rule is configuration/setup code, which includes getting a client. This means, to @agentgonzo 's point, that a `NotImplemented` could be thrown.

> I suspect we're "only" really talking about this because of JS support, where you can't have a blocking call. I wonder if it would be better to add a special case for that scenario - and specify that "clients which cannot support a blocking method call should instead provide an equivalent asynchronous variant" or something like that?

Yes, that, combined with the fact with already have a provider with an async underlying SDK, is the only reason I added anything about asynchrony in the spec at all, and perhaps that was a mistake. I think for now we should merge without having this nailed down, and we can revisit this specific issue later.

moredip (approved):

specification/flag-evaluation/flag-evaluation.md
justinabrahms: @toddbaert I'm not sure. I think we leave this as is, merge it, then discuss in an issue.

I would want to think through what a sync hook vs a sync hook might look like before I committed one way or the other, but this is good enough to make progress on imo.

toddbaert:
Rebased, squashed, added co-authors.

PR #41, opened 2022-04-27
Add provider spec

toddbaert:
Added provider spec, with minor associated changes to the existing flag-evaluation spec file.

### DIFF
--- a/specification/flag-evaluation/flag-evaluation.json
+++ b/specification/flag-evaluation/flag-evaluation.json
@@ -62,56 +62,56 @@
     {
         "id": "1.13",
         "clean id": "1_13",
-        "content": "The `evaluation details` structure's `variant` field SHOULD contain a identifier logically corresponding to the return flag value.",
-        "RFC 2119 keyword": "SHOULD"
+        "content": "In cases of normal execution, the `evaluation details` structure's `variant` field MUST contain the value of the `variant` field in the `flag resolution` structure returned by the configured `provider`, if the field is set.",
+        "RFC 2119 keyword": "MUST"
     },
     {
         "id": "1.14",
         "clean id": "1_14",
-        "content": "The `evaluation details` structure's `reason` field SHOULD indicate the semantic reason for the returned flag value. Possible values vary by provider, but may include `\"TARGETING_MATCH\"`, `\"SPLIT\"`, `\"DISABLED\"`, `\"DEFAULT\"`, `\"UNKNOWN\"` and `\"ERROR\"`.",
-        "RFC 2119 keyword": "SHOULD"
+        "content": "In cases of normal execution, the `evaluation details` structure's `reason` field MUST contain the value of the `reason` field in the `flag resolution` structure returned by the configured `provider`, if the field is set.",
+        "RFC 2119 keyword": "MUST"
     },
     {
         "id": "1.15",
         "clean id": "1_15",
-        "content": "In cases of normal execution, the `evaluation details` structure's `error code` field MUST not be set, or otherwise must contain a null or falsy value.",
+        "content": "In cases of abnormal execution, the `evaluation details` structure's `error code` field MUST identify an error occurred during flag evaluation, having possible values `\"PROVIDER_NOT_READY\"`, `\"FLAG_NOT_FOUND\"`, `\"PARSE_ERROR\"`, `\"TYPE_MISMATCH\"`, or `\"GENERAL\"`.",
         "RFC 2119 keyword": "MUST"
     },
     {
         "id": "1.16",
         "clean id": "1_16",
-        "content": "In cases of abnormal execution, the `evaluation details` structure's `error code` field SHOULD identify an error occurred during flag evaluation, with possible values `\"PROVIDER_NOT_READY\"`, `\"FLAG_NOT_FOUND\"`, `\"PARSE_ERROR\"`, `\"TYPE_MISMATCH\"`, or `\"GENERAL\"`.",
+        "content": "In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` SHOULD indicate an error.",
         "RFC 2119 keyword": "SHOULD"
     },
     {
         "id": "1.17",
         "clean id": "1_17",
-        "content": "In cases of abnormal execution (network failure, unhandled error, etc) the `reason` field in the `evaluation details` SHOULD indicate an error.",
-        "RFC 2119 keyword": "SHOULD"
+        "content": "The `evaluation options` structure's `hooks` field denotes a collection of hooks that the client MUST execute for the respective flag evaluation, in addition to those already configured.",
+        "RFC 2119 keyword": "MUST"
     },
     {
         "id": "1.18",
         "clean id": "1_18",
-        "content": "The `evaluation options` structure's `hooks` field denotes a collection of hooks to be executed for the respective flag evaluation.",
-        "RFC 2119 keyword": null
+        "content": "Methods, functions, or operations on the client MUST NOT throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.",
+        "RFC 2119 keyword": "MUST NOT"
     },
     {
         "id": "1.19",
         "clean id": "1_19",
-        "content": "No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.",
-        "RFC 2119 keyword": null
+        "content": "In the case of abnormal execution, the client SHOULD log an informative error message.",
+        "RFC 2119 keyword": "SHOULD"
     },
     {
         "id": "1.20",
         "clean id": "1_20",
-        "content": "In the case of abnormal execution, the client SHOULD log an informative error message.",
+        "content": "The `client` SHOULD provide asynchronous or non-blocking mechanisms for flag evaluation.",
         "RFC 2119 keyword": "SHOULD"
     },
     {
         "id": "1.21",
         "clean id": "1_21",
-        "content": "The `client` SHOULD provide asynchronous or non-blocking mechanisms for flag evaluation.",
-        "RFC 2119 keyword": "SHOULD"
+        "content": "The `client` MUST transform the `evaluation context` using the `provider's` `context transformer` function, before passing the result of the transformation to the provider's flag resolution functions.",
+        "RFC 2119 keyword": "MUST"
     },
     {
         "id": "Condition 1.8",
@@ -120,14 +120,8 @@
             {
                 "id": "1.8.1",
                 "clean id": "1_8_1",
-                "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure.```// example boolean flag evaluationboolean myBool =  client.getBooleanValue('bool-flag', false);// example overloaded string flag evaluation with optional paramsstring myString = client.getStringValue('string-flag', 'N/A', evaluationContext, options);// example number flag evaluationnumber myNumber = client.getNumberValue('number-flag', 75);// example overloaded structure flag evaluation with optional paramsMyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);```See [evaluation context](../evaluation-context/evaluation-context.md) for details.",
+                "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure.",
                 "RFC 2119 keyword": "MUST"
-            },
-            {
-                "id": "1.8.2",
-                "clean id": "1_8_2",
-                "content": "The `client` SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.",
-                "RFC 2119 keyword": "SHOULD"
             }
         ]
     },
--- a/specification/flag-evaluation/flag-evaluation.md
+++ b/specification/flag-evaluation/flag-evaluation.md
@@ -6,7 +6,7 @@
 
 The `evaluation API` allows for the evaluation of feature flag values, independent of any flag control plane or vendor. In the absence of a [provider](../provider/providers.md) the `evaluation API` uses the "No-op provider", which simply returns the supplied default flag value.
 
-#### API Initialization and Configuration
+### API Initialization and Configuration
 
 ##### Requirement 1.1
 
@@ -62,7 +62,7 @@ OpenFeature.getClient({
 
 The name is a logical identifier for the client.
 
-#### Client Usage
+### Client Usage
 
 ##### Requirement 1.6
 
@@ -93,26 +93,28 @@ var myValue = client.getValue('my-flag', false);
 > ##### Conditional Requirement 1.8.1
 >
 > > The `client` **MUST** provide methods for typed flag evaluation, including boolean, numeric, string, and structure.
->
-> > ```
-> > // example boolean flag evaluation
-> > boolean myBool =  client.getBooleanValue('bool-flag', false);
-> >
-> > // example overloaded string flag evaluation with optional params
-> > string myString = client.getStringValue('string-flag', 'N/A', evaluationContext, options);
-> >
-> > // example number flag evaluation
-> > number myNumber = client.getNumberValue('number-flag', 75);
-> >
-> > // example overloaded structure flag evaluation with optional params
-> > MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);
-> > ```
->
+
+```
+// example boolean flag evaluation
+boolean myBool =  client.getBooleanValue('bool-flag', false);
+
+// example overloaded string flag evaluation with optional params
+string myString = client.getStringValue('string-flag', 'N/A', evaluationContext, options);
+
+// example number flag evaluation
+number myNumber = client.getNumberValue('number-flag', 75);
+
+// example overloaded structure flag evaluation with optional params
+MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, evaluationContext, options);
+```
+
 > > See [evaluation context](../evaluation-context/evaluation-context.md) for details.
 >
 > ##### Conditional Requirement 1.8.2
 >
 > > The `client` **SHOULD** guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.
+
+#### Detailed Flag Evaluation
 
 ##### Requirement 1.9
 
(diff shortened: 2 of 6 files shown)


specification/flag-evaluation/flag-evaluation.md
```diff
@@ -151,21 +151,19 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
 
 ##### Requirement 1.15
 
-> In cases of normal execution, the `evaluation details` structure's `error code` field **MUST** not be set, or otherwise must contain a null or falsy value.
+> In cases of normal execution, the `evaluation details` structure's `error code` field **MUST** contain the value of the `error code` field in the `flag resolution` structure returned by the configured `provider`.
 
 ##### Requirement 1.16
 
-> In cases of abnormal execution, the `evaluation details` structure's `error code` field **SHOULD** identify an error occurred during flag evaluation, with possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+> In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** identify an error occurred during flag evaluation, having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
```
toddbaert, line 166: Some minor changes here. the main idea is that the evaluation layer should basically be reflecting the data/errors coming up from the provider.

Changed 1.16 to a  "MUST" because the evaluation layer MUST never throw, so it's up to it to decide on an the error code and populate it, even if the provider fails to.

specification/glossary.md
```diff
@@ -74,6 +78,18 @@ An SDK-compliant secondary function that is abstracted by the Feature Flag API,
...
+The process of retrieving a feature flag value in it's entirety, including:
+
+- any effects resulting from hooks
+- resolving a flag value from a configured provider
+- falling back to a supplied default, in the case of abnormal execution
+
+### Flag Value Resolution
+
+The process of a provider retrieving a feature flag value from it's particular source-of-truth.
+
```
toddbaert, line 92: I think this is a good distinction, but let me know if you think I'm being too pedantic. I think it could be a good basic for knowing how to name methods/functions.

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+
+> In cases of normal execution, the `provider` **SHOULD** populate the `flag resolution` structure's `variant` field with a string identifier corresponding to the returned flag value.
+
+For example, the flag value might be `3.14159265359`, and the variant field's value might be `"pi"`.
+
+The value of the variant field might only be meaningful in the context of the flag management system associated with the provider. For example, the variant may be a UUID corresponding to the variant in the flag management system, or an index corresponding to the variant in the flag management system.
+
+##### Requirement 2.6
+
+> In cases of normal execution, the `provider` **SHOULD** populate the `flag resolution` structure's `reason` field with a string indicating the semantic reason for the returned flag value.
```
justinabrahms, line 68:
```suggestion
> The `provider` **SHOULD** populate the `flag resolution` structure's `reason` field with a string indicating the semantic reason for the returned flag value.
```

There's a provision for `ERROR` in the reason, so they should put that in there for both cases, yes?

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+
+```
+// example boolean flag value resolution
+FlagResolution resolveBooleanValue(string flagKey, boolean defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
+
+// example string flag value resolution
+FlagResolution resolveStringValue(string flagKey, string defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
+
+// example number flag value resolution
+FlagResolution resolveNumberValue(string flagKey, number defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
```
justinabrahms, line 44: condition around number -> (long,int) ?

toddbaert: With the last PR, I added https://github.com/open-feature/spec/blob/main/specification/types.md#number, which I think covers this problem in a general way, at least that was the hope.

specification/provider/providers.md
toddbaert: Yes!

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -192,3 +190,9 @@ Implementations may define a standard logging interface that can be supplied as
 > The `client` **SHOULD** provide asynchronous or non-blocking mechanisms for flag evaluation.
 
 It's recommended to provide non-blocking mechanisms for flag evaluation, particularly in languages or environments wherein there's a single thread of execution.
+
+##### Requirement 1.22
+
+> The `client` **MUST** transform the `evaluation context` using the `provider's` `context transformer` function, before passing the result of the transformation to the provider's flag resolution functions.
```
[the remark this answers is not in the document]
toddbaert: The client actually RUNS the transformer (by default a no-op), before passing that transformed context to the provider's resolution methods. That's what I was trying to capture here.

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -151,21 +151,19 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
 
 ##### Requirement 1.14
 
-> The `evaluation details` structure's `reason` field **SHOULD** indicate the semantic reason for the returned flag value. Possible values vary by provider, but may include `"TARGETING_MATCH"`, `"SPLIT"`, `"DISABLED"`, `"DEFAULT"`, `"UNKNOWN"` and `"ERROR"`.
+> In cases of normal execution, the `evaluation details` structure's `reason` field **MUST** contain the value of the `reason` field in the `flag resolution` structure returned by the configured `provider`.
 
 ##### Requirement 1.15
 
-> In cases of normal execution, the `evaluation details` structure's `error code` field **MUST** not be set, or otherwise must contain a null or falsy value.
+> In cases of normal execution, the `evaluation details` structure's `error code` field **MUST** contain the value of the `error code` field in the `flag resolution` structure returned by the configured `provider`.
```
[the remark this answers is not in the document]
toddbaert: Ya, this one is a bit tricky - maybe it's an edge case. I was thinking of a scenario wherein the provider didn't throw or otherwise abnormally terminate, but set the error. Maybe we dont need to define such a situation.

specification/flag-evaluation/flag-evaluation.json
```diff
@@ -62,26 +62,26 @@
     {
         "id": "1.13",
         "clean id": "1_13",
-        "content": "The `evaluation details` structure's `variant` field SHOULD contain a identifier logically corresponding to the return flag value.",
-        "RFC 2119 keyword": "SHOULD"
+        "content": "In cases of normal execution, the `evaluation details` structure's `variant` field MUST contain the value of the `variant` field in the `flag resolution` structure returned by the configured `provider`.",
+        "RFC 2119 keyword": "MUST"
```
agentgonzo, line 66: can you remind me of the difference between the flag value (1.10) and the variant value here?

specification/flag-evaluation/flag-evaluation.json
```diff
@@ -62,26 +62,26 @@
...
     },
     {
         "id": "1.14",
         "clean id": "1_14",
-        "content": "The `evaluation details` structure's `reason` field SHOULD indicate the semantic reason for the returned flag value. Possible values vary by provider, but may include `\"TARGETING_MATCH\"`, `\"SPLIT\"`, `\"DISABLED\"`, `\"DEFAULT\"`, `\"UNKNOWN\"` and `\"ERROR\"`.",
-        "RFC 2119 keyword": "SHOULD"
+        "content": "In cases of normal execution, the `evaluation details` structure's `reason` field MUST contain the value of the `reason` field in the `flag resolution` structure returned by the configured `provider`.",
+        "RFC 2119 keyword": "MUST"
```
agentgonzo, line 72: Do all providers give a reason? If not, you're going to get a lot of default, or null reasons here.

specification/flag-evaluation/flag-evaluation.json
```diff
@@ -62,26 +62,26 @@
...
     },
     {
         "id": "1.15",
         "clean id": "1_15",
-        "content": "In cases of normal execution, the `evaluation details` structure's `error code` field MUST not be set, or otherwise must contain a null or falsy value.",
+        "content": "In cases of normal execution, the `evaluation details` structure's `error code` field MUST contain the value of the `error code` field in the `flag resolution` structure returned by the configured `provider`.",
```
agentgonzo, line 77: Errr... in normal execution, there shouldn't be an error!

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
agentgonzo: I think it's an unnecessary requirement. It's implicit that the provider must transform the content.

specification/glossary.md
```diff
@@ -62,9 +62,13 @@ Application authors use it via the Feature Flag SDK.
 
 The subset of the [Feature Flag API](#feature-flag-api) that the Application Author uses to evaluate flags.
 
+### Flag Management System
+
+A source-of-truth for flag values and rules. Flag management systems may include Saas feature flag vendors, custom "in-house" feature flag infrastructure, or open-source implementations.
```
agentgonzo, line 67:
```suggestion
A source-of-truth for flag values and rules. Flag management systems may include SaaS feature flag vendors, custom "in-house" feature flag infrastructure, or open-source implementations.
```

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
-# Providers
+# Provider
 
 ## Overview
 
-// TODO: this is a stub.
+The `provider` API defines interfaces that Provider Authors can use to abstract a particular flag management system, thus enabling the use of the `evaluation API` by Application Authors.
```
agentgonzo, line 5:
```suggestion
The `provider` API defines interfaces that Provider authors can use to abstract a particular flag management system, thus enabling the use of the `evaluation API` by Application Authors.
```

agentgonzo: unless you want to define "Provider Authors" as a specific actor, I would lowercase this, otherwise it looks like it's a Named Thing.

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+
+> In cases of abnormal execution, the `provider` **MUST** populate the `flag resolution` structure's `value` field with the default flag value.
+
+##### Requirement 2.5
+
+> In cases of normal execution, the `provider` **SHOULD** populate the `flag resolution` structure's `variant` field with a string identifier corresponding to the returned flag value.
+
+For example, the flag value might be `3.14159265359`, and the variant field's value might be `"pi"`.
+
+The value of the variant field might only be meaningful in the context of the flag management system associated with the provider. For example, the variant may be a UUID corresponding to the variant in the flag management system, or an index corresponding to the variant in the flag management system.
```
agentgonzo, line 64: This (if I understand it) conflicts with the flag evaluation 1.14 above which moved to a MUST.

I am not certain that all providers will have a variant name, and so for the moment I think that we should keep the variant value as a SHOULD rather than a MUST. I may have misunderstood though.

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+
+##### Requirement 2.6
+
+> The `provider` **SHOULD** populate the `flag resolution` structure's `reason` field with a string indicating the semantic reason for the returned flag value.
+
+Possible values vary by provider, but might include such values as `"TARGETING_MATCH"`, `"SPLIT"`, `"DISABLED"`, `"DEFAULT"`, `"UNKNOWN"` or `"ERROR"`.
+
+##### Requirement 2.7
+
+> In cases of normal execution, the `provider` **MUST NOT** populate the `flag resolution` structure's `error code` field, or otherwise must populate it with a null or falsy value.
```
agentgonzo, line 74: 👍 (but again in conflict with what you have above)

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+
+##### Requirement 2.8
+
+> In cases of abnormal execution, the `provider` **MUST** indicate an error using the idioms of the implementation language, with an associated error code having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+
+The provider might throw an exception, return an error, or populate the `error code` object on the returned `flag resolution` structure to indicate a problem during flag value resolution.
```
agentgonzo, line 80: I thought that we didn't want the provider throwing an error?

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+  //...
+
+  // implementation of context transformer
+  MyProviderContext transformContext(EvaluationContext context) {
+    return new MyProviderContext(context.email, context.ip, context.httpMethod);
+  }
+
+  //...
+}
+```
```
agentgonzo, line 127: See above. I think this is unnecessary as it's implicit that the provider has to transform the content from "what openfeature gives it" into "what the provider's SDK actually needs"

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: @agentgonzo @justinabrahms 

> I think it's an unnecessary requirement. It's implicit that the provider must transform the content.

The provider must transform the context, yes, but if a context transform is defined, it's actually not the provider that calls that method. See the example for the [launchdarkly experimental provider](https://github.com/open-feature/sdk-research/blob/main/packages/js-launchdarkly-provider/src/lib/js-launchdarkly-provider.ts#L26-L34)

Here a context transformer is defined. It just sets a user ID, and puts all the custom attributes into a `custom` field, as the LaunchDarkly API dictates. Though the provider defines this function, it never runs it. The caller (the OpenFeature client) actually does before calling the flag resolution methods. This is nice, because you can see the flag resolution methods now take the transformed context, with the correct type as the parameter (`LDUser` supplied by the generic arg to the class): https://github.com/open-feature/sdk-research/blob/main/packages/js-launchdarkly-provider/src/lib/js-launchdarkly-provider.ts#L71-L88

This gives us a bit type safety through generics, and avoids calling transform code in every flag resolution method.

specification/flag-evaluation/flag-evaluation.json
toddbaert: Many do not, some do, this is an optional field. In many cases it will be missing, null, or undefined.

specification/flag-evaluation/flag-evaluation.json
toddbaert: See discussion here: https://github.com/open-feature/spec/pull/41#discussion_r860195055, duplicated.

specification/provider/providers.md
toddbaert: It's a term in the glossary: https://github.com/open-feature/spec/blob/main/specification/glossary.md#provider-author

specification/provider/providers.md
toddbaert: Most of the sample providers throw errors, consistent with the "throw early, catch late" mantra. They are all caught in the client, which calls the providers (providers are not directly accessed by application authors), and must never throw.

specification/provider/providers.md
toddbaert: I was thinking it would populate the field with whatever the provider returned, but that could absolutely include a null or falsey or undefined value. Is that confusing?

specification/provider/providers.md
toddbaert: See my other comment here: https://github.com/open-feature/spec/pull/41#discussion_r860825935

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: I will remove it entirely.

specification/provider/providers.md
toddbaert: I will remove `1.15` since it seem to cover a very unusual use-case and caues confusion (I think that's what you're referencing).

specification/flag-evaluation/flag-evaluation.json
agentgonzo: Ah, I understand now. In which case, I think this is confusing as I read 'MUST' to imply that it must return a value. Perhaps tweak the wording
```suggestion
        "content": "In cases of normal execution, the `evaluation details` structure's `reason` field MUST contain the value of the `reason` field in the `flag resolution` structure returned by the configured `provider` if the `provider` returns it (if the `provider` does not give a reason, then it does not need to be set.",
        "RFC 2119 keyword": "MUST"
```

Or some better language

specification/provider/providers.md
agentgonzo: I think it is. I think we have some 'MUST' clauses that imply it should _always_ set a value, rather than "it should set a value if the provider has one". I think a tweak to the laungage will make it more obvious (but if it's just me that is confused then it's fine to continue as you are)

toddbaert: I agree with you. I will tweak.

specification/flag-evaluation/flag-evaluation.json
justinabrahms: This is like the name of the underlying value. Like you might have a giant json blob, but this is the ID or hash or descriptor of it.

specification/glossary.md
justinabrahms: It's a spec. More things in the glossary are awesome. :D

specification/flag-evaluation/flag-evaluation.json
toddbaert: This is an optional field. I think this is another case where I was trying to simply say that the client should pass the value returned from provider, which may or may not be set. I will add clarification here.

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
moredip: I think it's nice for us to give something like the transformer to help simplify the provider implementation, but I also don't think it should be part of the spec. It really feels like more of an implementation detail. Maybe there's something we could put into the spec that's defines this as a suggested patterns for people implementing a client.

specification/glossary.md
moredip: This is a bit nit-picky, but both Evaluation and Resolution suffer from being both a verb and a noun in English. To clarify that this is a verb, maybe we could use "Evaluating a Flag" and "Resolving a Flag Value" instead?

toddbaert: This is exactly the kind of nit-pick I love.

specification/glossary.md
```diff
@@ -62,9 +62,13 @@ Application authors use it via the Feature Flag SDK.
...
 The subset of the [Feature Flag API](#feature-flag-api) that the Application Author uses to evaluate flags.
 
+### Flag Management System
+
+A source-of-truth for flag values and rules. Flag management systems may include SaaS feature flag vendors, custom "in-house" feature flag infrastructure, or open-source implementations.
+
 ### Provider
 
-An SDK-compliant feature flag implementation which adheres to the Feature Flag API. Implementations may include Saas feature flag vendors, custom "in-house" feature flag infrastructure, or open-source implementations.
+An implementation which resolves flag values from a particular flag management system, allowing the use of the [Evaluation API](./flag-evaluation/flag-evaluation.md#flag-evaluation) as an abstraction for the system in question.
```
moredip, line 71: seems like we should explicitly say that a provider has to be spec-conforming.

moredip:
Do we want to specify any semantics around how a provider should handle transient errors on their side. Like if there's a network connectivity issue, should to provider immediately fail, or have a retry/fallback mechanism. 

One option is to leave this entirely unspecified and up to each provider. Another would be to give a provider implementer some guidance or suggestions - e.g. make a best effort to handle transient failure, but don't hang evaluation for "a long time" (whatever that means).

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
justinabrahms: I like a may. That's exactly what may is for. "Hey, we think this might be a good idea. Why don't you do that if you want." Either works for me though.

specification/provider/providers.md
```diff
@@ -1,5 +1,159 @@
...
+FlagResolution resolveStructureValue(string flagKey, JsonObject defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
+```
+
+##### Requirement 2.4
+
+> In cases of normal execution, the `provider` **MUST** populate the `flag resolution` structure's `value` field with the resolved flag value.
+
+##### Requirement 2.4
+
+> In cases of abnormal execution, the `provider` **MUST** populate the `flag resolution` structure's `value` field with the default flag value.
```
moredip, line 56: In Requirement 2.8 we say that a provider is allowed to throw during abnormal execution. Saying they MUST also populate the default value seems to contradict that somewhat.

specification/provider/providers.md
moredip: I think it might be worth mentioning that here - "a provider might throw an exception (which will always be caught within the client and not surfaced to application authors)" or something.

specification/types.md
```diff
@@ -30,7 +30,16 @@ A structure containing the following fields:
 - reason (string, optional)
 - variant (string, optional)
 
-### evaluation options
+### Flag Resolution
```
moredip, line 33: again, Flag Resolution is ambiguously either a verb or a noun. Would Evaluation Result or Evaluation Outcome be better?

specification/flag-evaluation/flag-evaluation.md
toddbaert: I think that needs to be considered. If we want to allow authors to actually be able to write conditional logic around these errors, they need to be rigorously defined in the API (but then we'll lose fidelity mapping them from the underlying SDKs). If we only need errors for debugging, then they can be more free-form. It may merit further discussion.

toddbaert:
> Do we want to specify any semantics around how a provider should handle transient errors on their side. Like if there's a network connectivity issue, should to provider immediately fail, or have a retry/fallback mechanism.
> 
> One option is to leave this entirely unspecified and up to each provider. Another would be to give a provider implementer some guidance or suggestions - e.g. make a best effort to handle transient failure, but don't hang evaluation for "a long time" (whatever that means).

This might be relevant to the "event" discussion. Most SDKs have readiness events, etc. We could build an events API which might cover both cases.

I think for now, I would err on the side of not defining this until we need to - but I'm not sure. If there's a pressing use-case, we can talk more about it.

specification/provider/providers.md
toddbaert: You're right, this is really the job of the client (caller) in this case.

specification/types.md
toddbaert: How about `ResolutionDetails`? It's consistent with `EvaluationDetails`, and in the context of this type, I don't actually know if the verb/noun confusion is really as problematic. Whether you think of it as a verb or a noun, these are the details associated with it.

I'm going to updated it with `ResolutionDetails`. I'm totally open to changing the term "resolution" entirely. I'd simply like to have a term that denotes the solely the aspect of the flag evaluation that the provider does (in opposition to the entire evaluation which involves hooks, the client, etc).

specification/flag-evaluation/flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: I've made everything related to the context transformer a "MAY".

specification/types.md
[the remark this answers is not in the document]
toddbaert: Updated.

toddbaert: Updated.

moredip:
> @moredip any additional concerns after my updates to your comments?

Nope, I think it's good

beeme1mr (approved):
It looks like the major issues have been addressed now. I think it's time to merge it as is since nothing here is final. Any issues can be handled with a tickets and/or a PR.

toddbaert:
Squashed, added co-authors.

Issue #36, opened 2022-04-25
CI Processes for spec


toddbaert:
I'm interested in creating some CI processes for the spec repo. Some functionality I think would be helpful:

- [x] linting markdown, specifically in according with the style-guide (https://www.npmjs.com/package/markdownlint)
- [x] ensuring links between markdown are valid (https://www.npmjs.com/package/markdown-link-check)
- [x] generate the `RFC 2119` assertion JSON files automatically (currently manual), and run some basic validation on them.
- [x] validate ToCs?
- [X] JSON file validation (thx @justinabrahms)
- [X] check JSON assertion files has been done (thx @dyladan)

I think we can do all these with some simple packages, enhancements to the Makefile, and github actions.

PR #45, opened 2022-04-30
First pass at a hooks definition.

justinabrahms:
I'm explicitly not structuring this into the requirements language at first. I'd like to get some general feedback on the content before I fight with the spec parser. I also want to avoid the situation where some folks comment on the json and others on the text. :)


specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> Immutable hook data **MUST** be passed to each hook through a parameter. It is merged into the object in the precedence order API -> Client -> Invocation.
+
+> The hook **MUST** not alter the `immutable state` object.
+
+### Hook evaluation
+
+> `Immutable state` **MUST** passed between each hook.
+
+> As hooks are run, they **MUST** be added to an `evaluatedHooks` property within the `EvaluationContext`.
```
beeme1mr, line 55: There was some debate on whether or not this was a good idea to specify in the spec.

https://github.com/open-feature/spec/pull/35#discussion_r855316168

We concluded that it's a good idea but not necessary at this stage of the spec. What do you think?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> If a hook throws an error, it **MUST** be in `evaluatedHooks`.
```
beeme1mr, line 57: This requirement isn't clear to me. Could you elaborate what you mean?

justinabrahms: Add the current hook to the evaluated hooks before running it so, if it throws an error, you see it in evaluated hooks. 

Do:
```
Evaluated hooks.add(h)
h.run()
```

Not:
```
H.run()
Evaluated hooks.add(H)
```

Does that make sense?

specification/flag-evaluation/hooks.md
justinabrahms: I'm game to drop it for now and see if the need pops up before GA.

specification/flag-evaluation/hooks.md
toddbaert: I think this is a bit redundant, it's implicit in the original statement. But I agree we don't need it now.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
 
 ## Overview
 
-// TODO: this is a stub.
+Hooks are a mechanism which can tie into the lifecycle of flag evaluation. They
+operate similarly to middleware in many web frameworks.
+
```
toddbaert, line 7: "tie in" is vague. I suggest something like "hooks are a mechanism by which arbitrary behavior (or maybe "logic" if you prefer) to flag evaluation".

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+### Hook context
+
+Hook context exists to provide hooks with information about the invocation.
+
+> Hook context **MUST** provide: the flag key, flag type, evaluation context, default value and a list of  executed hooks by stage.
```
toddbaert, line 12: > flag type may not be possible in some languages. You may want to make this conditional, like some of the other sections: 

> The language type system differentiates between strings, numbers, booleans and structures.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> Hook context **SHOULD** provide: provider, client
+
+> flag key, flag type, default value properties **MUST** be immutable.
+
+> The evaluation context **MUST** be mutable only within the `before` hook.
```
toddbaert, line 18: I like both of these ideas, but I'm not sure all languages have immutability concepts, so you might want to make this conditional to available language mechanisms.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+
+### Hook creation and parameters
+
+> Hooks **MUST** specify at least one stage: `before(HookContext, ImmutableState)`, `after(HookContext, FlagEvaluationDetails, , ImmutableState)`, `error(HookContext, Exception, ImmutableState)`, or `finally(HookContext, ImmutableState)`.
+
+> Condition: If `finally` is a reserved word in the language, `afterAll` should be used.
```
toddbaert, line 25: "stage" is a good term for this.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+### Hook registration & ordering
+
+> The API, Client and invocation **MUST** have a method for registering hooks which accepts `flag evaluation options`
+
+> Hooks **MUST** be evaluated in the following order: API before, client before, invocation before, invocation after, client after, API after, invocation finally, client finally, API finally. If errors are present, the order is: invocation error, client error, API error.
```
toddbaert, line 31: We've used "invocation" here a few times, and I think it may be good to explain that "invocation" here is specifically a "flag evaluation invocation". Maybe when you format this, you could do this in a supplemental bit of text, not in a normative spec requirement.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> When an error occurs in hook evaluation, further evaluation **MUST** stop and error hook evaluation begins.
+
+> If an error is encountered in the error flow, it **MUST** not be returned to the user.
```
toddbaert, line 35: Throwing from an error stage is something we said could be a possible exception to the "never throw" rule. If we want this, we can make such an exception in the `flag evaluation` spec. I think it has some utility.

@beeme1mr @agentgonzo @moredip what do you guys think?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> If an error occurs in the `before` or `after` hooks, the `error` hooks **MUST** be invoked.
+
+> If an error occurs in the `finally` hook, it **MUST** not trigger the `error` hook.
+
+### Flag evaluation options
```
toddbaert, line 41: You may want to link to the `types.md` file here: https://github.com/open-feature/spec/blob/062a083702a442aee184e2838c90911eb8b778b5/specification/types.md

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> `Flag evalution options` must contain a list of hooks to evaluate.
+
+> `Flag evaluation options` may contain `immutable state`, a map of data to be provided to hook invocations.
```
toddbaert, line 45: Though the state is immutable, I think I'd prefer not to see that detail be part of the interfaces, especially since not all languages support immutable structures. I think `state` is fine, I've seen such things in other APIs.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> Immutable hook data **MUST** be passed to each hook through a parameter. It is merged into the object in the precedence order API -> Client -> Invocation.
```
toddbaert, line 47:
```suggestion
> The `immutable state` **MUST** be passed to each hook through a parameter. It is merged into the object in the precedence order API -> Client -> Invocation.
```

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+> The hook **MUST** not alter the `immutable state` object.
```
toddbaert, line 49: The way this is phrased seems like it targets the hook implementer, not the SDK implementer. I think it should instead say that the `immutable state` should ideally be an immutable object, if the language has such concepts (perhaps use a condition).

toddbaert (commented):
Seems like a very good first pass, I left a few comments and look forward to the full version.

The parser script is a bit finicky - it could benefit from some improvement on edge-cases - let me know if you have issues with it.

Also, once https://github.com/open-feature/spec/issues/36 is done, we should be able to .gitignore all the JSON files and do some nice automation.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+
+### Hook creation and parameters
+
+> Hooks **MUST** specify at least one stage: `before(HookContext, ImmutableState)`, `after(HookContext, FlagEvaluationDetails, , ImmutableState)`, `error(HookContext, Exception, ImmutableState)`, or `finally(HookContext, ImmutableState)`.
+
+> Condition: If `finally` is a reserved word in the language, `afterAll` should be used.
+
+### Hook registration & ordering
+
+> The API, Client and invocation **MUST** have a method for registering hooks which accepts `flag evaluation options`
```
toddbaert, line 29: is "API" or "global" a better descriptor here?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,58 @@
...
+> Hook context **SHOULD** provide: provider, client
+
+> flag key, flag type, default value properties **MUST** be immutable.
+
+> The evaluation context **MUST** be mutable only within the `before` hook.
+
+
+### Hook creation and parameters
+
+> Hooks **MUST** specify at least one stage: `before(HookContext, ImmutableState)`, `after(HookContext, FlagEvaluationDetails, , ImmutableState)`, `error(HookContext, Exception, ImmutableState)`, or `finally(HookContext, ImmutableState)`.
```
toddbaert, line 23:
```suggestion
> Hooks **MUST** specify at least one stage: `HookContext | void before(HookContext, ImmutableState)`, ` FlagEvaluationDetails | void after(HookContext, FlagEvaluationDetails, , ImmutableState)`, `void error(HookContext, Exception, ImmutableState)`, or `void finally(HookContext, ImmutableState)`.
```

toddbaert: I added the return values (which I think are fairly important).

I wonder if it's better to use plan prose instead of pseudo-code to describe the parameters and return values.

toddbaert: > Would be interested to hear what FlagEvaluationDetails return value is used for.

The idea was you could use it for for a few interesting things in after. For example. maybe you want to turn some JSON flag value into a class instance - you could mutate the `.value` field to actually return an instance - the hook would be a little factory that calls a constructor.

specification/flag-evaluation/hooks.md
toddbaert: That works for me.

specification/flag-evaluation/hooks.md
justinabrahms: @toddbaert This would imply that subsequent hooks couldn't assume that the flag value is the type that the flag says it is, right?

specification/flag-evaluation/hooks.md
justinabrahms: We would still know what type was requested, right?  Like they will have to call `getBooleanValue`, no?

toddbaert: > We would still know what type was requested, right? Like they will have to call getBooleanValue, no?

"getBooleanValue" only exists in languages implementations that have types. A completely type-less language would just implement `getValue`

specification/flag-evaluation/hooks.md
toddbaert: No, but we might support an generic arg in order to provide that type information in a hook definition. I'm not 100% convinced we need this feature, but it seems like it could be useful.

justinabrahms: https://github.com/open-feature/spec/issues/48

specification/flag-evaluation/hooks.md
agentgonzo: If we're going to have a rule of "never throw", then I don't think we should start adding exceptions to it.

IMO, a hook is a non-critical part of the evaluation flow (the critical part being the flag evaluation/resolution itself). If you throw an error, it means that the `getXyzValue()` won't return an evaluated flag value to the user (bad) IIUI.

agentgonzo: Is it important to actually return an error in any of the hooks to the caller? What would they do with such an error? Would it be sufficient to merely log the error in the provider?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,75 @@
...
+
+> The `finally` hook **MUST** run after the `before`, `after`, and `error` stages. It accepts a `hook context` (required) and `state` (optional). There is no return value.
+
+> Condition: If `finally` is a reserved word in the language, `afterAll` should be used.
+
+### Hook registration & ordering
+
+> The API, Client and invocation **MUST** have a method for registering hooks which accepts `flag evaluation options`
+
+> Hooks **MUST** be evaluated in the following order: API before, client before, invocation before, invocation after, client after, API after, invocation finally, client finally, API finally. If errors are present, the order is: invocation error, client error, API error.
```
InTheCloudDan, line 54: To me on initial pass this appears to be a lot of hook stages. Have we actually verified that there is use cases for each one? In my view it'll be cleaner to start with few hook stages and add them as needed vs having a large number upfront and finding it's becoming a maintenance or implementation headache.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,75 @@
...
+
+> Hook context **MUST** provide: the flag key, flag type, evaluation context, default value and a list of  executed hooks by stage.
+
+> Hook context **SHOULD** provide: provider, client
+
+> flag key, flag type, default value properties **MUST** be immutable.
+
+> The evaluation context **MUST** be mutable only within the `before` hook.
+
+### State
```
InTheCloudDan, line 26: What is the exact use case of `State`? One question I have is does it risk causing flags to evaluate differently on different instances of a service?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,75 @@
...
+
+> State **MUST** be a map of objects.
+
+> Condition: If your language supports it, state **MUST** be immutable.
```
beeme1mr, line 30:
```suggestion
> Condition: State **MUST** be immutable, if the implementation language supports a mechanism for marking data as immutable.
```

specification/flag-evaluation/hooks.md
toddbaert: I see your concern. It think it looks like there's a lot more complexity here than there actually is. The There is really only: `before`, `after`, `error`, and `finally` - it's just that there's 3 places to attach hooks, and so there's a need to define how they interact.

In code, there's not too much complexity here (I hope). From the Application Author's perspective, it's just:

```
OpenFeature.addHooks(new Hook1);

//...

Client client = OpenFeature.getClient();
client.addHooks(new Hook2);

//...

client.getValue('my-flag', 'defaultValue', new Hook3);
```

We are defining the order of these hooks relative to eachother, which isn't too tough in implementation logic either. It looks something like this:

```
// client logic...

// here we are using the correct order (note if it wasn't defined, we'd have to order them somewhow anyway
Hooks[] allHooks = [ ...OpenFeature.getHooks(), ...this.getHooks, ...evaluationOptions.hooks ]

// evaluation all our before hooks
for (h in allHooks) {
  h.before();
}

provider.resolveValue(...);

// evaluation all our after hooks
for (h in allHooks.reverse()) {
  h.after()
}

// more client logic...
```

Let me know if that doesn't address your concern. It might be that there's an easier way to describe this, maybe with a point for list, or with prose. I'm not sure, but I get the point @justinabrahms was trying to make.

InTheCloudDan: @toddbaert thanks for the example in code. That does clear it up.

toddbaert: You're welcome! If anything, this might mean that when we do the full markdown spec, adding code examples in non-normative auxiliary sections for this requirement is a good idea.

specification/flag-evaluation/hooks.md
justinabrahms: I'm primarily interested in, on a per-invocation basis, specifying information to a hook. Imagine I have a hook that emits metrics. I want that in almost all cases. I'd like the ability to say "For this particular request.. don't emit metrics".   I'm thinking we use `state` for that.

specification/flag-evaluation/hooks.md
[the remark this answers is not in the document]
justinabrahms: Updated it to be last wins and included a code sample.

specification/flag-evaluation/hooks.md
justinabrahms: Updated to be conditional.

specification/flag-evaluation/hooks.md
InTheCloudDan: That makes sense, I'm wondering if there's a way we can clarify or scope this expected usage down. We're quickly layering on top of the evaluation process, hooks, then state without an initial release.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
 
 ## Overview
 
-// TODO: this is a stub.
+Hooks are a mechanism whereby application developers can add arbitrary behavior
```
moredip, line 5: This is a good opportunity for us to use the User Roles in the glossary. Maybe we could add something like:

They operate similarly to middleware in many web frameworks. Integration Authors create integrations between external systems and the OpenFeature flag evaluation workflow in the form of hooks, which an Application Integrator can then enable within their Feature Flag SDK configuration.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+to flag evaluation. They operate similarly to middleware in many web frameworks.
+
+### Definitions
+
+**Stage**: An explicit portion of the flag evaluation lifecycle. e.g. `before` being "before the evaluation is run.
+**Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)` is an invocation.
+**API**: The global API singleton.
```
moredip, line 12: Could we add a definition for hook itself here?

Hook: a function (callback?) that is called by the OpenFeature framework at a specific invocation stage, with hook context passed in.

specification/flag-evaluation/hooks.md
toddbaert: I'm OK with the concept, and I think I see the value from other discussions with @justinabrahms . It might be that we need an example use-case or other form of explanation (and perhaps to change the name to something that more clearly defines the purpose; "state" may be a bit vague?).

Once this is formatted to a spec document, that supplementary info (and even a code sample, perhaps) can be added in a non-normative section.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+### Hook context
+
+Hook context exists to provide hooks with information about the invocation.
+
+> Hook context **MUST** provide: the flag key, evaluation context, default value and a list of  executed hooks by stage.
+
+> Condition: You **MUST** provide `flag type` if the language type system differentiates between strings, numbers, booleans, and structures.
```
moredip, line 20: not sure this needs to be conditional. Given that the evaluation and provider API requires type-specific methods, wouldn't we always know the flag type regardless of the host language's type system?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+> Hook context **SHOULD** provide: provider, client
```
moredip, line 22: are provider and client an actual runtime instance of the provider and client that the hook could interact with? or just a name? or something else?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+> flag key, flag type, default value properties **MUST** be immutable.
```
moredip, line 24: maybe add something here like "if the language does not support immutability, the hook MUST NOT modify these properties"

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+> State **MUST** be a map of objects.
+
+> Condition: State **MUST** be immutable, if the implementation language supports a mechanism for marking data as immutable.
+
+### Hook creation and parameters
+
+> Hooks **MUST** specify at least one stage.
+
+> The `before` stage *MUST* run before flag evaluation occurs. It accepts a `hook context` (required) and `state` (optional) as parameters and returns either a `HookContext` or nothing.
```
moredip, line 38: given the definitions above, I think it would be more precise to say "*MUST* run before invocation occurs"

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+```
+HookContext|void before(HookContext, State)
+```
+
+> The `after` stage **MUST** run after flag evaluation occurs. It accepts a `hook context` (required), `flag evaluation details` (required) and `state` (optional). It has no return value.
```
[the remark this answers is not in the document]
toddbaert: We could say "flag evaluation invocation".

specification/flag-evaluation/hooks.md
toddbaert: In the demo repo, it's the instances. We have some hooks (OTel for instance) that grabs some metadata from them (such as the provider name, client version, client name, etc). This is a good point.

specification/flag-evaluation/hooks.md
justinabrahms: That seems not quite right to me. The application author writes something like:

```
client.getBooleanValue(...)
```

That's a flag invocation.

Within the framework code, it does something like:

```
runBefore(hooks)
callProviderToGetRealData()
runAfter(hooks)
```

This is trying to say that the `runBefore` function needs to happen before the `callProviderToGetRealData` function.

specification/flag-evaluation/hooks.md
justinabrahms: @beeme1mr I have a very concrete use-case for it. I'd _prefer_ to keep it in this spec, as when it merges I'm going to update the java sdk code to do the thing I'm saying and then I can use the code internally. That said, if you feel strongly, I'll move it out.

The explicit use-case here is I have a metrics service. Sometimes, we want to control precisely when metrics are emitted (due to some pre-fetching logic we have), but usually not. I want to register a "please emit the relevant metrics" hook which triggers by default.

Sometimes, when I want to control when those metrics emit, I want to not have that piece of code run. To do that, I have to provide some state to the function. The options for state that I've considered:

1. We could use `evaluatedHooks`. I could register a `DoNotEmitHooks` hook, which is a sentinel value that `EmitHooks` looks for. If it sees it in the evaluation order before it.. it won't emit hooks. (eew).
2. We could use `EvaluationContext` which is the map of data we send up for evaluation. It could have users put data in there to communicate to the hooks, but isn't really used for targeting. (eew)
3. We could pass some explicit hook-specific state. That's what I'm proposing here.
4. We could call methods on the `EmitMetricsHook` itself like `dontEmitMetricsOnNextCall`, but I worry about the thread safety of this approach.
5. Maybe a secret 5th option that you can think of?

@moredip I hear you about rapidly iterating. This came up when we were trying to port one of the hooks we care about into the janky java sdk I wrote. We didn't actually have the relevant primitives to accomplish the goal, hence the proposal.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+> The `error` hook runs when errors are encountered in the `before` or `after` state or when flag evaluation errors. It accepts `hook context` (required), `exception` for what went wrong (required), and `state` (optional). It has no return value.
```
moredip, line 47: s/state/stage/

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+> The `finally` hook **MUST** run after the `before`, `after`, and `error` stages. It accepts a `hook context` (required) and `state` (optional). There is no return value.
+
+> Condition: If `finally` is a reserved word in the language, `afterAll` should be used.
```
moredip, line 51: afterAll might be a confusing name for people who associate that with testing frameworks. In testing frameworks it has a different semantic meaning - run this hook once after all the tests have finished.

would `eventually` be better? or `alwaysAfter`? or `finallyAfter`?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+Client client = OpenFeature.getClient();
+client.addHooks(new Hook2());
+
+//...
+
+client.getValue('my-flag', 'defaultValue', new Hook3());
+```
+
+> Hooks **MUST** be evaluated in the following order: API before, client before, invocation before, invocation after, client after, API after, invocation finally, client finally, API finally. If errors are present, the order is: invocation error, client error, API error.
```
moredip, line 70: for the errors scenario, I think it would be clearly to specify the full sequence, which I _think_ is:

- API before
- client before
- invocation before
- invocation error
- client error
- API error
- invocation finally
- client finally
- API finally

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+  // evaluation all our finally hooks
+  for (h in allHooks.reverse()) {
+    h.finally(...);
+  }
+}
+```
+
+> When an error occurs in hook evaluation, further evaluation **MUST** stop and error hook evaluation begins.
+
+> If an error is encountered in the error stage, it **MUST** not be returned to the user.
```
moredip, line 103: for consistency, it seems like this should also be the case for errors in the finally stage?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+  }
+} finally {
+  // evaluation all our finally hooks
+  for (h in allHooks.reverse()) {
+    h.finally(...);
+  }
+}
+```
+
+> When an error occurs in hook evaluation, further evaluation **MUST** stop and error hook evaluation begins.
```
moredip, line 101: in this section I was a bit confused about when we were talking about errors in the provider evaluation vs errors in the hook evaluation. It might be clearer if the language was a bit more specific. For example, rather than saying "if an error is encountered in the error stage" we could say "if an error is encountered during evaluation of a hook in the error stage", or just "during evaluation of an error hook"

specification/flag-evaluation/hooks.md
moredip: I found the work "state" a bit misleading. I assumed it was a place for hooks to maintain state that could then be picked up by the hook in a later evaluation stage. I think what's being describe here is more like per-invocation configuration for a hook. Assuming I'm reading that correctly, then maybe "Hook Hints" would be a clearer name than state?

toddbaert: @moredip  Agreed, that's really my primary concern with this requirement as well.

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,126 @@
...
+
+> If an error occurs in the `before` or `after` hooks, the `error` hooks **MUST** be invoked.
+
+> If an error occurs in the `finally` hook, it **MUST** not trigger the `error` hook.
+
+### [Flag evaluation options](../types.md#evaluation-options)
+
+> `Flag evalution options` must contain a list of hooks to evaluate.
+
+> `Flag evaluation options` may contain `state`, a map of data to be provided to hook invocations.
```
moredip, line 113: maybe we'd specify this somewhere else, but I'm left a bit unclear as to how exactly an Application Author would pass this state in to an invocation

specification/flag-evaluation/hooks.md
moredip: I'm still confused as to how we wouldn't have that information, even in a completely type-less language. Why would a type-less language just implement getValue? Wouldn't we still track the type as metadata somewhere? Probably a conversation to be had outside of this PR.

specification/flag-evaluation/hooks.md
toddbaert: I imagine this would be done via the [EvaluationOptions](https://github.com/open-feature/spec/blob/main/specification/types.md#evaluation-options). To your earlier point, I think a `HookHints` option would be apropos here.

specification/flag-evaluation/hooks.md
toddbaert: There exists languages where there simply isn't rigorous concepts around data, even at runtime. There are not types - but just "bytes" that can be interpreted in multiple ways.

We will probably never deal with such languages, so I'm not married to keeping this requirement.

specification/flag-evaluation/hooks.md
justinabrahms: Going w/ `finallyAfter`.

specification/flag-evaluation/hooks.md
justinabrahms: The intent of this line was:  If you have an error from the provider and that somehow triggers an error in the error flow.. don't overwrite their original error with this new one.

I'm not actually clear what an error in finally should do. Our whole MO seems to be "never throw, except maybe around client initialization", but also I don't want to have this logic complexity in the finally handler.

```python
try:
  before()
  eval()
  after()
except:
  error()
finally:
  try:
    finally()
  except:
    error()
```

specification/flag-evaluation/hooks.md
justinabrahms: > If an error occurs during the evaluation of `before` or `after` hooks, any remaining hooks in the `before` or `after` stages **MUST** not be invoked.

^ I'm going to try that. The intent is errors are kinda like an exception in that it stops the flow and bubbles up to the error stage.

specification/flag-evaluation/hooks.md
justinabrahms: Moved to `HookHints`. Much better name. :) Can we resolve this thread?

specification/flag-evaluation/hooks.md
moredip: ah got it. I misunderstood the meaning of invocation in the definitions. I thought it just meant a specific flag evaluation, but I think the intention is for invocation to mean the call that the Application Author makes into the evaluation API, vs the call the framework makes into the provider API.

specification/flag-evaluation/hooks.md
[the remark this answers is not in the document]
moredip: no I think I just misunderstood what invocation was defining. We can leave it as is.

specification/flag-evaluation/hooks.md
InTheCloudDan: 👍 on resolving

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,174 @@
...
+### Hook creation and parameters
+
+
+##### Requirement 3.1
+
+> Hooks **MUST** specify at least one stage.
+
+##### Requirement 3.2
+
+> The `before` stage **MUST** run before flag evaluation occurs. It accepts a `hook context` (required) and `state` (optional) as parameters and returns either a `HookContext` or nothing.
```
toddbaert, line 71: Seems like we have a conflict here between `state` and `HookHints`? They are the same thing, right?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,174 @@
...
+**Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)` is an invocation.
+**API**: The global API singleton.
+
+### Hook context
+
+Hook context exists to provide hooks with information about the invocation.
+
+###### Requirement 1.1
+
+> Hook context **MUST** provide: the flag key, evaluation context, default value and a list of  executed hooks by stage.
```
toddbaert, line 21: Did we decide to remove the "list of executed hooks by stage"?

specification/flag-evaluation/hooks.md
```diff
@@ -2,4 +2,174 @@
...
+
+
+##### Condition 1.2
+
+> If the language type system differentiates between strings, numbers, booleans, and structures.
+>
+>##### Condition 1.2.1
+>
+> Condition: You **MUST** provide `flag type`.
```
toddbaert, line 30: I think we decide to remove this condition, and just make this a requirement.

Also, avoid use of "you". Use imperative mood / passive voice:

"`Flag type` must be provided" or "`Hook context must contain the flag type`"

toddbaert:
I can't speak for @moredip , but besides the comments I just made (which I think are already resolved and just yet to be updated) I'm good with this. I will approve but recommend you follow up with @moredip

toddbaert (approved):

moredip:
no additional concerns on my end 👍

moredip (approved):

PR #52, opened 2022-05-05
New parser for requirements

justinabrahms:
Diff should give a good sense of the differences.

Main one, is this parser understands heirarchy of `#`'s within markdown. 

Nice things:
1. No more putting conditions in weird quote blocks.
2. Actually generate the correct headings (e.g. h1, h3) instead of making everything an h5.
3. The format of the json is consistent (e.g. every node has a `children` property)


toddbaert:
Thanks again. I think besides one typo this should be merged.

toddbaert (approved):

PR #53, opened 2022-05-06
parser: add support for .specignore file

dyladan:
When running the parser I was annoyed that some non-specification files had JSON files created, so I modified the parser to support a basic `.specignore` file which is just a line-separated list of UNIX style globs.

Also cleaned up some unused and duplicate imports in the parser


toddbaert:
Nice change, looks good to me, runs fine for me. Thanks a lot!

PR #54, opened 2022-05-06
Add a github actions workflow to ensure JSON files are up to date

dyladan:
This ensures that the markdown and JSON files can never be out of sync.

### DIFF
--- a/.github/workflows/json-changed.yaml
+++ b/.github/workflows/json-changed.yaml
@@ -1,0 +1,31 @@
+# This action ensures that any PR targeting the main branch has up-to-date generated JSON files.
+# It runs the parser and checks for changes in git tracked files.
+# If any changes are detected, the workflow fails
+
+name: generated
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - main
+jobs:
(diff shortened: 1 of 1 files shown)


toddbaert (approved):
:bow:

.github/workflows/json-changed.yaml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  json:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+
+      - name: Generate JSON files
+        run: python ./tools/specification_parser/specification_parser.py
```
justinabrahms, line 21: We should also run the linter that's in the same directory.

justinabrahms (changes requested):

.github/workflows/json-changed.yaml
toddbaert: This is probably not a bad idea - should it be in a new different though? I'm not that familiar with GH actions, there might be some reason to use a different step for improved messaging/granularity? Maybe that would also slow things down :shrug:

PR #57, opened 2022-05-06
Add lint job

dyladan:


.github/workflows/markdown-checks.yaml
```diff
@@ -29,3 +29,12 @@ jobs:
           else
             echo "JSON files up to date. Looks good!"
           fi
+
+  markdown-lint:
```
justinabrahms, line 33:
```suggestion
  json-lint:
```

dyladan: I didn't realize this was linting the JSON. Is there any point to lint generated files at all?

toddbaert: It might be a technically easier way to validate the normative statements, but to your point, we could embed that functionality directly into the parser.

dyladan: to validate them how. are the statements themselves checked by the linter and not just the format/spacing?

toddbaert: > to validate them how. are the statements themselves checked by the linter and not just the format/spacing?

Yes - it checks that an RFC2119 keyword is present, for instance.

dyladan: no lint is probably a fine name for that

justinabrahms (approved):

toddbaert (approved):
Looks great, thanks again.

PR #58, opened 2022-05-06
Add check to make sure toc is up to date

dyladan:


specification/glossary.md
```diff
@@ -9,23 +9,26 @@ This document defines some terms that are used across this specification.
 
 - [Feature Flag](#feature-flag)
 - [User Roles](#user-roles)
-  - [Application Author](#application-author)
-  - [Application Integrator](#application-integrator)
-  - [Provider Author](#provider-author)
-  - [Integration Author](#integration-author)
-  - [Library Author](#library-author)
+  * [Application Author](#application-author)
```
dyladan, line 12: Looks like this is already catching an out of date ToC 😄

justinabrahms (approved):

package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "@openfeature/specification",
+  "private": true,
```
dyladan, line 3: private prevents this from accidentally being published to NPM

package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "@openfeature/specification",
```
dyladan, line 2: name insures that the package-lock.json name will not be different if someone uses a different directory name for their cloned project like I did. If this happens, the TOC check can fail.

toddbaert (approved):

PR #59, opened 2022-05-09
Potentially clarify a missed case around hook evaluation and finally.

justinabrahms:
Had a formatting snafu. Not sure if we should drop this entirely or account for it in the bullets.

### DIFF
--- a/specification/flag-evaluation/hooks.md
+++ b/specification/flag-evaluation/hooks.md
@@ -109,17 +109,23 @@ client.getValue('my-flag', 'defaultValue', new Hook3());
 > - before: API, Client, Invocation
 > - after: Invocation, Client, API
 > - error (if applicable): Invocation, Client, API
-> - finally: Invocation, Client, API> If an error occurs in the `finally` hook, it **MUST NOT** trigger the `error` hook.
+> - finally: Invocation, Client, API
 
 ##### Requirement 4.3
 
-> If an error occurs in the `before` or `after` hooks, the `error` hooks **MUST** be invoked.
+> If an error occurs in the `finally` hook, it **MUST NOT** trigger the `error` hook.
+
+In practice, this means that errors that occur in the finally hook will bubble up.
 
 ##### Requirement 4.4
 
-> If an error occurs during the evaluation of `before` or `after` hooks, any remaining hooks in the `before` or `after` stages **MUST NOT** be invoked.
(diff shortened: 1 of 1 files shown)


specification/flag-evaluation/hooks.md
```diff
@@ -109,18 +109,24 @@ client.getValue('my-flag', 'defaultValue', new Hook3());
...
 > - error (if applicable): Invocation, Client, API
-> - finally: Invocation, Client, API> If an error occurs in the `finally` hook, it **MUST NOT** trigger the `error` hook.
+> - finally: Invocation, Client, API
 
 ##### Requirement 4.3
 
-> If an error occurs in the `before` or `after` hooks, the `error` hooks **MUST** be invoked.
+> If an error occurs in the `finally` hook, it **MUST NOT** trigger the `error` hook.
+
+In practice, this means that errors that occur in the finally hook will bubble up.
```
toddbaert, line 118: This makes sense to me, maybe we should add something similar about errors in error hooks.

specification/flag-evaluation/hooks.md
```diff
@@ -109,18 +109,24 @@ client.getValue('my-flag', 'defaultValue', new Hook3());
...
 
 ##### Requirement 4.4
 
-> If an error occurs during the evaluation of `before` or `after` hooks, any remaining hooks in the `before` or `after` stages **MUST NOT** be invoked.
+> If an error occurs in the `before` or `after` hooks, the `error` hooks **MUST** be invoked.
 
 ##### Requirement 4.5
 
+> If an error occurs during the evaluation of `before` or `after` hooks, any remaining hooks in the `before` or `after` stages **MUST NOT** be invoked.
```
toddbaert, line 126: I think generally, we can say this is true for any stage. If an error occurs in error, or finally, I think we'd expect no remaining hooks to run and the error to be returned or thrown.

toddbaert:
Couple minor comments - and our CI process is reminding you to run the JSON generation :partying_face:

specification/flag-evaluation/hooks.md
justinabrahms: Interestingly, in 4.6, it says 

> If an error is encountered in the error stage, it **MUST NOT** be returned to the user.

The big question is: How bullet proof do we expect hook authors to write their code?

I think I'll need to look at other implementations (OTel, maybe) to figure out what they've learned.

PR #60, opened 2022-05-10
Add evaluation context spec

toddbaert:
First pass at evaluation context.

Lot's of questions here. I opted not to leverage something like the [elastic common schema](https://www.elastic.co/guide/en/ecs/current/ecs-client.html) for now, I believe it might be overkill. I used it, as well as [OIDCs standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) only as inspiration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ _check_python:
 		fi;
 .PHONY: markdown-toc
 markdown-toc:
-	@if ! npm ls markdown-toc; then npm install; fi
+	@if ! npm ls markdown-toc; then npm ci; fi
 	@for f in $(ALL_DOCS); do \
 		if grep -q '<!-- tocstop -->' $$f; then \
 			echo markdown-toc: processing $$f; \
--- a/specification/evaluation-context/evaluation-context.json
+++ b/specification/evaluation-context/evaluation-context.json
@@ -1,0 +1,16 @@
+[
+    {
+        "id": "Requirement 3.1",
+        "clean id": "requirement_3_1",
+        "content": "The `evaluation context` structure MUST define an optional `targeting key` field of type string, identifying the subject of the flag evaluation.",
+        "RFC 2119 keyword": "MUST",
+        "children": []
+    },
+    {
+        "id": "Requirement 3.2",
+        "clean id": "requirement_3_2",
+        "content": "The evaluation context MUST support the inclusion of custom fields, having keys of type `string`, and values of type `boolean | string | number | datetime | structure`.",
+        "RFC 2119 keyword": "MUST",
+        "children": []
+    }
+]
--- a/specification/evaluation-context/evaluation-context.md
+++ b/specification/evaluation-context/evaluation-context.md
@@ -1,5 +1,25 @@
 # Evaluation Context
 
+**Status**: [Experimental](../README.md#document-statuses)
+
 ## Overview
 
-// TODO: this is a stub.
+The `evaluation context` provides ambient information for the purposes of flag evaluation. Contextual data may be used as the basis for targeting, including rule-based evaluation, overrides for specific subjects, or fractional flag evaluation.
+
+### Fields
+
+NOTE: Field casing is not specified, and should be chosen in accordance with language idioms.
+
+see: [types](../types.md)
+
+##### Requirement 3.1
+
+> The `evaluation context` structure **MUST** define an optional `targeting key` field of type string, identifying the subject of the flag evaluation.
+
+The targeting key uniquely identifies the subject (end-user, or client service) of a flag evaluation. Providers may require this field for fractional flag evaluation, rules, or overrides targeting specific users. Such providers may behave unpredictably if a targeting key is not specified at flag resolution.
+
+##### Requirement 3.2
+
+> The evaluation context **MUST** support the inclusion of custom fields, having keys of type `string`, and values of type `boolean | string | number | datetime | structure`.
+
+see: [structure](../types.md#structure), [datetime](../types.md#datetime)
--- a/specification/flag-evaluation/flag-evaluation.json
+++ b/specification/flag-evaluation/flag-evaluation.json
@@ -165,8 +165,8 @@
     {
         "id": "Requirement 1.21",
         "clean id": "requirement_1_21",
-        "content": "The `client` MUST transform the `evaluation context` using the `provider's` `context transformer` function, before passing the result of the transformation to the provider's flag resolution functions.",
-        "RFC 2119 keyword": "MUST",
+        "content": "The `client` SHOULD transform the `evaluation context` using the `provider's` `context transformer` function if one is defined, before passing the result of the transformation to the provider's flag resolution functions.",
+        "RFC 2119 keyword": "SHOULD",
         "children": []
     }
 ]
--- a/specification/types.md
+++ b/specification/types.md
@@ -20,6 +20,10 @@ A numeric value of unspecified type or size. Implementation languages may furthe
 
 Structured data, presented however is idiomatic in the implementation language, such as JSON or YAML.
 
+### Datetime
+
+A language primitive for representing a date and time, including timezone information.
+
 ### Evaluation Details
 
 A structure representing the result of the [flag evaluation process](./glossary.md#evaluating-flag-values), and made available in the [detailed flag resolution functions](./flag-evaluation/flag-evaluation.md#detailed-flag-evaluation), containing the following fields:
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -48,11 +48,11 @@ jobs:
       - uses: actions/setup-python@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '18'
+          node-version: "18"
 
       - name: Generate ToC
         run: make markdown-toc
-  
+
       - name: Check for file changes
         run: |
           if [[ $(git diff --name-only) ]]
--- a/specification/README.md
+++ b/specification/README.md
@@ -6,8 +6,9 @@
 - [Types](./types.md)
 - [Design Principles](design-principles.md)
 - [Evaluation API](./flag-evaluation/flag-evaluation.md)
-- [Hooks](./flag-evaluation/hooks.md)
 - [Providers](./provider/providers.md)
+- [Evaluation Context](./evaluation-context/evaluation-context.md)
+- [Hooks](./flag-evaluation/hooks.md)
 
 ## Conformance
 
--- a/specification/flag-evaluation/flag-evaluation.md
+++ b/specification/flag-evaluation/flag-evaluation.md
@@ -191,6 +191,6 @@ It's recommended to provide non-blocking mechanisms for flag evaluation, particu
 
 ##### Requirement 1.21
 
-> The `client` **MUST** transform the `evaluation context` using the `provider's` `context transformer` function, before passing the result of the transformation to the provider's flag resolution functions.
+> The `client` **SHOULD** transform the `evaluation context` using the `provider's` `context transformer` function if one is defined, before passing the result of the transformation to the provider's flag resolution functions.
 
 See [context transformation](../provider/providers.md#context-transformation) for details.
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -28,6 +28,8 @@ This document defines some terms that are used across this specification.
   * [Flag](#flag)
   * [Variant](#variant)
   * [Values](#values)
+  * [Targeting](#targeting)
+  * [Fractional Evaluation](#fractional-evaluation)
   * [Rule](#rule)
 
 <!-- tocstop -->
@@ -132,6 +134,14 @@ reverse: [5,4,3,2,1]
 wonky: [3,5,2,1,4]
 standard: [1,2,3,4,5]
 ```
+
+### Targeting
+
+The application of rules, specific user overrides, or fractional evaluations in feature flag resolution.
+
+### Fractional Evaluation
+
+Assigning flag values randomly according to a configured proportion or percentage (ie: 50/50).
 
 ### Rule
 
--- a/specification/provider/providers.md
+++ b/specification/provider/providers.md
@@ -23,7 +23,7 @@ The `provider` API defines interfaces that Provider Authors can use to abstract 
 resolveBooleanValue(flagKey, defaultValue, context, options);
 ```
 
-See: [flag resolution structure](../types.md#flag-resolution), [flag value resolution](../glossary.md#flag-value-resolution)
+see: [flag resolution structure](../types.md#flag-resolution), [flag value resolution](../glossary.md#flag-value-resolution)
 
 ##### Condition 2.3
 


specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+
+##### Requirement 3.1
+
+> The `evaluation context` structure **MUST** define a required `targeting key` field of type string, identifying the subject of the flag evaluation.
+
+The targeting key uniquely identifies the subject (end-user, or client service) of a flag evaluation. Providers may use this as a unique identifier for fractional-flags, rules, or overrides targeting specific users.
+
+##### Requirement 3.2
+
+> The evaluation context **MUST** define the following optional fields: `email` (string), `first name` (string), `last name`(string), `name`(string), `client ip`(string), `tz`(string), `locale`(string), `date`(string | date), `environment`(string), `application`(string), and `version`(string).
```
toddbaert, line 19: Tables look awful in the normative statements/JSON, for now I'm mentioning the fields in the statements and leaving the table in the auxiliary non-normative section.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+
+| Name        | Type           | Purpose                                            | Example(s)                                                |
+| ----------- | -------------- | -------------------------------------------------- | --------------------------------------------------------- |
+| email       | string         | Subject RFC 5322 conformant email                  | "first.last@example.com"                                  |
+| first name  | string         | Subject given name                                 | "Jane"                                                    |
+| last name   | string         | Subject last name                                  | "Smith"                                                   |
+| name        | string         | Subject full name and titles and suffixes          | "Ms. Jane Smith"                                          |
+| client ip   | string         | Client IP address (IPv4 or IPv6)                   | "216.7.132.218","2001:0db8:85a3:0000:0000:8a2e:0370:7334" |
+| tz          | string         | Subject "zoneinfo" timezone                        | "Brazil/West","America/New_York"                          |
```
toddbaert, line 30: Could also call this `zoneinfo` instead of `tz`.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+
+See the table below for details.
+
+| Name        | Type           | Purpose                                            | Example(s)                                                |
+| ----------- | -------------- | -------------------------------------------------- | --------------------------------------------------------- |
+| email       | string         | Subject RFC 5322 conformant email                  | "first.last@example.com"                                  |
+| first name  | string         | Subject given name                                 | "Jane"                                                    |
+| last name   | string         | Subject last name                                  | "Smith"                                                   |
+| name        | string         | Subject full name and titles and suffixes          | "Ms. Jane Smith"                                          |
+| client ip   | string         | Client IP address (IPv4 or IPv6)                   | "216.7.132.218","2001:0db8:85a3:0000:0000:8a2e:0370:7334" |
```
toddbaert, line 29: Is `remote address` a better name? Unleash uses that, Launchdarkly uses `ip`. `ip` seems vague, I'd like to imply that we mean a client IP, not the host in the case of server-side flag evaluation.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+| tz          | string         | Subject "zoneinfo" timezone                        | "Brazil/West","America/New_York"                          |
+| locale      | string         | Subject RFC 5646 locale                            | "pt-BR","en-US"                                           |
```
toddbaert, line 31: LaunchDarkly has `Country`, and gives the example of `"Uganda"`, but says the value only need be a string. I think it's better for us to specify a specific format. It seems this would be valid for LD anyway.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+| date        | date \| string | Date primitive or date in extended ISO 8601 format | "2022-05-10T01:53:29.823Z"                                |
+| environment | string         | Identifier for deployed environment                | "prod", "development"                                     |
+| application | string         | Application identifier                             | "fibonacci-api"                                           |
```
toddbaert, line 34: `environment` and `application` are defined by Unleash, and I think they are generally useful.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+| version     | string         | Application version                                | "1.3.1","v2-beta"                                         |
```
toddbaert, line 35: Cloudbees has an `app_release`. I think `version` is a slightly better term that accomplishes the same thing, and could be mapped accordingly in the Cloudbees provider.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+
+\*NOTE: Field casing is not specified, and should be chosen in accordance with language idioms.
```
toddbaert, line 37: I don't think we have to worry about the casing much since moving this data over the wire isn't really a concern... we just want to make sure it's reflected in the `evaluation context` structure in any implementation.

Makefile
```diff
@@ -19,7 +19,7 @@ _check_python:
 		fi;
 .PHONY: markdown-toc
 markdown-toc:
-	@if ! npm ls markdown-toc; then npm install; fi
+	@if ! npm ls markdown-toc; then npm ci; fi
```
toddbaert, line 22: Took me a while to figure out what was happening, but this `npm install` was causing the `git diff` in the ToC check to detect changes. I guess this should take care of it?

specification/evaluation-context/evaluation-context.md
agentgonzo: >`first name` (string), `last name`(string)

Not all cultures have a firstname/lastname construct. I also don't see what advantage you get for flag evaluation by having first and last names separately over just the `name` field.

specification/evaluation-context/evaluation-context.md
agentgonzo: previous c constructs used to use just `addr`. I think the 'remote' is implied, so maybe just `address`?

specification/flag-evaluation/flag-evaluation.json
```diff
@@ -165,8 +165,8 @@
     {
         "id": "Requirement 1.21",
         "clean id": "requirement_1_21",
-        "content": "The `client` MUST transform the `evaluation context` using the `provider's` `context transformer` function, before passing the result of the transformation to the provider's flag resolution functions.",
-        "RFC 2119 keyword": "MUST",
+        "content": "The `client` MAY transform the `evaluation context` using the `provider's` `context transformer` function, before passing the result of the transformation to the provider's flag resolution functions.",
+        "RFC 2119 keyword": "MAY",
```
agentgonzo, line 169: SHOULD?

specification/types.md
```diff
@@ -20,6 +20,10 @@ A numeric value of unspecified type or size. Implementation languages may furthe
 
 Structured data, presented however is idiomatic in the implementation language, such as JSON or YAML.
 
+### Date
+
+A language primitive for representing a date, possibly including timezone information.
```
agentgonzo, line 25: I would say it `MUST` include timezone information, otherwise it's just ambiguous. ISO8601 should be used.

agentgonzo, line 25:
```suggestion
A language primitive for representing a date, including timezone information.
```

specification/evaluation-context/evaluation-context.md
beeme1mr: `address` could be confused with a home address. Especially, with the information collected above such as name and email address. I actually don't mind `ip`  in this situation.

specification/evaluation-context/evaluation-context.md
beeme1mr: I think `tz` is commonly used and understood.

https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

specification/evaluation-context/evaluation-context.md
agentgonzo: ip is fine

specification/evaluation-context/evaluation-context.md
beeme1mr: `RFC 5646` is for language not location. `ISO_3166-2` may be better for this purpose.

toddbaert: Rfc5646 is for language, but that was my intended purpose of this, not location. I think locale is fairly important.

You could also use tz as a proxy for location.

specification/evaluation-context/evaluation-context.json
```diff
@@ -0,0 +1,23 @@
...
+        "content": "The `evaluation context` structure MUST define a required `targeting key` field of type string, identifying the subject of the flag evaluation.",
+        "RFC 2119 keyword": "MUST",
+        "children": []
+    },
+    {
+        "id": "Requirement 3.2",
+        "clean id": "requirement_3_2",
+        "content": "The evaluation context MUST define the following optional fields: `email` (string), `first name` (string), `last name`(string), `name`(string), `client ip`(string), `tz`(string), `locale`(string), `date`(string | date), `environment`(string), `application`(string), and `version`(string).",
+        "RFC 2119 keyword": "MUST",
+        "children": []
```
beeme1mr, line 14: It would be great if the children section could contain the optional fields.

specification/evaluation-context/evaluation-context.md
toddbaert: I see your point @agentgonzo , but `firstName` and `lastName` are defined properties in [LaunchDarkly's context](https://docs.launchdarkly.com/home/users/attributes), so not having them might add some unpredictability between that provider and others. They are also included in the [standard OIDC claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims).

Let's see what others think.

specification/evaluation-context/evaluation-context.md
toddbaert: I will change this to `ip`

specification/evaluation-context/evaluation-context.json
toddbaert: `children` is for nested conditional requirements. Can you clarify what you mean?

specification/flag-evaluation/flag-evaluation.json
toddbaert: Will update this to `**SHOULD**`

specification/evaluation-context/evaluation-context.md
dabeeeenster: Some vendors have semver-aware operators, allowing you to control flags with something along the lines of `if version >= 2.4.12`. I'm not sure this necessitates its own semver field, but something to consider.

dabeeeenster: I think this falls back on how the interfaces to the individual rule operators are going to work. We decided to have semver specific operators; evaluating whether something was a semver version or not turned out to be quite cpu expensive. I think having a single field is the right way to go but wanted to flag it - the other properties here have less ambiguous meanings.

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,41 @@
...
+| ----------- | -------- | ----------------------------------------- | --------------------------------------------------------- |
+| email       | string   | Subject RFC 5322 conformant email         | "first.last@example.com"                                  |
+| first name  | string   | Subject given name                        | "Jane"                                                    |
+| last name   | string   | Subject last name                         | "Smith"                                                   |
+| name        | string   | Subject full name and titles and suffixes | "Ms. Jane Smith"                                          |
+| client ip   | string   | Client IP address (IPv4 or IPv6)          | "216.7.132.218","2001:0db8:85a3:0000:0000:8a2e:0370:7334" |
+| tz          | string   | Subject "zoneinfo" timezone               | "Brazil/West","America/New_York"                          |
+| locale      | string   | Subject RFC 5646 locale                   | "pt-BR","en-US"                                           |
+| timestamp   | datetime | The time of event/request                 | "2022-05-10T01:53:29.823Z"                                |
```
toddbaert: Unleashed has a `currentTime` field. I guess there could be a use case for this. I've added the stipulation that it should be in client local time, as I think that could be useful.

Does that make sense? Is it too prescriptive? Do we even really need this field?

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,42 @@
...
+**Status**: [Experimental](../README.md#document-statuses)
+
 ## Overview
 
-// TODO: this is a stub.
+The `evaluation context` provides ambient information for the purposes of flag evaluation. Contextual data may be used as the basis for targeting, including rule-based evaluation, overrides for specific subjects, or fractional flag evaluation.
+
+### Fields
+
+##### Requirement 3.1
```
justinabrahms, line 11: Why start at 3.1?

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,42 @@
...
+
+> The `evaluation context` structure **MUST** define a required `targeting key` field of type string, identifying the subject of the flag evaluation.
+
+The targeting key uniquely identifies the subject (end-user, or client service) of a flag evaluation. Providers may use this as a unique identifier for fractional-flags, rules, or overrides targeting specific users.
+
+##### Requirement 3.2
+
+> The evaluation context **MUST** define the following optional fields: `email` (string), `first name` (string), `last name`(string), `name`(string), `ip`(string), `tz`(string), `locale`(string), `country code` (string), `timestamp`(date), `environment`(string), `application`(string), and `version`(string).
```
justinabrahms, line 19: Can you say a little more about the interplay between MUST and optional?  Like.. it has to return null if something were to ask for that value?

toddbaert: > Can you say a little more about the interplay between MUST and optional? Like.. it has to return null if something were to ask for that value?

It has to be defined, but can be empty/null. In Java, I would expect some getters/setters for these values, I think, as well as some other place to store additional attributes (some HashMap, perhaps).

specification/evaluation-context/evaluation-context.md
toddbaert: Oh - I had each md file start at a different number. So Section 1 was evaluation API, second 2 was providers, 3 is evaluation context, etc. 

I see now that `hooks` starts at 1, so we werent on the same page here. We should review this.

justinabrahms: Seems like we should also add it to the linter. :)

specification/evaluation-context/evaluation-context.md
s-sen: Does this mean that the client can do something like context.getAttribute("ip") and expect null if the provider does not support this? Or does the context need a specific API for for each of these attributes e.g. context.getEmailAttribute()?

toddbaert: > Does this mean that the client can do something like context.getAttribute("ip") and expect null if the provider does not support this? Or does the context need a specific API for for each of these attributes e.g. context.getEmailAttribute()?

@s-sen  Continuing with a hypothetical Java implementation, I think for improved discoverability, it would be ideal if the context had a `context.getEmailAttribute()`. This way, the compiler/editor can help developers explore the standard fields, and even potentially give them creative ideas for rule-based flags. It also implicitly shows that these attributes are "first class". All additional custom attributes would work like `context.getCustomAttribute("myCustomAttr")`. That's my thinking.

@justinabrahms

justinabrahms: @toddbaert I suppose I'm not entirely sold on having these as 'first class' attributes. I'm concerned about why we need them to be first class. 

I don't want developers to see this as a TODO list (especially b/c I'd prefer our developers to not be shoving in PII like the names of users into the flagging system). 

If providers need this information, I'm concerned that it will bloat over time and be really noisy. If they don't need this information.. then I'm not sure why we're dictating these fields.

toddbaert: @justinabrahms I think it goes back to the fact that some providers DO have these as first class fields (launch darkly). Since they do, I think we need it in OpenFeature as well. If we don't, we end up in the following hypothetical situation:

- Launch Darkly defines `ip` for the location of the `ip` field in their SDK.
- In their `provioder` doc, they say authors should populate a field in the OpenFeaure provider called `ip`, and their provider will pull this value from there.
- Application Author does as instructed above
- Application Author switches to a new custom provider, `provider x`, which dictates the ip should be defined in a field called `remote_addr`.
- Application Author must update all their code to now put the IP here.

If this was a "well-defined" field, each provider would "know" where to get these values, and we have no concerns switching between providers.

justinabrahms: I better understand what you mean. These keys are for providers to agree upon so they can map their own sdk attributes to them. I'm fine with this for now, but would like to talk about PII in the future. 

https://github.com/open-feature/spec/issues/64

specification/evaluation-context/evaluation-context.json
beeme1mr: I was just saying that it may be easier to read if each field was a separate child requirement.

specification/evaluation-context/evaluation-context.md
beeme1mr: I could go either way but it doesn't seem necessary to me. It looks like Unleash uses it for time based releases but it defaults to the current time if nothing is set.

dabeeeenster: This seems superfluous to me, and also opens a rabbit hole of time zones, time formats etc etc

Often epoch time is more useful when driving flag rules as it means simple < > against ints rather than manipulating datetime formats. But that can just be set as a regular int. So I guess my vote would be to not have it. It's also implicit in terms of when the request was sent?

beeme1mr: Couldn't the provide just send the system time? Would you ever want to send a time that isn't the current time? Even if you did, excluding it from the semantics here wouldn't prevent you from doing that. I would recommend we omit it now and reevaluate if someone opens an issue in the future.

toddbaert: >Couldn't the provide just send the system time? Would you ever want to send a time that isn't the current time?

I REALLY don't think it's a good idea for providers to add data to the context themselves, especially in the "official" fields. In fact, I was going to add that as a point to the spec. I think this is a PII concern.

specification/evaluation-context/evaluation-context.json
```diff
@@ -0,0 +1,23 @@
+[
+    {
+        "id": "Requirement 3.1",
+        "clean id": "requirement_3_1",
+        "content": "The `evaluation context` structure MUST define a required `targeting key` field of type string, identifying the subject of the flag evaluation.",
```
justinabrahms, line 5: I think this is in conflict with Hooks 1.1 which says: 
> Hook context MUST provide: the flag key, flag type, evaluation context, and the default value.

If you have to provide an `evaluation context` and an `evaluation context MUST define a required 'targeting key' field`, then that conflicts with Flag Evaluation 1.7 which says context is optional.

> The client MUST provide methods for flag evaluation, with parameters flag key (string, required), default value (boolean | number | string | structure, required), evaluation context (optional), and evaluation options (optional), which returns the flag value.

Previous to this requirement I was specifying an empty EvaluationContext if one didn't exist. Can/should I just generate a random value as the targeting key? Something else?

justinabrahms, line 5: I think this is in conflict with Hooks 1.1 which says: 
> Hook context MUST provide: the flag key, flag type, evaluation context, and the default value.

If you have to provide an `evaluation context` and an `evaluation context MUST define a required 'targeting key' field`, then that conflicts with Flag Evaluation 1.7 which says context is optional.

> The client MUST provide methods for flag evaluation, with parameters flag key (string, required), default value (boolean | number | string | structure, required), evaluation context (optional), and evaluation options (optional), which returns the flag value.

Previous to this requirement I was specifying an empty EvaluationContext if one didn't exist. Can/should I just generate a random value as the targeting key? Something else?

specification/evaluation-context/evaluation-context.json
```diff
@@ -0,0 +1,23 @@
...
+        "RFC 2119 keyword": "MUST",
+        "children": []
+    },
+    {
+        "id": "Requirement 3.2",
+        "clean id": "requirement_3_2",
+        "content": "The evaluation context MUST define the following optional fields: `email` (string), `first name` (string), `last name`(string), `name`(string), `ip`(string), `tz`(string), `locale`(string), `country code` (string), `timestamp`(date), `environment`(string), `application`(string), and `version`(string).",
```
justinabrahms, line 12:
```suggestion
        "content": "The evaluation context MUST provide access the following optional attributes: `email` (string), `first name` (string), `last name`(string), `name`(string), `ip`(string), `tz`(string), `locale`(string), `country code` (string), `timestamp`(date), `environment`(string), `application`(string), and `version`(string).",
```

Optional suggestion: Making it clear that you want setters/getters for java, not raw field access.

justinabrahms, line 12:
```suggestion
        "content": "The evaluation context MUST provide access the following optional attributes: `email` (string), `first name` (string), `last name`(string), `name`(string), `ip`(string), `tz`(string), `locale`(string), `country code` (string), `timestamp`(date), `environment`(string), `application`(string), and `version`(string).",
```

Optional suggestion: Making it clear that you want setters/getters for java, not raw field access.

toddbaert: > MUST provide access the following optional attributes

I see what you're getting at, but this suggest read-only to me (access, but not mutation).

toddbaert: > MUST provide access the following optional attributes

I see what you're getting at, but this suggest read-only to me (access, but not mutation).

specification/evaluation-context/evaluation-context.json
toddbaert: Seems like another duplicate suggestion, see my other response.

toddbaert: Ya... I understand your confusion. This could be a bit sticky...

Let me try to explain my reasoning for this, and perhaps we can come up with a solution:

Some vendor SDKs require a context parameter (specifically, LaunchDarkly requires a `User`), and that parameter MUST be set for ALL flag evaluations, and MUST have a key (user identifier). Others (Flagsmith) don't require a user identifier, but if there isn't one, attributes cannot be used in the evaluation (the flag evaluation must be "contextless"). 

What I was trying to suggest (and failed to communicate) was that if any context fields were set, the `targetingKey` must also be set, this way, we can at least gracefully handle the Flagsmith case. It doesn't, however, really solve the LaunchDarkly issue. I'm also not sure it's really the right choice. To put all this another way, we have a couple problems:

- How do we deal with vendors that ALWAYS require a user identifier? I don't think we want to make context required... 
- How do we handle SDKs where the vendor doesn't require a user identifier for "contextless" flag evaluations, but requires one if any attributes are defined?

The empty `evaluation context` implementation detail is dependent on what we decide above, which admittedly, I'm a bit stuck on.

cc @beeme1mr

toddbaert: @InTheCloudDan Maybe you can weigh in on the above. Is it reasonable to auto-generate a LDUser key? Will that cause inconsistent behavior in percentage rollout situations, for instance?

specification/evaluation-context/evaluation-context.json
justinabrahms: I can get on board with that. I think the thing I actually care about is s/fields/attributes.

specification/evaluation-context/evaluation-context.json
justinabrahms: I don't know what the correct behavior is for LD.  I would expect the provider to either say: "You didn't specify a targeting key. Giving you the default value (e.g. light error)" or "You didn't specify a targeting key. I'm going to make one up for you."

InTheCloudDan: @justinabrahms we require the users to generate that, our signature enforces the `user/context` to be passed into the SDK at least on the server side.

specification/evaluation-context/evaluation-context.json
InTheCloudDan: Where are all of these required optional fields coming from? I understand they're optional but why are their optionally enforced is it because of differences between all of the existing commercial providers?

specification/evaluation-context/evaluation-context.md
InTheCloudDan: @toddbaert from the LaunchDarkly perspective these are optional fields and agree with @justinabrahms on not having them as `first class` attributes, we leave them up to the customer to decide on usage and marking private.

specification/evaluation-context/evaluation-context.json
```diff
@@ -0,0 +1,23 @@
...
+        "RFC 2119 keyword": "MUST",
+        "children": []
+    },
+    {
+        "id": "Requirement 3.3",
+        "clean id": "requirement_3_3",
+        "content": "The evaluation context MUST support the inclusion of custom fields, having keys of type `string`, and values of type `boolean | string | number | datetime`.",
```
InTheCloudDan, line 19: @toddbaert Can we add a `list` and `object` to these fields. 

For `datetime` is it expected to be RFC3339 or what specification are we looking to require here?

specification/evaluation-context/evaluation-context.json
toddbaert: At the risk of speaking for @justinabrahms , I think he's of the opinion we should definite these fields now. See my long explanation here: https://github.com/open-feature/spec/pull/60#discussion_r871587025

Let me know if this addresses your concern or not.

specification/evaluation-context/evaluation-context.json
toddbaert: @InTheCloudDan yep, the question is what LD would do for context-less evaluations. All user data is stored in the evaluation context. If a flag evaluation is done without a context parameter being passed, what would a hypothetical launch darkly provider do?

specification/evaluation-context/evaluation-context.json
InTheCloudDan: For the `datetime` it seems like it would be clearer across language if we decided on which standard we expect, that way being explicit there's less room for subtle bugs across languages.

Also thoughts on adding `array` and `object` to the custom fields type? We will need support of those

specification/evaluation-context/evaluation-context.json
InTheCloudDan: @toddbaert we would need to discuss internally, generating an ID is something we have not previously done except maybe in the case of our React SDK on the client side so I would need to confirm that.

specification/evaluation-context/evaluation-context.json
InTheCloudDan: It appears to be a decent number of defined fields. I view it as we want to start with less of a surface API and build on it over time versus having a lot of fields up front that we don't know will be used and things are always harder to remove in the future then add.

specification/evaluation-context/evaluation-context.json
InTheCloudDan: Just trying to write out ideas, OF could support a `default context` function at the client layer that returns a context if none is passed at an evaluation call site. But it would still be a difference where for some providers it is allowed to be empty and others it cannot be.

specification/evaluation-context/evaluation-context.json
justinabrahms: @InTheCloudDan In the java SDK, I'm planning to use [java.time.ZonedDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html). Internally, we're representing it as a string, but that's opaque to the consumers of it.

justinabrahms:
https://github.com/open-feature/java-sdk/pull/5 - javasdk pr around eval context. May be useful to look at for folks, especially around large number of setters/getters on it for attributes.

specification/evaluation-context/evaluation-context.json
patricioe: at Spit.io we have the same problem. 
> Is it reasonable to auto-generate a LDUser key?
At Split, the implications of randomizing this are huge. If a flag is set with a percentage rollout, then a user can get different flag values/treatment as we use consistent hashing for the bucketing logic and rely in the user identity (userId for instance) to be the constant identifier. So without the notion of targeting key, at least Split cannot be a provider you can just add unless the above user code that interacts with the reference SDK changes.

justinabrahms: What I'm hearing is that either:

-  the targeting key is set through something like a hook
- you have to always pass an evaluation context with at least a targeting key

Anyone see other options?

toddbaert: I understand that some providers require it. It's trivially easy to add a global hook to set it, and it will be come even easier to do so once we've implemented implicit context propagation via continuations or thread-local storage, so here's what I propose:

It stays optional. Providers that require it can throw or otherwise error during evaluation with a error log saying it must be set. This will cause them to default, and application authors will see that this provider requires the `targeting key` to be set, and address it with a global hook, or at evaluation. Like I said above, this will become even easier when we can use implicit context propagation to grab user details off of the thread/golang-context/node-continuation.

Thoughts? 

@justinabrahms @patricioe @InTheCloudDan

specification/evaluation-context/evaluation-context.md
toddbaert: I've removed `firstName`, `lastName` and `name`.

specification/evaluation-context/evaluation-context.json
toddbaert: I've separated all the fields into their own requirement, but kept the table for an overview as well.

specification/evaluation-context/evaluation-context.md
toddbaert: I've removed the `timestamp` field.

specification/evaluation-context/evaluation-context.json
toddbaert: > For the datetime it seems like it would be clearer across language if we decided on which standard we expect, that way being explicit there's less room for subtle bugs across languages.

I've removed this field for now based on your other feedback @InTheCloudDan . If we decide we want to make it standard we can revisit this question.

> Also thoughts on adding array and object to the custom fields type? We will need support of those.

In the spirit of keeping this simple, are these really necessary? If they are, I'd suggest the language we use elsewhere: `structured data` since "object" and "array" don't always mean exactly the same things across languages (for example, when you say "object" you probably mean a "Map" in Java.

justinabrahms (approved):

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,53 @@
...
+The `evaluation context` provides ambient information for the purposes of flag evaluation. Contextual data may be used as the basis for targeting, including rule-based evaluation, overrides for specific subjects, or fractional flag evaluation.
+
+### Fields
+
+See the table below for a summary of pre-defined fields.
+
+| Name          | Type   | Purpose                             | Example(s)                                                 |
+| ------------- | ------ | ----------------------------------- | ---------------------------------------------------------- |
+| targeting key | string | Subject identifier                  | "user@domain.com", "cb638b68-db72-11ec-9d64-0242ac120002"  |
+| ip            | string | Client IP address (IPv4 or IPv6)    | "216.7.132.218", "2001:0db8:85a3:0000:0000:8a2e:0370:7334" |
```
agentgonzo, line 16: Why do we need to have the client spectify this? It seems like something the API layer could (should?) automatically add via `gethostaddr` or similar. It's very odd to have a client go "this is my IP". Often the sever can also obtain it directly if needed from the host addr or proxy header info. I don't see any benefit in including it here (I'm happy to be educated)

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,53 @@
...
+| country code  | string | Subject ISO 3166-2 country code     | "BR", "US"                                                 |
```
agentgonzo, line 17: this also seems a bit odd.... how would it even work in practise? Have the developer call getLocale then pass the country code? It seems like something that the API could add directly if its required.

agentgonzo (approved):
I'm OK with this as a starting point. I have a feeling that this spec will probably change dramatically as we start to develop the API and provider implementations.

My gut feeling is that we're trying to define the minimal set of common fields for the evaluation context among our different vendors, but in reality this minumal set is the empty set because the different vendors have such different implementations of what is mandatory.

In practise though, I don't think that anything in here is prohibitive. If a vendor requires a given field that is not specifed in this context, it can be part of the contract for that specific provider to provide that extra field. Similarly, if a vendor doesn't require a field made mandatory by this spec (country code seems the obvious one), then a dummy value can then be used and ignored by the provider. Personally, I would remove the requirements (at the openfeature API level) for specific fields (3.1-3.6) and just keep the requirement about the _format_ of the context object (3.7 - string keys and bool/string/number values) - but I seem to be in the minority for this so won't block the issue.

I guess we will find out in the implementation how problematic the two above situations are for developers and adjust the spec/API in subsequent releases. I see this all as alpha anyway until GA (and even after GA, it'll be beta and subject to subsequent change).

specification/evaluation-context/evaluation-context.md
```diff
@@ -1,5 +1,53 @@
...
+| environment   | string | Identifier for deployed environment | "prod", "development"                                      |
+| application   | string | Application identifier              | "fibonacci-api"                                            |
+| version       | string | Application version                 | "1.3.1" ,"v2-beta"                                         |
```
agentgonzo, line 20: this feels like something that should be set globally at the provider/client, rather that in the context for each individual evaluation.

specification/evaluation-context/evaluation-context.md
beeme1mr: Something like `gethostaddr` would return the server IP, not the IP address of the user making the request. While it may be possible to automate collecting the client IP in some situations, I would argue that we don't want to. It should be up to the developer/business to decide if that information is required. It also becomes a bit more tricky if the server is behind proxies because that information may only be available via a header such as `X-Forwarded-For`

specification/evaluation-context/evaluation-context.md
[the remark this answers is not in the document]
beeme1mr: Evaluation Context can be set at three levels: globally, client, and during flag evaluation. The evaluation context is then merged before flag evaluation occurs.

specification/evaluation-context/evaluation-context.md
agentgonzo: >Something like gethostaddr would return the server IP, not the IP address of the user making the request. 

For sever-side (backend) flag evaluation, that is true. But for browser SDKs, the OpenFeature APIs are called in the javascript in the end-user's browser, not the server. Thus the IP specified in this evaluation context would/could be looked up locally by the browser. The browser code accessing the OpenFeature SDK won't know about any proxies anyway, and will probably see the local IP address as 192.168.1.1 on the local subnet. I really don't see how setting the `ip` in the evaluation context helps anyone here.

specification/evaluation-context/evaluation-context.md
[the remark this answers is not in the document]
agentgonzo: 👍 OK. I thought it was just in the flag evaluation. It makes total sense to set stuff like that at the global/client level.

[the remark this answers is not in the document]
toddbaert: I will mention this in auxiliary sections.

toddbaert:
@agentgonzo:

A quote from you:

> If a vendor requires a given field that is not specifed in this context, it can be part of the **contract for that specific provider** to provide that extra field.

It's "contracts" between application author and provider that scare me a bit. App author should be mostly able to completely ignorant of the provider in use, ideally. In a sense, contract between these parties reduces portability. If provider-X uses `ip` and provider-Y uses `remote_addr` then to switch between them you are re-writing some code. If we standardize these fields this is much less likely to happen, at least with "commonly used" fields. Maybe I'm overly concerned about this :shrug:

agentgonzo:
I might be overthinking it. I got a bit confused by the concept of "MUST provide an optional field". You're right in that for common fields, we should have a consistent way of naming that concept (`ip` vs `remote_addr` as you say), so yeah, I do now see the reason for these fields. (However, I'm not convinced by the need for the ip field - but it's not one that CloudBees uses as a named field, so it may be something I am ignorant of).

In short, I retract my objection to these fields 👍

specification/evaluation-context/evaluation-context.json
InTheCloudDan: Can there be some more information around setting this as a global hook? So all of the evaluations would have the same key or the hook would return a random key?

For providers that require it this appears to be a suboptimal experience due to the effects it can have on evaluation/return value as @patricioe pointed out.

toddbaert: >Can there be some more information around setting this as a global hook? So all of the evaluations would have the same key or the hook would return a random key?

@InTheCloudDan 

Ya I think what I said before was a bit unclear. I meant that a global hook might be used to dynamically add a unique user identifier based on some contextual data. Perhaps you have the user's `email` in the context - you could hash that in a hook and set the `targeting key` that way. At the end of the day, as is the case now with any SDK that requires a "user identifier" of some kind, it will be up to the author to make sure a basis for identifying the user is available at the time of flag evaluation.

The implicit context propagation features that will be spec'd out later will help with this by defining a way to inject values into the flag evaluation dynamically without adding them directly in a flag evaluation, as we do [here](https://github.com/open-feature/playground/blob/main/packages/app/src/app/transaction-context.middleware.ts) in the playground.

I hope that's helpful.

toddbaert:
I think we're in a place to merge this. It seems like based on the community meeting, we're on the same page now about not defining any default fields except `targeting key`. There's still some challenges with that one in particular, but they deserve an issue outside this PR which @beeme1mr is creating, as discussed.

I'll be squashing and merging this today unless I hear objections or concerns!

PR #66, opened 2022-05-20
Fix glossary link in README

luiscrs14:


oleg-nenashev (approved):
Thanks for catching it!

PR #87, opened 2022-05-27
Markdown lint

justinabrahms:
Refs https://github.com/open-feature/spec/issues/36

May close it out entirely, depending on how strongly we feel about auto-TOC. 

If we want that, we might explore this: https://github.com/marketplace/actions/toc-generator. The [other options](https://stackoverflow.com/questions/9721944/automatic-toc-in-github-flavoured-markdown) I saw didn't seem the greatest. (pandoc doesn't pass lint, for example)

### DIFF
--- a/.markdown-link-check-config.json
+++ b/.markdown-link-check-config.json
@@ -1,0 +1,5 @@
+{
+  "ignorePatterns": [
+    {"pattern": ".*split.io.*"}
+  ]
+}
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -1,0 +1,1 @@
+default: true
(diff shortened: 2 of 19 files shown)


justinabrahms:
Looks like wrapping is causing our rendering to not have spaces in the json. I'll need to look into that.

oleg-nenashev (approved):
I would not enforce such short line breaks, but otherwise looks ok

justinabrahms:
I don't much care of the line lengths, but we should pick some standard. Happy to pick something else. 🤷‍♂️

justinabrahms:
Merging now so as to not cause a million merge conflicts for someone else. If we want to change the line length, that seems fine to me.

PR #90, opened 2022-05-31
Language & number consistency

### LINKED ISSUE #38: Finally hook errors?
What happens when a finally hook errors?

Here's some example evaluation logic:

```
try {
  run_before_hooks()
  get_value()
  run_after_hooks()
} catch() {
  error_handling()
} finally {
  run_finally_hooks()
}
```

If the finally hooks throw errors, they won't be caught which breaks 1.19 from the spec.

Otherwise, our code looks like this (which is very janky).

```
try {
  try {
    run_before_hooks()
    get_value()
    run_after_hooks()
  } catch() {
    error_handling()
  } finally {
    run_finally_hooks()
  }
} finally {
  error_handling() # again
}
```

toddbaert:
This spec has 2 commits, I strongly suggest you look at them one at a time, since the second is purely whitespace changes.

1st commit:

- breaks the flag-evaluation.md doc into sections, more like the hooks.md doc, to prevent so much number thrashing in the future
- improves consistency of terminology across files (see comments)
- specifies that both `error` and `finally` user-defined hooks may throw. This may be controversial, but I think it makes sense (and finally can already throw, as per spec). User-defined code being able to throw seems like a plus too me. See comments.

2nd commit: 

- sets `proseWrap: never` in prettier. This improves readability in PRs. As it stands, the fact that line-wrapping is on adds a lot of unneeded deltas in commits, since adding a new word might totally change the way the wrapping ends up. If we keep everything to a single line, it makes diffing much easier. This is mentioned in the [style guide](https://github.com/open-feature/spec#style-guide)

Fixes #38

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -159,7 +159,7 @@
         {
             "id": "Requirement 1.5.1",
             "machine_id": "requirement_1_5_1",
-            "content": "The `evaluation options` structure's `hooks` field denotes a collection of hooks that the client MUST execute for the respective flag evaluation, in addition to those already configured.",
+            "content": "The `evaluation options` structure's `hooks` field denotes an ordered collection of hooks that the client MUST execute for the respective flag evaluation, in addition to those already configured.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -351,7 +351,7 @@
         {
             "id": "Requirement 4.3.4",
             "machine_id": "requirement_4_3_4",
-            "content": "When `before` hooks have finished executing, any resulting `EvaluationContext`  MUST be merged with the invocation `EvaluationContext` with the invocation `EvaluationContext` taking precedence in the case of any conflicts.",
+            "content": "When `before` hooks have finished executing, any resulting `EvaluationContext` MUST be merged with the invocation `EvaluationContext` with the invocation `EvaluationContext` taking precedence in the case of any conflicts.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -381,7 +381,15 @@
             "machine_id": "condition_4_3_8",
             "content": "`finally` is a reserved word in the language.",
             "RFC 2119 keyword": null,
-            "children": []
+            "children": [
+                {
+                    "id": "Conditional Requirement 4.3.8.1",
+                    "machine_id": "conditional_requirement_4_3_8_1",
+                    "content": "Instead of `finally`, `finallyAfter` SHOULD be used.",
+                    "RFC 2119 keyword": "SHOULD",
+                    "children": []
+                }
+            ]
         },
         {
             "id": "Requirement 4.4.1",
--- a/specification.json
+++ b/specification.json
@@ -1,70 +1,70 @@
 {
     "rules": [
         {
-            "id": "Requirement 1.1",
-            "machine_id": "requirement_1_1",
+            "id": "Requirement 1.1.1",
+            "machine_id": "requirement_1_1_1",
             "content": "The `API`, and any state it maintains SHOULD exist as a global singleton, even in cases wherein multiple versions of the `API` are present at runtime.",
             "RFC 2119 keyword": "SHOULD",
             "children": []
         },
         {
-            "id": "Requirement 1.2",
-            "machine_id": "requirement_1_2",
+            "id": "Requirement 1.1.2",
+            "machine_id": "requirement_1_1_2",
             "content": "The `API` MUST provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 1.3",
-            "machine_id": "requirement_1_3",
+            "id": "Requirement 1.1.3",
+            "machine_id": "requirement_1_1_3",
             "content": "The `API` MUST provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 1.4",
-            "machine_id": "requirement_1_4",
+            "id": "Requirement 1.1.4",
+            "machine_id": "requirement_1_1_4",
             "content": "The API MUST provide a function for retrieving the `provider` implementation.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 1.5",
-            "machine_id": "requirement_1_5",
+            "id": "Requirement 1.1.5",
+            "machine_id": "requirement_1_1_5",
             "content": "The `API` MUST provide a function for creating a `client` which accepts the following options:  - name (optional): A logical string identifier for the client.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 1.6",
-            "machine_id": "requirement_1_6",
+            "id": "Requirement 1.2.1",
+            "machine_id": "requirement_1_2_1",
             "content": "The client MUST provide a method to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 1.7",
-            "machine_id": "requirement_1_7",
+            "id": "Requirement 1.3.1",
+            "machine_id": "requirement_1_3_1",
             "content": "The `client` MUST provide methods for flag evaluation, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Condition 1.8",
-            "machine_id": "condition_1_8",
+            "id": "Condition 1.3.2",
+            "machine_id": "condition_1_3_2",
             "content": "The language type system differentiates between strings, numbers, booleans and structures.",
             "RFC 2119 keyword": null,
             "children": [
                 {
-                    "id": "Conditional Requirement 1.8.1",
-                    "machine_id": "conditional_requirement_1_8_1",
+                    "id": "Conditional Requirement 1.3.2.1",
+                    "machine_id": "conditional_requirement_1_3_2_1",
                     "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure.",
                     "RFC 2119 keyword": "MUST",
                     "children": []
                 },
                 {
-                    "id": "Conditional Requirement 1.8.2",
-                    "machine_id": "conditional_requirement_1_8_2",
+                    "id": "Conditional Requirement 1.3.3",
+                    "machine_id": "conditional_requirement_1_3_3",
                     "content": "The `client` SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.",
                     "RFC 2119 keyword": "SHOULD",
                     "children": []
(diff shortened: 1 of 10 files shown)


specification/flag-evaluation/flag-evaluation.md
```diff
@@ -165,68 +165,66 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
-> functions or methods for the purposes for configuration or setup.
+> functions or methods for the purposes for configuration or setup, and
+> user-defined code such as hooks.
 
-##### Requirement 1.19
+User-defined code in `error` and `finally` hooks may throw or otherwise
+abnormally terminate. This has the benefit of providing a means of making errors
+optionally "loud" for debugging purposes (errors can be re-thrown in `error`
+hooks, for instance).
+
```
toddbaert, line 175: This is already the case for `finally`, I think it makes sense for `error` too...

specification/flag-evaluation/flag-evaluation.md
```diff
@@ -244,7 +242,19 @@ It's recommended to provide non-blocking mechanisms for flag evaluation,
...
+> The `evaluation options` structure's `hooks` field denotes a collection of
+> hooks that the client **MUST** execute for the respective flag evaluation, in
+> addition to those already configured.
+
+See [hooks](./hooks.md) for details.
+
+#### Context Transformation
+
+##### Requirement 1.6.1
```
toddbaert, line 198: This didn't change, was just moved.

specification/flag-evaluation/hooks.md
```diff
@@ -9,28 +9,29 @@ to flag evaluation. They operate similarly to middleware in many web frameworks.
 
 **Hook**: Application author/integrator-supplied logic that is called by the
 OpenFeature framework at a specific stage. **Stage**: An explicit portion of the
-flag evaluation lifecycle. e.g. `before` being "before the evaluation is run.
-**Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)`
-is an invocation. **API**: The global API singleton.
+flag evaluation lifecycle. e.g. `before` being "before the
+[resolution](../glossary.md#resolving-flag-values) is run. **Invocation**: A
```
toddbaert: Updated all cases where we meant "resolution" (getting value from a provider) but said "evaluation" (overall process of getting a value, including resolution, hooks, etc).

specification/flag-evaluation/hooks.md
```diff
@@ -140,23 +144,26 @@ client.getValue('my-flag', 'defaultValue', new Hook3());
 > If an error occurs in the `finally` hook, it **MUST NOT** trigger the `error`
 > hook.
 
-In practice, this means that errors that occur in the finally hook will bubble
-up.
+In practice, this means that abnormal termination in the finally hook will be
+exposed to the application author.
```
toddbaert: Minor wording change to be more language agnostic.

specification/flag-evaluation/hooks.md
```diff
@@ -169,19 +176,17 @@ val = client.get_boolean_value('my-key', False, evaluation_options={
...
-##### Requirement 4.5.3
+##### Requirement 4.5.2
 
 > `HookHints` **MUST** be passed to each hook.
 
-##### Requirement 4.5.4
+##### Requirement 4.5.3
 
-> The hook **MUST NOT** alter the `HookHints` object.
+> The hook **MUST NOT** alter the `HookHints` structure.
```
toddbaert, line 158: `object` => `structure`, which is a glossary term.

specification.json
```diff
@@ -383,6 +376,13 @@
             "RFC 2119 keyword": "MUST",
             "children": []
         },
+        {
+            "id": "Condition 4.3.8",
+            "machine_id": "condition_4_3_8",
+            "content": "`finally` is a reserved word in the language.",
+            "RFC 2119 keyword": null,
+            "children": []
```
justinabrahms, line 384: I think this should have children? You can make this happen by increasing the heading level of the thing under it.

specification/flag-evaluation/hooks.md
```diff
@@ -74,21 +77,22 @@ EvaluationContext|void before(HookContext, HookHints)
 
 ##### Requirement 4.3.4
 
-> When `before` hooks have finished executing, any resulting `EvaluationContext`
-> **MUST** be merged with the invocation `EvaluationContext` with the invocation
-> `EvaluationContext` taking precedence in the case of any conflicts.
+> When `before` hooks have finished executing, any resulting
+> `EvaluationContext` > **MUST** be merged with the invocation
```
justinabrahms: This has an intermixed `>` character. May want to ensure the rest of these don't have that too.

specification/flag-evaluation/hooks.md
```diff
@@ -97,13 +101,13 @@ EvaluationContext|void before(HookContext, HookHints)
 > stages. It accepts a `hook context` (required) and `HookHints` (optional).
 > There is no return value.
 
-##### Condition 3.8
+##### Condition 4.3.8
 
 > `finally` is a reserved word in the language.
 >
-> ##### 3.8.1
+> ##### 4.3.8.1
```
justinabrahms, line 83: In the new parser, don't quote this, just increase the heading level by 1.

specification.json
```diff
@@ -1,171 +1,171 @@
...
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 1.18",
-            "machine_id": "requirement_1_18",
-            "content": "Methods, functions, or operations on the client MUST NOT throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.",
+            "id": "Requirement 1.4.9",
+            "machine_id": "requirement_1_4_9",
+            "content": "Methods, functions, or operations on the client MUST NOT throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the `default value` in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup, and user-defined code such as hooks.",
```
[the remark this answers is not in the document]
toddbaert: > I know I keep waffling on this. This makes me a little uneasy. I'm trying to think through the API from a user's perspective and I think we're going to end up with {reference ugly code}

I'm guilty of waffling too. For now, I feel that at the very least, for simplicity sake, we need to treat finally and error the same (if we allow `finally` to throw, we allow `error` to throw).

>This would mean that users don't get a guarantee that error hooks run (e.g. if error hook 1 throws, then error hook 2 wouldn't get invoked).

This sort of thing already happens in before/after - if a before hook throws, no other before hooks are executed. I dont think it's too astonishing to assume the same thing would occur in the error handler, but we should define it if we go this way.

On the side of throwing, I would say that `error` and `finally` hooks will likely be simple and low-risk enough that they rarely throw. Hooks are also not "SDK-code" proper, but code the app-author adds, which I think we can think about a bit differently from the main SDK functionality (Cloudbees considers "user-space" code "different" from primary SDK code as well, and allows some configuration around the error behavior, see [here](https://docs.cloudbees.com/docs/cloudbees-feature-management-api/latest/api-reference/java-api#_setuserspaceunhandlederrorhandler). I would also recommend that they throw unchecked exceptions, so that in Java, for instance, the compiler won't force you to wrap API calls in a try/catch. Throwing also allows for intentional re-throwing from an error, which might be useful in debugging situations, for instance, when developing a provider.

Lastly, the hooks semantics basically mirror a try/catch. Re-throwing in a catch results in propagation of the exception back up through the stack, and the rest of the catch block not executing. This is analogous to the behavior this PR defines for hooks.

On the side of swallowing... This is objectively the "safer" route. It does also allow us to guarantee all `error` and `finally` hooks are executed, and app code will NEVER break, not even if the author writes a really broken hook. I'm not 100% this is the least-astonishing way to go, though.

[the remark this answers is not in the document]
beeme1mr: > This sort of thing already happens in before/after - if a before hook throws, no other before hooks are executed. I dont think it's too astonishing to assume the same thing would occur in the error handler, but we should define it if we go this way.

An argument could be made for either approach. However, I think it may be worth considering catching errors so other hooks are still executed. Imagine if you have a hook that modifies context but runs after a misbehaving hook, this would lead to unexpected behavior.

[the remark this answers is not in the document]
beeme1mr: Ah, that makes sense.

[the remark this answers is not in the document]
toddbaert: >When we start pulling in libraries for hooks, I get less sure. Now I need to validate that the hook I'm taking a dependency on isn't going to throw and crash my application, returning 5xx to the user.

Yep, that's a fair point. We do expect some of these to come from third party / a contrib repository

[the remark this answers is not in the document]
toddbaert: I had another thought - what if a `finally` involves closing some resource and and an error results in NOT closing it, and therefor a leak? I think at the moment, the safest path is to never throw... so swallow everything and run ALL the `error` and `finally` hooks. That's the path we can take now to have the least regrets later.

I think if you guys are in agreement, I will update the spec for this. One other relevant issue is https://github.com/open-feature/spec/issues/39. We would likely want to log an error on this interface eventually, even if for now we just print to stderr.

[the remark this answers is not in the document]
justinabrahms: Always run finally and error hooks. Always swallow the errors.  I feel good about that.

tools/specification_parser/lint_json_output.py
```diff
@@ -24,11 +24,11 @@ def main(f):
             for entry in entries:
                 if entry.get('RFC 2119 keyword') is None and \
                    'condition' not in entry['id'].lower():
-                    print(f"{j}: Rule {entry['id']} is missing a RFC 2119 keyword", file=sys.stderr)
+                    print(f"{jsonfile.name}: Rule {entry['id']} is missing a RFC 2119 keyword", file=sys.stderr)
```
toddbaert, line 27: This variable didn't exist, looks like a partial rename.

specification.json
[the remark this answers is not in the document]
toddbaert: OK, I updated the spec for this: https://github.com/open-feature/spec/pull/90/files#r887153737

specification/flag-evaluation/hooks.md
```diff
@@ -137,26 +114,23 @@ client.getValue('my-flag', 'defaultValue', new Hook3());
...
-up.
+In languages with try/catch semantics, this means that exceptions thrown in `finally` hooks should be caught, and not propagated up the call stack.
 
 ##### Requirement 4.4.4
 
-> If an error occurs in the `before` or `after` hooks, the `error` hooks
-> **MUST** be invoked.
+> If an `error` hook abnormally terminates, evaluation **MUST** proceed normally, including the execution of any remaining `error` hooks.
+
+In languages with try/catch semantics, this means that exceptions thrown in `error` hooks should be caught, and not propagated up the call stack.
```
toddbaert, line 125: @justinabrahms @beeme1mr Here are the new requirements around errors/finally, as discussed.

specification/flag-evaluation/hooks.md
```diff
@@ -137,26 +114,23 @@ client.getValue('my-flag', 'defaultValue', new Hook3());
...
-up.
+In languages with try/catch semantics, this means that exceptions thrown in `finally` hooks should be caught, and not propagated up the call stack.
 
 ##### Requirement 4.4.4
 
-> If an error occurs in the `before` or `after` hooks, the `error` hooks
-> **MUST** be invoked.
+> If an `error` hook abnormally terminates, evaluation **MUST** proceed, including the execution of any remaining `error` hooks.
+
+In languages with try/catch semantics, this means that exceptions thrown in `error` hooks should be caught, and not propagated up the call stack.
```
toddbaert, line 125: @beeme1mr @justinabrahms This resolves the error/finally discussion.

specification.json
```diff
@@ -408,15 +408,15 @@
...
+            "RFC 2119 keyword": null,
             "children": []
         },
         {
             "id": "Requirement 4.4.4",
             "machine_id": "requirement_4_4_4",
-            "content": "If an error is encountered in `error` stage, it MUST be returned to the application author.",
-            "RFC 2119 keyword": "MUST",
+            "content": "If an `error` hook abnormally terminates, evaluation should proceed normally, including the execution of any remaining `error` hooks.",
+            "RFC 2119 keyword": null,
```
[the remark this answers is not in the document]
toddbaert: Yep, I also found a minor bug with the linter python script with this. Fixed it with this PR.

beeme1mr (approved):

justinabrahms (approved):

Issue #84, opened 2022-05-27
[Proposal] Remove Get Provider from the Flag Evaluation API


beeme1mr:
### Proposal

The [flag evaluation API requirement 1.3](https://github.com/open-feature/spec/blob/main/specification/flag-evaluation/flag-evaluation.md#requirement-13) states that a function to retrieve the provider implementation is required. This was originally defined in the spec to allow hooks to access the provider. However, this causes confusion because the application developer is not meant to interact with the provider directly.

This issue could be solved by replacing `Get Provider` with a `provider name` getter.

### Reference

-  https://cloud-native.slack.com/archives/C0344AANLA1/p1653426865053149

toddbaert:
I think the use-cases were all associated with similar getting some basic data from the provider, such as the name (used in the OTel hook, for instance). I propose adding a `provider metadata` concept, that for now will contain just the provider name. We can add additional metadata as required, so it's somewhat future-proof. We can expose this instead of the provider itself.

toddbaert:
I've re-opened this because @weyert has brought up a use-case, and I think it's brings up a good point.

We decided to remove the provider accessor from the `client` object. I think this makes sense. The `client` is basically the primary interface for OpenFeature to the application author, so I don't want to pollute it. However, especially right now while we are still missing some features such as eventing and tracking, we may want to provide SOME access to the underlying provider, which might expose some of these from it's underlying SDK, or even provider an means of accessing that SDK.

I propose we continue NOT to expose the provider instance on the client, or in hooks, but that we DO have a means of retrieving it from the top-level API object, ie: `OpenFeature.getProvider()`. This will return the registered provider, which the author will have to cast to their particular implementation, this allowing them to access the specific features of their underlying SDK, if their provider makes them available.

@justinabrahms  I'm particularly interested in your take here.

cc @weyert @davejohnston @benjiro @ajhelsby

justinabrahms:
Can you share the use case? I'm not opposed to it, but preference towards generally not going stuff. :)

weyert:
> Can you share the use case? I'm not opposed to it, but preference towards generally not going stuff. :)
> […](#)

Yes, the use case is to be able to call `identify` and `alias`, `groupIdentify`, and `captureEvent` on the client. That's my use case so that I can identify user also from the backend and not only front-end code. Or prepare user's organisations by calling `groupIdentify`.

[...]

PR #93, opened 2022-06-07
No client/provider instances, add metadata

toddbaert:
This replaces the instances of `client` and `provider` from the hookContext, instead adding `metadata` fields which contain `name` props. I chose a metadata field because it gives us some flexibility to add more such properties later

Closes: https://github.com/open-feature/spec/issues/84

Let me know what you guys think.


specification.json
```diff
@@ -24,7 +24,7 @@
         {
             "id": "Requirement 1.1.4",
             "machine_id": "requirement_1_1_4",
-            "content": "The API MUST provide a function for retrieving the `provider` implementation.",
+            "content": "The API MUST provide a function for retrieving the metadata field of the configured `provider`.",
```
davejohnston, line 27: I think this is good - it answers the query I had, and I think will avoid any confusion, or abuse of the API.

davejohnston (approved):
Looks good - this satisfies the query I had about not exposing the provider directly.

Issue #44, opened 2022-04-28
Don't allow exceptions during client creation


moredip:
I just noticed that the current draft spec has:

> Requirement 1.19
> No methods, functions, or operations on the client should ever throw exceptions, or otherwise abnormally terminate. Flag evaluation calls must always return the default value in the event of abnormal execution. Exceptions include functions or methods for the purposes for configuration or setup.

I understand the motivation for never throwing exceptions, but I think we're failing to meet the spirit of this for client-side JS if we allow exceptions when creating a client, because of the unique lifecycle of browser-based apps. For a client-side app the client is often created just before an evaluation is done, so an exception thrown during client creation is pretty much just as disruptive as an exception when you call a method.

Because of this, I would propose that the spec not allow exceptions during client creation.

toddbaert:
This also might relate to a question that's come up a few times: do we want to expose events (specifically readiness events) from the provider interface? Providers could sidestep this by requiring SDKs that have a readiness event simply be ready BEFORE passing them into the provider. For instance, the CloudbeesProvider may require an already-initialized Cloudbees SDK instance. Not sure.

toddbaert:
I guess the only question here is if we should specifically add a requirement that `.getClient()` calls never throw. I think it's probably a good thing to do.

PR #94, opened 2022-06-07
Client creation must not throw

toddbaert:
See title, based on discussion here: https://github.com/open-feature/spec/issues/44

Closes: https://github.com/open-feature/spec/issues/44

### DIFF
--- a/specification/flag-evaluation/flag-evaluation.md
+++ b/specification/flag-evaluation/flag-evaluation.md
@@ -62,6 +62,12 @@ OpenFeature.getClient({
 
 The name is a logical identifier for the client.
 
+#### Requirement 1.1.6
+
+> The client creation function **MUST NOT** throw, or otherwise abnormally terminate.
+
+Clients may be created in critical code paths, and even per-request in server-side HTTP contexts. Therefor, in keeping with the principle that OpenFeature should never cause abnormal execution of the first party application, this function should never throw. Abnormal execution in initialization should instead occur during provider registration.
+
 ### Client Usage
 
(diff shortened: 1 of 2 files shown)


specification/flag-evaluation/flag-evaluation.md
```diff
@@ -62,6 +62,12 @@ OpenFeature.getClient({
 
 The name is a logical identifier for the client.
 
+#### Requirement 1.1.6
+
+> The client creation function **MUST NOT** throw, or otherwise abnormally terminate.
+
+Clients may be created in critical code paths, and even per-request in server-side HTTP contexts. Therefor, in keeping with the principle that OpenFeature should never cause abnormal execution of the first party application, this function should never throw. Abnormal execution in initialization should instead occur during provider registration.
```
beeme1mr, line 69:
```suggestion
Clients may be created in critical code paths, and even per-request in server-side HTTP contexts. Therefore, in keeping with the principle that OpenFeature should never cause abnormal execution of the first party application, this function should never throw. Abnormal execution in initialization should instead occur during provider registration.
```

toddbaert: TIL: https://www.grammarly.com/blog/therefore-vs-therefor/#:~:text=Therefore%20is%20an%20adverb%20that,%E2%80%9D%20or%20%E2%80%9Cfor%20it.%E2%80%9D

moredip (approved):

toddbaert:
Another FYI @justinabrahms and @davejohnston , though I don't expect this will be much of a change.

PR #100, opened 2022-06-15
Subdirectories don't feel necessary.

justinabrahms:


beeme1mr:
Thanks for doing this. I agree, this that subdirectories are not necessary.

beeme1mr (approved):

beeme1mr:
I think we're linking to parts of the spec in other repos. That shouldn't stop us from making this change but we should audit the org and make sure links are updated accordingly.

PR #102, opened 2022-06-16
feat: start working on Tracking Events proposal

weyert:
A draft proposal for the Tracking Event to allow application developers to emit tracking events which can be useful when their feature flag system to support driving feature flag evaluation by previous events or person properties.

Use case:
Application developer has a dynamic cohort which puts users into  it after they have completed a step in the application, for example the onboarding wizard. A person gets added to the cohort after a specific event has been emitted in relation to the user.

Next a feature flag can be defined which is only active for users in this specific cohort

### DIFF
--- a/specification/events.md
+++ b/specification/events.md
@@ -1,0 +1,38 @@
+# Tracking Events
+
+## Overview
+
+Tracking Events are events that can be emitted by application developers.
+
+### Definitions
+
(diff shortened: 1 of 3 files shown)


beeme1mr:
I'm going to close this PR until a OFEP for tracking events have been accepted.

Issue #115, opened 2022-07-23
Float/Integer evaluation in typed langages


thomaspoignant:
I was looking at the different SDKs and we don't have a feature parity concerning the number flag evaluation.

Sometimes the number method retrieves `integer` (java SDK) and others are retrieving `float`.
It could be great to be explicit in the spec about what behavior we expect.

[Reference in the specs](https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md#condition-132)

toddbaert:
I'm hesitant to define exactly what types/bit-lengths/signs we expect, because there's so much variety across languages. I think it's best for the spec to stay vague here. What we do have is this: https://github.com/open-feature/spec/blob/main/specification/types.md#number, which we could perhaps link/reference better throughout the spec.

thomaspoignant:
What is not clear from me is if we want to have one `getNumber` function that is generic or do we also accept to have  multiple functions such as `getInteger`, `getFloat`, etc ...

thomaspoignant:
Resolves in #116

PR #116, opened 2022-07-23
Have different methods for float and int

thomaspoignant:
See https://github.com/open-feature/spec/issues/115 for the context.
Discussion also on [slack](https://cloud-native.slack.com/archives/C0344AANLA1/p1658587751496079).

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -59,24 +59,31 @@
         {
             "id": "Requirement 1.3.1",
             "machine_id": "requirement_1_3_1",
-            "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure.",
+            "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.",
             "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Condition 1.3.2",
+            "machine_id": "condition_1_3_2",
+            "content": "The implementation language differentiate between floats/ints.",
+            "RFC 2119 keyword": null,
             "children": [
                 {
                     "id": "Conditional Requirement 1.3.2.1",
                     "machine_id": "conditional_requirement_1_3_2_1",
-                    "content": "If languages differentiate between floats/ints, they SHOULD provide functions for both, consistent with language idioms.",
-                    "RFC 2119 keyword": "SHOULD",
-                    "children": []
-                },
-                {
-                    "id": "Conditional Requirement 1.3.3",
-                    "machine_id": "conditional_requirement_1_3_3",
-                    "content": "The `client` SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.",
+                    "content": "The client SHOULD provide functions for all types, consistent with language idioms.",
                     "RFC 2119 keyword": "SHOULD",
                     "children": []
                 }
             ]
+        },
+        {
+            "id": "Requirement 1.3.3",
+            "machine_id": "requirement_1_3_3",
+            "content": "The `client` SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.",
+            "RFC 2119 keyword": "SHOULD",
+            "children": []
         },
         {
             "id": "Requirement 1.4.1",
--- a/specification.json
+++ b/specification.json
@@ -66,7 +66,7 @@
         {
             "id": "Condition 1.3.2",
             "machine_id": "condition_1_3_2",
-            "content": "The implementation language differentiate between floats/ints.",
+            "content": "The implementation language differentiates between floating-point numbers and integers.",
             "RFC 2119 keyword": null,
             "children": [
                 {
--- a/specification.json
+++ b/specification.json
@@ -72,7 +72,7 @@
                 {
                     "id": "Conditional Requirement 1.3.2.1",
                     "machine_id": "conditional_requirement_1_3_2_1",
-                    "content": "The client SHOULD provide functions for strings, numbers, booleans and structures, consistent with language idioms.",
+                    "content": "The client SHOULD provide functions for floating-point numbers and integers, consistent with language idioms.",
                     "RFC 2119 keyword": "SHOULD",
                     "children": []
                 }
--- a/specification.json
+++ b/specification.json
@@ -72,7 +72,7 @@
                 {
                     "id": "Conditional Requirement 1.3.2.1",
                     "machine_id": "conditional_requirement_1_3_2_1",
-                    "content": "The client SHOULD provide functions for all types, consistent with language idioms.",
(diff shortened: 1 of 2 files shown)


specification/flag-evaluation.md
```diff
@@ -124,6 +111,17 @@ MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: '
...
+
+> If languages differentiate between floats/ints, they **SHOULD** provide functions for both, consistent with language idioms.
+
+```go
+// example in GO
+GetIntValue(flag string, defaultValue int64, evalCtx EvaluationContext, options ...EvaluationOption) (int64, error)
+
+GetFloatValue(flag string, defaultValue float64, evalCtx EvaluationContext, options ...EvaluationOption) (float64, error)
+```
+
```
toddbaert, line 124: `Conditional requirements` should be children of `Conditions`, so I think you want to put your "if" clause in a condition statement, and separate it from the requirement.

specification/flag-evaluation.md
```diff
@@ -93,19 +93,6 @@ client.getMetadata().getName(); // "my-client"
...
-// example flag evaluation
-var myValue = client.getValue("my-flag", false);
-```
-
-##### Condition 1.3.2
-
-> The language type system differentiates between strings, numbers, booleans and structures.
-
-###### Conditional Requirement 1.3.2.1
-
```
toddbaert, line 108: I think we should still include the parameters list, we can just merge it with the "typed flag evaluation" requirement.

toddbaert (changes requested):
I agree fully with the concept, thanks for noticing! I left some comments on structure, and I don't think we want to lose the description of the params for the flag evaluation methods; we can just merge the typed/untyped requirements, while keeping the params:

>with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional),

Added a few more SDK devs to the PR as well.

toddbaert:
@justinabrahms @agentgonzo @InTheCloudDan pinging you all on this one, since it's relevant to Java, and also the Cloudbees SDK differentiates between floats/ints, while the LD SDK uses an abstraction that can be interpreted as either. Justin [suggested](https://github.com/open-feature/java-sdk/issues/32) we use something more generic, like [Number](https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html), which I'm also OK with, but I think in that case we still might want to add something to the spec as @thomaspoignant suggests to make sure we have some way of support both floating point and values.

I think personally, I would prefer to just provide separate methods as @thomaspoignant has suggested here, but I'm interested in other takes...

justinabrahms:
I'm not opposed to separating floats vs ints. I'd just like to avoid it if we could.   But I'm hearing that maybe we can't? So... let's not. :)

specification/flag-evaluation.md
thomaspoignant: I have fixed that.

I have also moved the `Conditional Requirement 1.3.3` to a **requirement** since it does not have any condition associated.

specification/flag-evaluation.md
thomaspoignant: I have merged it.

benjiro:
This sounds fine to me, do we need to be more explicit about types so it's not left up to interpretation? Eg
float vs float64(double) vs float128(decimal)
int32 vs int64(long)

Thoughts? @toddbaert @thomaspoignant

thomaspoignant:
> float vs float64(double) vs float128(decimal)
> int32 vs int64(long)

It is a bit too detailed from my point of view.
Typescript does not have a diff between int and float, in java flot is not standard but they are using double, etc …
I would avoid being precise here.

toddbaert:
Agreed. We also mention this, in the `types.md` doc:

>A numeric value of unspecified type or size. Implementation languages may further differentiate between integers, floating point numbers, and other specific numeric types and provide functionality as idioms dictate.

https://github.com/open-feature/spec/blob/main/specification/types.md#number

specification.json
```diff
@@ -59,32 +59,32 @@
...
                     "machine_id": "conditional_requirement_1_3_2_1",
-                    "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure.",
-                    "RFC 2119 keyword": "MUST",
-                    "children": []
-                },
-                {
-                    "id": "Conditional Requirement 1.3.3",
-                    "machine_id": "conditional_requirement_1_3_3",
-                    "content": "The `client` SHOULD guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.",
+                    "content": "The client SHOULD provide functions for all types, consistent with language idioms.",
```
toddbaert, line 75: I think we should link to the "types" doc here if we say "all types"... but it might be better to say "strings, numbers, booleans and structures"

toddbaert (approved):
Sounds like we're in agreement. LGTM assuming we clarify the language around "all types".

specification.json
thomaspoignant: I have made this change, so and we have added a link to types for more details.

thomaspoignant:
BTW the ci is failing on this link `https://www.w3.org/TR/qaframe-spec/` that is working fine from where am I.

specification.json
```diff
@@ -59,32 +59,32 @@
...
-            "content": "The `client` MUST provide methods for flag evaluation, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.",
+            "content": "The `client` MUST provide methods for typed flag evaluation, including boolean, numeric, string, and structure, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Condition 1.3.2",
             "machine_id": "condition_1_3_2",
-            "content": "The language type system differentiates between strings, numbers, booleans and structures.",
+            "content": "The implementation language differentiate between floats/ints.",
```
toddbaert, line 69:
```suggestion
            "content": "The implementation language differentiates between floating-point numbers and integers.",
```

toddbaert (approved):

toddbaert:
>BTW the ci is failing on this link https://www.w3.org/TR/qaframe-spec/ that is working fine from where am I.

Re-ran it, seems to work now. Probably the site was slow for a time.

beeme1mr (approved):
This looks good to me too. It's a bummer we need it but not every language can be as great as JavaScript! 😄

toddbaert:
Anyone want to add a final approval for this? Seems like we have consensus.

specification/flag-evaluation.md
```diff
@@ -124,7 +111,24 @@ MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: '
...
 See [evaluation context](./evaluation-context.md) for details.
 
-###### Conditional Requirement 1.3.3
+##### Condition 1.3.2
+
+> The implementation language differentiates between floating-point numbers and integers.
+
+###### Conditional Requirement 1.3.2.1
+
+> The client **SHOULD** provide functions for strings, numbers, booleans and structures, consistent with language idioms.
```
justinabrahms, line 120: I'm thinking this line should only include numbers because the condition above says "assuming your language differentiates between floating and non floating numbers.."?

thomaspoignant: Oh good catch I've updated it.

justinabrahms (approved):

PR #118, opened 2022-07-28
restructure spec and include metadata for docs

beeme1mr:
At the last community meeting, we talked about including the spec in our documentation. This can be done with a git submodule and copying over the files as a prebuild step in Docusaurus. In preparation for that, I've updated the structure of the spec and included additional metadata. No changes to the spec itself were made.

Here is are a few examples of what it looks like in the docs.

image

image

Signed-off-by: Michael Beemer <michael.beemer@dynatrace.com>

### DIFF
--- a/.markdownlint.yml
+++ b/.markdownlint.yml
@@ -6,3 +6,5 @@ MD033:
     - details
 MD024:
   siblings_only: true
+MD025:
+  front_matter_title: ""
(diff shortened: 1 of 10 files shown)


toddbaert:
I think this is great.

As a side note, I like how the fact there's no delta in the `specification.json` proves no normative content has changed in PRs such as this.

toddbaert (approved):

benjiro (approved):

justinabrahms (approved):

skyerus (approved):

PR #119, opened 2022-08-02
Remove context transformers. Add provider hooks

justinabrahms:
refs https://github.com/open-feature/research/pull/22

### DIFF
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -117,7 +117,7 @@ A `provider hook` exposes a mechanism for `providers` to register [`hooks`](./04
 
 ##### Requirement 2.10
 
-> The provider interface **MUST** define a `provider hook` mechanism which can be optionally implemented in order to add `hook` instances to the client.
+> The provider interface **MUST** define a `provider hook` mechanism which can be optionally implemented in order to add `hook` instances to the evaluation life-cycle.
 
 ```
 class MyProvider implements Provider {
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -113,11 +113,11 @@ ResolutionDetails<MyStruct> resolveStructureValue(string flagKey, MyStruct defau
 
 #### Provider hooks
 
-Feature flag management systems often need to transform the context structures the user provides or generally address lifecycle concerns. A `provider hook` exposes a mechanism for providers to register [`Hooks`](./04-hooks.md) to tap into various stages of the flag evaluation lifecycle.
+A `provider hook` exposes a mechanism for `providers` to register [`hooks`](./04-hooks.md) to tap into various stages of the flag evaluation lifecycle. As one example, feature flag management systems often need to transform the context structures the user provides.
 
 ##### Requirement 2.10
 
-> The provider interface **MAY** define a `provider hook` mechanism which can be optionally implemented in order to add `hook` instances to the client.
+> The provider interface **MUST** define a `provider hook` mechanism which can be optionally implemented in order to add `hook` instances to the client.
 
 ```
 class MyProvider implements Provider {
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -111,59 +111,25 @@ ResolutionDetails<number> resolveNumberValue(string flagKey, number defaultValue
 ResolutionDetails<MyStruct> resolveStructureValue(string flagKey, MyStruct defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
 ```
 
-#### Context Transformation
+#### Provider hooks
 
-Feature flag management systems often define structures representing arbitrary contextual data pertaining to the runtime, user, or application. The context transformer defines a simple interface to transform the OpenFeature `evaluation context` to such a structure, mapping values appropriately.
-
-See [evaluation context](./03-evaluation-context.md).
+Feature flag management systems often need to transform the context structures the user provides or generally address lifecycle concerns. A `provider hook` exposes a mechanism for providers to register [`Hooks`](./04-hooks.md) to tap into various stages of the flag evaluation lifecycle.
 
 ##### Requirement 2.10
 
-> The provider interface **MAY** define a `context transformer` method or function, which can be optionally implemented in order to transform the `evaluation context` prior to flag value resolution.
+> The provider interface **MAY** define a `provider hook` mechanism which can be optionally implemented in order to add `hook` instances to the client.
 
-The OpenFeature `client` might apply the transformer function before passing the returned value (the `transformed context`) to the provider resolution methods, thus allowing the provider implementation to avoid implementing and calling such transformation logic repeatedly in flag value resolution methods.
-
-```typescript
+```
 class MyProvider implements Provider {
   //...
 
-  // implementation of context transformer
-  MyProviderContext transformContext(EvaluationContext context) {
-    return new MyProviderContext(context.email, context.ip, context.httpMethod);
+  private readonly hooks: Hook[] = [new MyProviderHook()];
+
+  // ..or alternatively..
+  getProviderHooks(): Hook[]  {
+    return [new MyProviderHook()];
   }
 
   //...
-}
-```
-
-See [evaluation context](./03-evaluation-context.md), [flag evaluation](./01-flag-evaluation.md#flag-evaluation).
-
-##### Condition 2.11
-
-> The implementation language supports generics (or an equivalent feature).
-
-###### Conditional Requirement 2.11.1
-
-> If the implementation includes a `context transformer`, the provider **SHOULD** accept a generic argument (or use an equivalent language feature) indicating the type of the transformed context.
->
-> If such type information is supplied, more accurate type information can be supplied in the flag resolution methods.
-
-```typescript
-// an example implementation in a language supporting interfaces, classes, and generics
-// T represents a generic argument for the type of the transformed context
-interface Provider<T> {
-
-  //...
-
-  // context transformer signature
-  T transformContext(EvaluationContext context);
-
-  //...
-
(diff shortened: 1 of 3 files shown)


specification/sections/02-providers.md
```diff
@@ -111,59 +111,22 @@ ResolutionDetails<number> resolveNumberValue(string flagKey, number defaultValue
 ResolutionDetails<MyStruct> resolveStructureValue(string flagKey, MyStruct defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
 ```
 
-#### Context Transformation
+#### Provider hooks
 
-Feature flag management systems often define structures representing arbitrary contextual data pertaining to the runtime, user, or application. The context transformer defines a simple interface to transform the OpenFeature `evaluation context` to such a structure, mapping values appropriately.
-
-See [evaluation context](./03-evaluation-context.md).
+Feature flag management systems often need to transform the context structures the user provides or general address lifecycle concerns. A `provider hook` provides a mechanism for providers to register [`Hook`s](./04-hooks.md) to tap into that lifecycle.
```
beeme1mr, line 116:
```suggestion
Feature flag management systems often need to transform the context structures the user provides or generally address lifecycle concerns. A `provider hook` exposes a mechanism for providers to register [`Hooks`](./04-hooks.md) to tap into various stages of the flag evaluation lifecycle.
```

specification/sections/02-providers.md
```diff
@@ -111,59 +111,22 @@ ResolutionDetails<number> resolveNumberValue(string flagKey, number defaultValue
...
 
 ##### Requirement 2.10
 
-> The provider interface **MAY** define a `context transformer` method or function, which can be optionally implemented in order to transform the `evaluation context` prior to flag value resolution.
-
-The OpenFeature `client` might apply the transformer function before passing the returned value (the `transformed context`) to the provider resolution methods, thus allowing the provider implementation to avoid implementing and calling such transformation logic repeatedly in flag value resolution methods.
+> The provider interface **MAY** define a `provider hook` method or function, which can be optionally implemented in order to add `hook` instances to the client.
```
beeme1mr, line 120: I'm not sure if this needs to be a method or function. For example, in TypeScript it could just be a property on the class.

specification/sections/02-providers.md
```diff
@@ -111,59 +111,22 @@ ResolutionDetails<number> resolveNumberValue(string flagKey, number defaultValue
...
 
 ```typescript
 class MyProvider implements Provider {
   //...
 
-  // implementation of context transformer
-  MyProviderContext transformContext(EvaluationContext context) {
-    return new MyProviderContext(context.email, context.ip, context.httpMethod);
+  List<Hook> providerHooks() {
```
beeme1mr, line 126: private readonly hook: Hook[] = [new MyProviderHook()];

beeme1mr (approved):
LGTM

beeme1mr:
Hey @justinabrahms, could you please signoff your commit?

justinabrahms:
Signed off again.

specification/sections/02-providers.md
```diff
@@ -111,59 +111,25 @@ ResolutionDetails<number> resolveNumberValue(string flagKey, number defaultValue
 ResolutionDetails<MyStruct> resolveStructureValue(string flagKey, MyStruct defaultValue, context: EvaluationContext, options: FlagEvaluationOptions);
 ```
 
-#### Context Transformation
+#### Provider hooks
 
-Feature flag management systems often define structures representing arbitrary contextual data pertaining to the runtime, user, or application. The context transformer defines a simple interface to transform the OpenFeature `evaluation context` to such a structure, mapping values appropriately.
-
-See [evaluation context](./03-evaluation-context.md).
+Feature flag management systems often need to transform the context structures the user provides or generally address lifecycle concerns. A `provider hook` exposes a mechanism for providers to register [`Hooks`](./04-hooks.md) to tap into various stages of the flag evaluation lifecycle.
```
toddbaert, line 116: I think context transformation is a key use case for provider hooks, so it's worth mentioning, however they are just as flexible as other hooks. I think this example should come after the description... So basically reverse the order of these 2 sentences.

specification/sections/02-providers.md
```diff
@@ -111,59 +111,25 @@ ResolutionDetails<number> resolveNumberValue(string flagKey, number defaultValue
...
-
-```typescript
-class MyProvider implements Provider {
-  //...
-
-  // implementation of context transformer
-  MyProviderContext transformContext(EvaluationContext context) {
-    return new MyProviderContext(context.email, context.ip, context.httpMethod);
-  }
+> The provider interface **MAY** define a `provider hook` mechanism which can be optionally implemented in order to add `hook` instances to the client.
```
toddbaert, line 120: I think we need to state that the provider abstraction MUST support attaching hooks, though obviously the provider author doesn't need to attach them to their provider. I think having some implementations not supporting provider hooks will be problematic, just too inconsistent.

toddbaert (commented):
I think we need to modify 4.4.1 to include provider hooks... (Also, I'm not sure what it's trying to say when it mentions flag evaluation options)

specification/sections/02-providers.md
toddbaert, line 116: I think context transformation is a key use case for provider hooks, so it's worth mentioning, however they are just as flexible as other hooks. If we're going to make this section about provider hooks generally,I think this example should come after the description... So basically reverse the order of these 2 sentences and highlight that context transformation is just one example use case.

weyert:
So we are now adding support for provider hooks which appears to call `getProviderHooks` on the provider implementation. I am wondering what happens when someone would calls`OpenFeature.clearHooks()`? 

Would this also remove these provider hooks or only the user added hooks? 

Or is my assumption incorrect that the `getProviderHooks` will register the defined hooks in `OpenFeature`-singleton or are two separate things?

justinabrahms:
@weyert Great call out. I would expect that this wouldn't remove provider hooks, which may be necessary for a provider to function.  Does that make sense to you?

s-sen:
What is the use case of removing Provider hook? This hook is optionally supplied by the Provider and the app developer has no knowledge of it. Certain default features carried out by provider hook can be overridden by the app developer using hook hints if the provider chooses to document such option. But clear hook operation initiated by the app developer should not remove provider hooks.

toddbaert:
It seems like we've basically identified application author-facing hooks and provider hooks. The clear hooks method is meant for the application author and probably shouldn't impact the provider hooks which the application author isn't aware of 

I'm not sure the clear hooks method is even in the spec (which IMO is fine).

toddbaert (approved):

skyerus (approved):

beeme1mr:
Hey @justinabrahms, we should be good to merge this PR once all conversations have been resolved. Once this is merged, I plan on releasing v0.2.0.

thomaspoignant (approved):

agentgonzo:
I'm late to the party here (I was off on holiday).

```
Hooks MUST be evaluated in the following order: - before: API, Client, Invocation, Provider - after: Provider, Invocation,
```
When talking about context transformation, why are we recommending that it is done via a hook? Why not just recommend having the provider layer do the transformation within the [Flag Evaluation stage](https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#flag-evaluation)?

As I understand it, the vender-specific format of the evaluation context is only needed by the vendor's SDK. As such, it (IMO) should be encapsulated within the provider layer. 

By transforming the context as a hook, we are introducing encapsulation leakage and passing the transformed context back to the OpenFeature SDK (only for it to be passed back into the provider layer by the [flag evaluation stage](https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#flag-evaluation)). 

Would it not be simpler (and better encapsulated) to leave the context transformation entirely within the flag evaluation stage of the provider?


WRT to the original messages on Slack about adding a provider hook to allow setting additional parameters to the context so that they can be logged by OpenTelemetry - that makes perfect sense and it's a good idea to add the vendor hooks as stated.

I'm just thinking that it's better to recommend transforming the context as part of the provider's Flag Evaluation stage rather than a hook.

agentgonzo:
Additionally, I think it would be better to change the `MUST` for the provider hook to be a `MAY`. I don't like enforcing behaviour that may end up being a no-op.

If we change it to `MAY`, then we can say "if you provide a hook, it WILL be called", but if you don't provide a hook, then the OpenFeature SDK will just skip over to the next stage. This is easily provided by the SDK itself as a default no-op call.

@justinabrahms @beeme1mr ?

toddbaert:
>By transforming the context as a hook, we are introducing encapsulation leakage and passing the transformed context back to the OpenFeature SDK (only for it to be passed back into the provider layer by the [flag evaluation stage](https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#flag-evaluation)).

This is a good observation, and I think I agree with your conclusion, and discussions yesterday with @beeme1mr and @justinabrahms kinda elucidated the same problem.

I think we can keep the ordering as is, since provider-related side effects should happen as close to the provider as possible, but we should remove the additional recommendation that context transformation be done in hooks.

> I'm just thinking that it's better to recommend transforming the context as part of the provider's Flag Evaluation stage rather than a hook.

Sounds like we're on the same page.

justinabrahms:
> Additionally, I think it would be better to change the `MUST` for the provider hook to be a `MAY`. I don't like enforcing behaviour that may end up being a no-op.

In java, we're just providing a default noop implementation as part of the interface. I don't feel strongly on this wording.

toddbaert:
https://github.com/open-feature/spec/pull/129 Here is the wording change, as well as a removal of the entire `context transformer` concept, which so far has only been adopted in the JS SDK.

PR #120, opened 2022-08-02
evaluation context requirement to allow fetching all custom fields

rgrassian-split:


specification/sections/03-evaluation-context.md
```diff
@@ -32,6 +32,10 @@ The targeting key uniquely identifies the subject (end-user, or client service)
 
 see: [structure](../types.md#structure), [datetime](../types.md#datetime)
 
+#### Requirement 3.1.3
+
+> The evaluation context **MUST** support fetching the custom fields by key and also fetching all of the keys and values.
```
justinabrahms, line 37:
```suggestion
> The evaluation context **MUST** support fetching the custom fields by key and also fetching all key value pairs.
```

justinabrahms (approved):

specification/sections/03-evaluation-context.md
```diff
@@ -32,6 +32,10 @@ The targeting key uniquely identifies the subject (end-user, or client service)
 
 see: [structure](../types.md#structure), [datetime](../types.md#datetime)
 
+#### Requirement 3.1.3
+
+> The evaluation context **MUST** support fetching the custom fields by key and also fetching all key value pairs.
```
kinyoklion, line 37: Could this extend to how the evaluation context works with `structure` data. In that something that is a `structure` should allow accessing all the keys, or accessing a value by a key?

https://github.com/open-feature/dotnet-sdk/issues/25

beeme1mr: @benjiro would you be able to answer that?

beeme1mr (approved):

specification/sections/03-evaluation-context.md
rgrassian-split: Yes this should be for all of the types, I can make it more specific like 
```suggestion
> The evaluation context **MUST** support fetching the custom fields by key and also fetching all key value pairs for each of the types (`boolean | string | number | datetime | structure`).
```
This would mean for each of the types there needs to be a method to access all of the key/value pairs, as well as a value by key, including for structure.
Does this answer the question? As far as fetching by nested keys like listed in the issue you linked it wouldn't add more functionality for that and is probably separate from this

kinyoklion: Ok. So you could get a structure by a key, but not get a keys from a structure, or values by key from a structure. It is a similar problem, but I could see how it probably needs to be an independent issue.

thomaspoignant (approved):

james-milligan (approved):

thomaspoignant (changes requested):
You don't have update the `specification.json` file, could you run `make` to have it in the PR ?

beeme1mr:
@rgrassian-split, please rerun the generator when you have a moment.

PR #121, opened 2022-08-03
EvaluationContext each key should be unique

thomaspoignant:
Following the discussion [here](https://cloud-native.slack.com/archives/C0344AANLA1/p1659469589273419) and this [issue](https://github.com/open-feature/java-sdk/issues/40), I believe it is necessary to explicitly say that each key from the `EvaluationContext` should be unique.

### DIFF
--- a/specification/sections/03-evaluation-context.md
+++ b/specification/sections/03-evaluation-context.md
@@ -30,9 +30,13 @@ The targeting key uniquely identifies the subject (end-user, or client service) 
 
 > The evaluation context **MUST** support the inclusion of custom fields, having keys of type `string`, and values of type `boolean | string | number | datetime | structure`.
 
-Each key should be unique.
-
 see: [structure](../types.md#structure), [datetime](../types.md#datetime)
+
+#### Requirement 3.1.4
+
+> The evaluation context fields **MUST** have a unique key.
+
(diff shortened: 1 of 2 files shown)


specification/sections/03-evaluation-context.md
```diff
@@ -30,6 +30,8 @@ The targeting key uniquely identifies the subject (end-user, or client service)
 
 > The evaluation context **MUST** support the inclusion of custom fields, having keys of type `string`, and values of type `boolean | string | number | datetime | structure`.
 
+Each key should be unique.
```
beeme1mr, line 33: Please move this to a new requirement called "Requirement 3.1.3" and make sure the requirement is under a markdown quote block. You'll need to run `make parse` to update the requirements JSON file.

Please also define the behavior of a collision.

beeme1mr: I just noticed that there's a PR to introduce 3.1.3. In that case, perhaps use 3.1.4. Having a dedicated section will make it easier for developers to confirm that they've explicitly tested this behavior.

justinabrahms (approved):

beeme1mr (approved):
Nice! Thanks for formalizing this in the spec.

InTheCloudDan (approved):

PR #125, opened 2022-08-08
Document context merging order for before hook ctx

### LINKED ISSUE #124: Hook context merging w/ api ctx
Requirement 4.3.4 says:
> When before hooks have finished executing, any resulting evaluation context MUST be merged with the invocation evaluation context with the invocation evaluation context taking precedence in the case of any conflicts.

How does provider and client context fix here?

I think the order that makes sense is (top-most wins):
1. invocation
2. before hook
3. client
4. api

That sound right?

justinabrahms:
Fixes #124

### DIFF
--- a/specification/sections/04-hooks.md
+++ b/specification/sections/04-hooks.md
@@ -87,7 +87,7 @@ EvaluationContext | void before(HookContext, HookHints);
 
 #### Requirement 4.3.4
 
-> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the invocation `evaluation context` with the invocation `evaluation context` taking precedence in the case of any conflicts.
+> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the other `evaluation context`s with the following order: invocation (highest precedence), before-hook, client, api (lowest precedence).
 
 #### Requirement 4.3.5
 
--- a/specification/sections/04-hooks.md
+++ b/specification/sections/04-hooks.md
@@ -87,7 +87,7 @@ EvaluationContext | void before(HookContext, HookHints);
 
 #### Requirement 4.3.4
 
-> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the other `evaluation context`s with the following order: invocation (highest precedence), before-hook, client, api (lowest precedence).
+> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the existing `evaluation context` in the following order: invocation (highest precedence), before-hook, client, api (lowest precedence).
 
 #### Requirement 4.3.5
 
--- a/specification/sections/04-hooks.md
+++ b/specification/sections/04-hooks.md
@@ -87,7 +87,7 @@ EvaluationContext | void before(HookContext, HookHints);
 
 #### Requirement 4.3.4
 
-> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the existing `evaluation context` in the following order: invocation (highest precedence), before-hook, client, api (lowest precedence).
+> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the existing `evaluation context` in the following order: before-hook (highest precedence), invocation, client, api (lowest precedence).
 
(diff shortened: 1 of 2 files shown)


specification/sections/04-hooks.md
```diff
@@ -87,7 +87,7 @@ EvaluationContext | void before(HookContext, HookHints);
 
 #### Requirement 4.3.4
 
-> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the invocation `evaluation context` with the invocation `evaluation context` taking precedence in the case of any conflicts.
+> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the other `evaluation context`s with the following order: invocation (highest precedence), before-hook, client, api (lowest precedence).
```
toddbaert, line 90:
```suggestion
> When `before` hooks have finished executing, any resulting `evaluation context` **MUST** be merged with the existing `evaluation context` in the following order: invocation (highest precedence), before-hook, client, api (lowest precedence).
```

toddbaert: Is this point a bit redundant with the presence of [3.2.2](https://github.com/open-feature/spec/blob/main/specification/sections/03-evaluation-context.md#requirement-322)? Can we / should we unify these somehow?

Do we also need to mention provider hooks?

toddbaert (approved):
I dont love that this is a bit redundant with [3.2.2](https://github.com/open-feature/spec/blob/main/specification/sections/03-evaluation-context.md#requirement-322), but that was true before this change and I'm not sure the best way to resolve it. I think this is an improvement.

beeme1mr (approved):

toddbaert:
I think this made perfect sense before provider hooks, but now with them in the mix, things get more complicated.

One of the primary use-cases of provider hooks is context transformation... if a provider hook does significant context transformation (say, into some vendor-specific object, or by serializing values in some required way) then it has to have the highest merge priority, right?

So do we do:

(top-most wins):
1. before hook
2. invocation
3. client
4. api

or 

(top-most wins):
2. provider before hooks
2. invocation
3. before hook
4. client
5. api

justinabrahms:
Huh. I don't like our options now. :-P 

You're thinking that a context transformation use-case is going to set an additional property on the context that's specific for the provider?

How about this (top-most wins)?

1. provider before hooks
2. invocation hooks
3. invocation context
4. client before hook
5. client context
6. api before hook
7. api context

toddbaert:
> You're thinking that a context transformation use-case is going to set an additional property on the context that's specific for the provider?

Well yes, but to be honest, I was thinking something even more potentially problematic - our context transformer allowed the context object to be an entirely new object/instance - for example an [LDUser](https://launchdarkly.github.io/java-server-sdk/com/launchdarkly/sdk/LDUser.html). I'm not sure that was understood :sweat_smile: . As you can imagine merging after that transformation is completely impossible.

At the very least, I think provider hooks may transform the context in a way to maintain compatibility with their particular backend, so they need to "win".

toddbaert:
Thinking about it more, considering implementation difficulty and cognitive burden on authors, I think I'm in favor of:

(top-most wins):
1. before hooks (all of them, including provider hooks)
2. invocation
3. client
4. api

So hooks will always win... I think this is OK and simple to understand. If the author REALLY wants to override something in their invocation, they can use an invocation before hook to do that, which will override everything except the provider hooks (which is ideal).

@justinabrahms @beeme1mr  what do you guys think about that?

justinabrahms:
I think I could be into that. So we're saying that:
- We don't really expect you to specify content in your invocation context that will collide w/ things the hooks provide.
- If you do have a conflict you want to exclude, we expect you to define an inline hook.
- You cannot override before hook values from a provider.

toddbaert:
> I think I could be into that. So we're saying that:
> 
> * We don't really expect you to specify content in your invocation context that will collide w/ things the hooks provide.
> * If you do have a conflict you want to exclude, we expect you to define an inline hook.
> * You cannot override before hook values from a provider.

Exactly.

toddbaert (approved):

thomaspoignant (approved):

PR #127, opened 2022-08-08
update fractional evaluation glossary definition

beeme1mr:
Signed-off-by: Michael Beemer <michael.beemer@dynatrace.com>

### DIFF
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -156,7 +156,7 @@ A string logically identifying the subject of evaluation (end-user, service, etc
 
 ### Fractional Evaluation
 
-Assigning flag values randomly according to a configured proportion or percentage (ie: 50/50).
+Pseudorandomly resolve flag values using a context property, such as a targeting key, based on a configured proportion or percentage (ie: 50/50).
 
 ### Rule
 


justinabrahms (approved):

toddbaert (approved):

specification/glossary.md
```diff
@@ -156,7 +156,7 @@ A string logically identifying the subject of evaluation (end-user, service, etc
 
 ### Fractional Evaluation
 
-Assigning flag values randomly according to a configured proportion or percentage (ie: 50/50).
+Pseudorandomly resolve flag values using a context property, such as a targeting key, based on a configured proportion or percentage (ie: 50/50).
```
thomaspoignant, line 159: Should we say that the value should be always the same for the a user?

skyerus (approved):

specification/glossary.md
toddbaert: I think that might be too detailed, but that's just my opinion. I like to keep the subject (user/client) vague generally, because I think there are valid non-user driven use cases... clients are not necessarily end-users.

I think this gets the idea across that some property (possibly a user identifier) will allow the split to be deterministic, so that it will be the same for that user over time. We could go into more detail in the doc, perhaps.

thomaspoignant: Yes the glossary is probably not the best place to put this, we may consider to put that in the spec.

thomaspoignant (approved):

PR #128, opened 2022-08-09
Provider error handling

justinabrahms:
I don't think we're really saying they should handle any and all exceptions.. right? The OF SDK should be dealing with exception handling.

### DIFF
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -85,7 +85,7 @@ Possible values vary by provider, but might include such values as `"TARGETING_M
 
 ##### Requirement 2.8
 
-> In cases of abnormal execution, the `provider` **MUST** indicate an error using the idioms of the implementation language, with an associated error code having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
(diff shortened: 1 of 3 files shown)


toddbaert:
> I don't think we're really saying they should handle any and all exceptions.. right? The OF SDK should be dealing with exception handling.

It says the provider should _indicate_ an error, not _handle_ them. As in it should use whatever language idioms are appropriate to signal to the caller that something went wrong. This could be throwing an exception (and not necessarily a new one). The non-normative text belong expands on this

> The provider might throw an exception, return an error, or populate the `error code` object on the returned `flag resolution` structure to indicate a problem during flag value resolution.

specification/sections/02-providers.md
```diff
@@ -85,7 +85,7 @@ Possible values vary by provider, but might include such values as `"TARGETING_M
 
 ##### Requirement 2.8
 
-> In cases of abnormal execution, the `provider` **MUST** indicate an error using the idioms of the implementation language, with an associated error code having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+> In cases of abnormal execution, the `provider` **SHOULD** indicate an error using the idioms of the implementation language, with an associated error code having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
```
toddbaert, line 88: I think in abnormal execution, it must be communicated to the caller in some idiomatic way that an error occurred.

justinabrahms:
Gotcha. That feels reasonable to me.

PR #139, opened 2022-09-22
chore: add additional document statuses

toddbaert:
Adds document statuses explaining stability of various spec documents and sub-sections.

This is important as we move toward's 1.0 SDK releases and a stable spec.

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -35,7 +35,24 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
 
 ## Document Statuses
 
-| Status               | Explanation                   |
-| -------------------- | ----------------------------- |
-| No Explicit "Status" | Equivalent to Experimental.   |
-| Experimental         | Breaking changes are allowed. |
+Documents and subsections within the specification are marked with statuses indicating their stability level. Functionality described in the specification graduates through these statuses with increasing stability. Stability levels apply only to normative sections within the specification; editorial changes to examples and explanations are exempt from these constraints.
+
+Possible statuses are described below:
+
+### Experimental
+
+Specification documents that are marked as `Experimental` contain functionality under active development. Breaking changes are allowed and may be made without deprecation notices or warnings with minor version updates.
+
+### Hardening
+
+Documents marked as `Hardening` describe functionality with an emphasis on stabilizing existing requirements. Breaking changes require consensus by the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members) but may still be made with minor version updates.
+
+### Stable
+
+Sections marked as `Stable` do not allow breaking changes without a major version update.
+
+### Mixed
+
+Specification documents marked as `Mixed` contain subsections of varying statuses specified above.
+
+NOTE: No explicit status = `Experimental`
--- a/specification/README.md
+++ b/specification/README.md
@@ -46,7 +46,7 @@ Possible statuses are described below:
 
 Specification documents that are marked as `Experimental` contain functionality under active development. Breaking changes are allowed and may be made without deprecation notices or warnings with minor version updates. We recommend you use these features in experimental environments and not in production.
 
-Put simply: 
+Put simply:
 
 > We're testing these features out. Things could change anytime.
 
@@ -54,7 +54,7 @@ Put simply:
 
 Documents marked as `Hardening` describe functionality with an emphasis on stabilizing existing requirements. Breaking changes require consensus by the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members) but may still be made with minor version updates. These features are suitable for use in production environments. Feedback is encouraged.
 
-Put simply: 
+Put simply:
 
 > We believe these features are ready for production use, and hope for feedback.
 
@@ -62,7 +62,7 @@ Put simply:
 
 Sections marked as `Stable` do not allow breaking changes without a major version update. They can be used in production with a high degree of confidence.
 
-Put simply: 
+Put simply:
 
 > These features are stable and battle-hardened.
 
--- a/specification/README.md
+++ b/specification/README.md
@@ -60,7 +60,7 @@ Put simply:
 
 ### Stable
 
-Sections marked as `Stable` do not allow breaking changes without a major version update.
+Sections marked as `Stable` do not allow breaking changes without a major version update. They can be used in production with a high degree of confidence.
 
 Put simply: 
 
--- a/specification/README.md
+++ b/specification/README.md
@@ -35,21 +35,36 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
 
 ## Document Statuses
 
-Documents and subsections within the specification are marked with statuses indicating their stability level. Functionality described in the specification graduates through these statuses with increasing stability. Stability levels apply only to normative sections within the specification; editorial changes to examples and explanations are exempt from these constraints.
+Documents and subsections within the specification are marked with statuses indicating their stability level.
+Functionality described in the specification graduates through these statuses with increasing stability.
+Stability levels apply only to normative sections within the specification; editorial changes to examples and explanations are exempt from these constraints.
+It is the responsibility of the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members) to consider and approve the graduation of documents.
 
 Possible statuses are described below:
 
 ### Experimental
 
-Specification documents that are marked as `Experimental` contain functionality under active development. Breaking changes are allowed and may be made without deprecation notices or warnings with minor version updates.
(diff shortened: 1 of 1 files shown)


specification/README.md
```diff
@@ -35,7 +35,24 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
...
+
+Specification documents that are marked as `Experimental` contain functionality under active development. Breaking changes are allowed.
+
+### Hardening
+
+Documents marked as `Hardening` describe functionality with an emphasis on stabilizing existing requirements. Breaking changes require consensus by the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members).
+
+### Stable
+
+Sections marked as `Stable` do not allow breaking changes.
```
dyladan, line 52: I would mention that stable documents can have breaking changes only with major version changes. Never is a long time.

specification/README.md
```diff
@@ -35,7 +35,24 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
...
+
+### Mixed
+
+Specification documents marked as `Mixed` contain subsections of varying statuses specified above.
+
+\* No explicit status = `Experimental`
```
dyladan, line 58: is this an accident?

specification/README.md
```diff
@@ -35,7 +35,24 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
...
-| -------------------- | ----------------------------- |
-| No Explicit "Status" | Equivalent to Experimental.   |
-| Experimental         | Breaking changes are allowed. |
+Documents and subsections within the specification are marked with statuses indicating their stability level. Functionality described in the specification graduates through these statuses with increasing stability. Stability levels apply only to normative sections within the specification; editorial changes to examples and explanations are exempt from these constraints.
+
+Possible statuses are described below:
+
+### Experimental
+
+Specification documents that are marked as `Experimental` contain functionality under active development. Breaking changes are allowed.
```
dyladan, line 44:
```suggestion
Specification documents that are marked as `Experimental` contain functionality under active development. Breaking changes are allowed and may be made without deprecation notices or warnings with minor version updates.
```

specification/README.md
```diff
@@ -35,7 +35,24 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
...
+
+### Hardening
+
+Documents marked as `Hardening` describe functionality with an emphasis on stabilizing existing requirements. Breaking changes require consensus by the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members).
```
dyladan, line 48:
```suggestion
Documents marked as `Hardening` describe functionality with an emphasis on stabilizing existing requirements. Breaking changes require consensus by the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members) but may still be made with minor version updates.
```

dyladan (approved):
overall looks good. made a few minor comments

specification/README.md
toddbaert: No, I wanted it to show up as a literal `*`. Is there a better markdown symbol for this?

specification/README.md
toddbaert: That makes sense to me. @beeme1mr we talked about this. I think we should explicitly add it here though... right?

beeme1mr: Sure

specification/README.md
beeme1mr: I think you can do this:

> NOTE: No explicit status = `Experimental`

agentgonzo (approved):
I'm not sure the benefit of `Mixed`, but it's not like it's a problem

InTheCloudDan:
Is there going to be requirements or guidelines on how the promotion is determined? For example what is required to go from Experimental -> Hardening?

toddbaert:
> Is there going to be requirements or guidelines on how the promotion is determined? For example what is required to go from Experimental -> Hardening?

@InTheCloudDan that's a good question. I think some kind of formality could be good. I'm not sure what that could be. I'll think about it :thinking: 

Anyone else have ideas for this?

toddbaert:
- added simple summary statements
- added recommendations about prod usage
- added statement describing that the TSC approves graduation:

> It is the responsibility of the [Technical Steering Committee](https://github.com/open-feature/community/blob/main/governance-charter.md#tsc-members) to consider and approve the graduation of documents.

@InTheCloudDan  @AloisReitbauer

beeme1mr (approved):

skyerus (approved):

specification/README.md
```diff
@@ -35,7 +35,39 @@ An implementation is not compliant if it fails to satisfy one or more of the "MU
...
+
+Sections marked as `Stable` do not allow breaking changes without a major version update. They can be used in production with a high degree of confidence.
+
+Put simply:
+
+> These features are stable and battle-hardened.
+
+### Mixed
+
+Specification documents marked as `Mixed` contain subsections of varying statuses specified above.
```
kinyoklion, line 71: I am trying to imagine what this transition would look like for one of the sections. If `evaluation API` is stable, for instance, and we want to add a new method to it.

The overall status would become `Mixed` and then the individual new sections would be `Experimental`?
If so a stable -> anything else transition seems undesirable.

toddbaert: > The overall status would become Mixed and then the individual new sections would be Experimental?

I think that's the idea. `Mixed` is a `pseudo-status` that just means you have to look at sections. OTel does something similar: https://github.com/open-telemetry/opentelemetry-specification/blob/e440dd50a16f4ff4f5ff47e420a49129127f7a37/specification/document-status.md

> If so a stable -> anything else transition seems undesirable.

Can you clarify this point? I'm not sure I understand.

kinyoklion: In the `evaluation API` for instance, the transition between experimental -> hardening -> stable would also seem to apply to the concept of their being an `evaluation API`. So, the fact that there is an evaluation API isn't really in contention after it hits stable.

Looking at the OTel spec though, things seem not so consistent with regards to sections. Some things are in an experimental folder, but don't have a status, stable documents mostly don't seem to have statuses on sections. One document has a mixed status, with a section with a mixed status.

I think if the sections actually have statuses, then the top level one is not very useful. 

https://github.com/open-telemetry/opentelemetry-specification/blob/ca593258953811eb33197976b8e565299b874853/specification/metrics/sdk_exporters/otlp.md

Even though this one has something confusing, I do think it shows a concept that could be reasonable. The general section is stable, then the document is mixed.

Overall though I like the idea of a secondary tag like feature-freeze more.

Like this is stable, and being added to, this is stable and feature frozen, this is just stable.

toddbaert: >Overall though I like the idea of a secondary tag like feature-freeze more.

We waffled on the feature-freeze idea for a while, but decided we wanted our spec flow to be unidirectional. Obviously I'm open to discussion here.

> I think if the sections actually have statuses, then the top level one is not very useful.

I can see this argument, and it's why I said here https://github.com/open-feature/spec/pull/139#issuecomment-1261087147 :

> > I'm not sure the benefit of `Mixed`, but it's not like it's a problem
> 
> Ya I was hesitant about Mixed. It's not really a meaningful status unto itself, but more of a `meta-status`. I feel confident about the rest.
> 
> TBH we can probably completely remove it. We can annotate sections with statuses, it just means you'll have navigate to them to see them.

Now that 2 people have brought it into question, I'm considering removing `mixed` and document-level statuses entirely, in favor of just adding statuses for sections. 

What do you think of this proposal?

cc @beeme1mr

moredip: FWIW, I'm +1 on removing the `mixed` status. Seems like the majority of documents are going to end up permanently in that category.

kinyoklion: (I retract support for feature-freeze if it isn't unidirectional. I was thinking more like, this is not accepting new things, basically to allow ancillary stuff to stabilize. Not really that it may become not stable. I was seeing mixed as non-unidirectional. 
mixed->stable->mixed,)

toddbaert: > (I retract support for feature-freeze if it isn't unidirectional. I was thinking more like, this is not accepting new things, basically to allow ancillary stuff to stabilize. Not really that it may become not stable. I was seeing mixed as non-unidirectional.
mixed->stable->mixed,)

Ya, our `hardening` concept is basically a unidirectional `feature freeze`. OTel adds and removes feature freeze in accordance with their priorities at any given time.

> FWIW, I'm +1 on removing the mixed status. Seems like the majority of documents are going to end up permanently in that category.

So that's 3 for nixing `mixed`, which I already though was low value. I think I will make changes to remove that status, and simply specify that statues are apply to document sections/subsections.

toddbaert: I've done the above: https://github.com/open-feature/spec/pull/139/commits/ed9a4bc01ba56529fd93e9e22a585de59d409f9c

kinyoklion (approved):

Issue #141, opened 2022-09-22
Evaluation Context related error status


kinyoklion:
For some providers contexts are going to require certain attributes to function correctly. For instance a targeting key.

It would be good if there was an error code for context related problems, for instance `MALFORMED_CONTEXT`.

This would help to more quickly narrow down usage errors.

toddbaert:
I agree this would be helpful. I'd accept `MALFORMED_CONTEXT`, but I'll also suggest `INVALID_CONTEXT` as another possible alternative. "Malformed" suggests to me something like a structural error, while "invalid" implies either structural or semantic issues (like a missing targeting key). I think either work though.

thomaspoignant:
Yes it would be great to have something like that in the providers to be explicit when a field is missing.

kinyoklion:
Resolved in https://github.com/open-feature/spec/pull/142

PR #142, opened 2022-09-23
feat!: make error code an enum and include optional error message

kinyoklion:
See https://github.com/open-feature/spec/issues/141 for context.

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -185,7 +185,7 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
 
 ##### Requirement 1.4.7
 
-> In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** contain a string identifying an error occurred during flag evaluation and the nature of the error.
+> In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** contain an `error code`.
 
 See [error code](../types.md#error-code) for details.
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -213,7 +213,7 @@ It's recommended to provide non-blocking mechanisms for flag evaluation, particu
 
 ##### Requirement 1.4.12
 
-> In cases of abnormal execution, the `evaluation details` structure's `error message` field **MAY** contain a string containing additional detail about the nature of the error.
+> In cases of abnormal execution, the `evaluation details` structure's `error message` field **MAY** contain a string containing additional details about the nature of the error.
 
 #### Evaluation Options
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -187,7 +187,7 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
 
 > In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** contain a string identifying an error occurred during flag evaluation and the nature of the error.
 
-Some example error codes include: `"TARGETING_KEY_MISSING"`, `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+Some example error codes include: `"TARGETING_KEY_MISSING"`, `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, `"INVALID_CONTEXT"`, or `"GENERAL"`.
 
 ##### Requirement 1.4.8
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -187,7 +187,7 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
 
 > In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** contain a string identifying an error occurred during flag evaluation and the nature of the error.
 
-Some example error codes include: `"TARGETING_KEY_MISSING"`, `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, `"INVALID_CONTEXT"`, or `"GENERAL"`.
+See [error codes](../types.md#error-code) for details.
 
 ##### Requirement 1.4.8
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -187,7 +187,7 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
 
 > In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** contain a string identifying an error occurred during flag evaluation and the nature of the error.
 
-See [error codes](../types.md#error-code) for details.
+See [error code](../types.md#error-code) for details.
 
 ##### Requirement 1.4.8
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -211,6 +211,10 @@ Implementations may define a standard logging interface that can be supplied as 
 
 It's recommended to provide non-blocking mechanisms for flag evaluation, particularly in languages or environments wherein there's a single thread of execution.
 
+##### Requirement 1.4.12
+
+> In cases of abnormal execution, the `evaluation details` structure's `error message` field **MAY** contain a string containing additional detail about the nature of the error.
+
 #### Evaluation Options
 
 ##### Requirement 1.5.1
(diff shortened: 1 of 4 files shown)


specification/sections/01-flag-evaluation.md
```diff
@@ -187,7 +187,7 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
 
 > In cases of abnormal execution, the `evaluation details` structure's `error code` field **MUST** contain a string identifying an error occurred during flag evaluation and the nature of the error.
 
-Some example error codes include: `"TARGETING_KEY_MISSING"`, `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+Some example error codes include: `"TARGETING_KEY_MISSING"`, `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, `"INVALID_CONTEXT"`, or `"GENERAL"`.
```
kinyoklion, line 190: I am not really sure if these are examples if they are required error codes. Perhaps wording from an earlier point where they were unspecified.

toddbaert: Ya I wonder if we shouldn't firm these up as well, and rigidly defined these. Perhaps we can use this PR for this purpose.

I think these are a bit different than `reason`. I'm more inclined to strictly enumerate them.

Also, do we need `TARGETING_KEY_MISSING` with `INVALID_CONTEXT`?

What are your thoughts?

kinyoklion: I do see how `TARGETING_KEY_MISSING` could be useful, as it is the only commonly defined field, but it is a subset of an `INVALID_CONTEXT`. 

I do think it would be good to have these enumerated in a table with a basic description.

toddbaert: @kinyoklion would you mind adding such a table? I added one recently in `types.md`.

But I also wonder if we should explicitly require the enumeration of these error codes in implementations. In other words, to requires this not be a free-form string. I think most SDKS have implemented it this way anyway:

https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/javasdk/ErrorCode.java

https://github.com/open-feature/dotnet-sdk/blob/main/src/OpenFeatureSDK/Constant/ErrorType.cs

kinyoklion: Yeah, I can make a table, and possibly clarify the language.

To me this sounded like it was specifically enumerated. 

MUST combined with `having possible values`.
```
In cases of abnormal execution, the `provider` **MUST** indicate an error using the idioms of the implementation language, with an associated error code having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, `"INVALID_CONTEXT"`, or `"GENERAL"`.
```

toddbaert: You're right, I think `Some example error codes include` in the section you've added your new code to sounds a bit too lax (though it's not technically a normative section). I think your table can help clarify these should be well-defined.

kinyoklion: I think I see part of the problem/complexity here.

We have `resolution details` and `evaluation details`. The `error code` for `resolution details` was specifically enumerated. The `error code` for `evaluation details` was not.

Should they be the same error code? Or should `evaluation details` allow for client error codes as a possibility? The error arose in the client instead of in the provider. (There are no such errors currently.)

toddbaert: > Should they be the same error code?

The intent was for them to be the same error codes. That ambiguity should be corrected.

kinyoklion: I've started down the path of centralizing this, but I have a couple concerns. 

In the dotnet SDK the `evaluation details` contains a string error, and the `resolution details` contains an enumeration. I think that there are good reasons that someone may do this.

Primarily because extending an enumeration is often breaking.

The second item is more of a question. Do we just remove the enumerated list from the requirement and expect it will come from the type.

agentgonzo (approved):

specification/types.md
```diff
@@ -62,6 +62,20 @@ A set of pre-defined reasons is enumerated below:
...
+
+An enumerated error code represented idiomatically in the implementation language.
+
+| Error Code            | Explanation                                                                                 |
+|-----------------------|---------------------------------------------------------------------------------------------|
+| PROVIDER_NOT_READY    | The value was resolved before the provider was ready.                                       |
+| FLAG_NOT_FOUND        | The flag could not be found.                                                                |
+| PARSE_ERROR           | An error was encountered parsing data, such as a flag configuration.                        |
+| TYPE_MISMATCH         | The type of the flag value does not match the expected type.                                |
+| TARGETING_KEY_MISSING | The provider requires a targeting key and one was not provided in the `evaluation context`. |
```
justinabrahms, line 75: Is this different than the invalid context one?

specification/sections/01-flag-evaluation.md
thomaspoignant: After thinking a bit more about it, I think it is good to have a standard _(aka an enum)_ for compatibility reasons, it will make it easier to switch from one provider to another.

toddbaert: > The second item is more of a question. Do we just remove the enumerated list from the requirement and expect it will come from the type.

I think that's fine if we specifically reference the type in the normative section (not with a link, but the name of the structure).

> In the dotnet SDK the evaluation details contains a string error, and the resolution details contains an enumeration. I think that there are good reasons that someone may do this.

> Primarily because extending an enumeration is often breaking.

hmmm. I'm not sure about this. I think I see your point about a string being a less brittle solution, but we also chose to make these an enum because we anticipate it's more likely that logic might written around these values. Maybe that's more true for providers and less true for the evaluation API (the part of the SDK `application authors` see)?

I'm interested in other's thoughts on this one. I'm not opposed to either direction, but I want to make the optimal choice.

kinyoklion: I think an enumeration here could be fine. We will just want to be careful extending it in the future. In many languages extending an enumeration, that is provided by an API, would be considered a breaking change (versus one which is consumed by an API, which would not.)

moredip: Maybe this is already part of the discussion and I missed it, but would we consider having a standard enum with fixed values, but also allow the provider to optionally include a free-form string. This would give the provider the option of giving additional information to a developer debugging an error. 

Rather than just `PROVIDER_NOT_READY`, a developer could also see "_You are attempting to evaluate a flag before the FlagsRUs client has performed its initial config synchronization. See https://docs.flagsrus.io/errors/provider_not_ready for more information_"

kinyoklion: We've not talked about it as a part of this change, but I have had the same thought. Prior to this change I have many errors mapping to `General`. I've needed to implement logging in my provider to provide the breadcrumbs needed to inform the user what has gone wrong. 

Even with this I will have many that map to `INVALID_CONTEXT`, because I have a number of context fields, that if included, need to be of specific types. It could be nice to pass an associated error message/description with debug detail.

toddbaert: Interesting. Would this message only be present in error cases?

kinyoklion: That would fit my expectation. Optional string that may be present in the case of abnormal execution (errorMessage/errorDetail/errorDescription). Does that sound right, @moredip?

beeme1mr: That seems reasonable to me. It could look something like this in the JS SDK:

// Error thrown in the provider
```typescript
throw new ParseError("Excepted string response type but received Boolean");
```

// In response details object
```json
{
  ":value": "the default value",
  "reason": "ERROR",
  "errorCode": "PARSE_ERROR",
  "errorMessage": "Expected a string response type but received a boolean"
}
```

toddbaert: I like this a lot.

kinyoklion: Do we want to tackle this in this PR, or a followup?

toddbaert: I think we should do it here since the discussion is here, as long as you don't mind the scope creep! :sweat_smile:

specification/sections/02-providers.md
```diff
@@ -85,10 +85,12 @@ As indicated in the definition of the [`flag resolution`](../types.md#resolution
 
 ##### Requirement 2.8
 
-> In cases of abnormal execution, the `provider` **MUST** indicate an error using the idioms of the implementation language, with an associated error code having possible values `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"PARSE_ERROR"`, `"TYPE_MISMATCH"`, or `"GENERAL"`.
+> In cases of abnormal execution, the `provider` **MUST** indicate an error using the idioms of the implementation language, with an associated `error code`.
```
kinyoklion, line 88: I feel like there should be something about the `error message` here as well. Or a similar requirement that includes the message. I am not sure how that would be forumated.

specification/sections/01-flag-evaluation.md
kinyoklion: I've taken a stab at it.

beeme1mr (approved):
Looks good, thanks!

specification/types.md
justinabrahms: This seems fine to me.

benjiro (approved):

thomaspoignant (approved):

Issue #146, opened 2022-09-29
Proposed timeline for spec and SDK stability


toddbaert:
As discussed in our Sept 29 community meeting, we'd like to solidify the spec and the SDKs. This may (or may not) involve releasing a 1.0 of each SDK, but will certainly involve adding some stability guarantees to the spec (see https://github.com/open-feature/spec/pull/139).

The timeline proposed is:

#### Oct 7th:
  - breaking changes are in for all SDKs, and SDKs are compliant with 0.5.0.
  - `evaluation API` and `provider` spec documents are marked [Hardening](https://github.com/open-feature/spec/blob/e2629c1873f2fdb615f43c545835f4d33b7e30ec/specification/README.md#hardening)
#### Oct 14th:
- all providers updated with the SDK breaking changes, compliant with latest respective SDK
#### Oct 17th:
- pending the successful updates of a majority of providers, we release SDKs with enhanced stability guarantees
- this could mean SDKs release 1.0 versions, 1.0rcs, or perhaps just explicitly state they're ready for stability testing (interested in thoughts on this).

Note: The stability guarantees linked above do not prevent us from adding new features, just from breaking old ones. This effort is to provide assurances around the use of the existing interfaces, particularly `providers` and the application-author-facing `Evaluation API`.

beeme1mr:
The spec now has explicit stability guarantees and 1.0 SDKs have been released.

PR #149, opened 2022-10-13
chore: section clarity, harden API, providers

toddbaert:
**Marking Evaluation API and Providers `hardening`** and a few non-functional changes here that improve structure. 

specifically:

- explicitly mark and number all sections
- section headings have consistent numbering  (provider section numbering was different than the other docs)
- removed "draft" language 

An alternative to marking all of `provider` and `evaluation API` as hardening would be just marking all existing sections therein as hardening, which might be better since we'd have to do that anyway if we added a new experimental section.

see: https://github.com/open-feature/spec/issues/146 for more on release goals.

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -185,28 +185,28 @@
             "children": []
         },
         {
-            "id": "Requirement 2.1",
-            "machine_id": "requirement_2_1",
+            "id": "Requirement 2.1.1",
+            "machine_id": "requirement_2_1_1",
             "content": "The provider interface MUST define a `metadata` member or accessor, containing a `name` field or accessor of type string, which identifies the provider implementation.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 2.2",
-            "machine_id": "requirement_2_2",
+            "id": "Requirement 2.2.2",
+            "machine_id": "requirement_2_2_2",
             "content": "The `feature provider` interface MUST define methods to resolve flag values, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required) and `evaluation context` (optional), which returns a `flag resolution` structure.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Condition 2.3",
-            "machine_id": "condition_2_3",
+            "id": "Condition 2.2.3",
+            "machine_id": "condition_2_2_3",
             "content": "The implementing language type system differentiates between strings, numbers, booleans and structures.",
             "RFC 2119 keyword": null,
             "children": [
                 {
-                    "id": "Conditional Requirement 2.3.1",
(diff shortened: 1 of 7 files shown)


specification.json
```diff
@@ -185,101 +185,101 @@
...
+            "machine_id": "requirement_2_1_1",
             "content": "The provider interface MUST define a `metadata` member or accessor, containing a `name` field or accessor of type string, which identifies the provider implementation.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
-            "id": "Requirement 2.2",
-            "machine_id": "requirement_2_2",
+            "id": "Requirement 2.2.2",
+            "machine_id": "requirement_2_2_2",
```
beeme1mr, line 196: Shouldn't this be 2.2.1?

specification/sections/01-flag-evaluation.md
```diff
@@ -4,15 +4,15 @@ description: The specification that defines the developer facing feature flag ev
 toc_max_heading_level: 4
 ---
 
-# Flag Evaluation API
+# 1. Flag Evaluation API
```
beeme1mr, line 7: Is a number required here? If so, can it just be `1` instead of `1.`?

beeme1mr (commented):
Changing the spec version numbers may require numerous changes to tests in various SDKs. That doesn't necessary mean we can make the change but we should be aware of the impact.

specification.json
beeme1mr: This change will also impact the remaining 2.2.X IDs.

specification/sections/01-flag-evaluation.md
toddbaert: I think everything we apply a level to needs a number.

`1 Flag Evaluation API` looked very weird to me, and most section numbers in a sectioned document end with a `.` see https://datatracker.ietf.org/doc/html/rfc6749#section-3 for example.

specification.json
toddbaert: Good catch, fixed.

toddbaert:
> Changing the spec version numbers may require numerous changes to tests in various SDKs. That doesn't necessary mean we can make the change but we should be aware of the impact.

100%. I know this changes some tests, but this change should actually reduce such thrashing in the future and makes things more consistent. Provider was the only section where the conditions didn't have 3 decimal places.

toddbaert:
An alternative to marking all of provider and evaluation API as hardening would be just marking all existing sections as hardening, which might be better since we'd have to do that anyway if we added a new experimental section.

benjiro (approved):

tcarrio (approved):

agentgonzo (approved):

beeme1mr (approved):

toddbaert:
@justinabrahms do you have any objections to this?

PR #150, opened 2022-10-17
chore: clarify timezone requirement

toddbaert:
A small change to make the timezone info on the `Datetime` type optional. Some SDKs already do not include timezone info.

Since it's loosening this definition, this is non-breaking.

### DIFF
--- a/specification/types.md
+++ b/specification/types.md
@@ -28,7 +28,7 @@ Structured data, presented however is idiomatic in the implementation language, 
 
 ### Datetime
 
-A language primitive for representing a date and time, including timezone information.
+A language primitive for representing a date and time, optionally including timezone information.
 
 ### Evaluation Details
 
--- a/specification/types.md
+++ b/specification/types.md
@@ -28,7 +28,7 @@ Structured data, presented however is idiomatic in the implementation language, 
 
 ### Datetime
 
-A language primitive for representing a date and time, optionally including timezone information.
+A language primitive for representing a date and time, optionally including timezone information. If no timezone is specified, the date and time will be treated as UTC.
 
 ### Evaluation Details
 


agentgonzo:
Which SDKs don't support timezones ? Is it that they "don't" (and can thus be fixed to include them) or "can't" by virtue of the language.

I struggle to believe that they can't. 

Ambiguous times are a terrible terrible thing to put in at the design stage. At the very least, the spec should explicitly state that if the timezone is not present, it WILL be treated at UTC.

toddbaert:
> Which SDKs don't support timezones (and can they be taken out the back and shot?) ? Is it that they "don't" (and can thus be fixed to include them) or "can't" by virtue of the language.
> 
> I struggle to believe that they can't.
> 
> Ambiguous times are a terrible terrible thing to put in at the design stage. At the very least, the spec should explicitly state that if the timezone is not present, it WILL be treated at UTC but I would prefer to exhaust all other options before making timezone optional (unless it is something that is by definition zoneless like UNIX time).

Maybe this is a phrasing issue. I was basically trying to be less constrained, since I interpreted the previous requirement as "you must set a locale". I'm trying to refer to this change: https://github.com/open-feature/java-sdk/issues/59

agentgonzo:
OK, it sounds like it's a phrasing issue and I'm overthinking it.
As long as the interpretation is "no timezone => UTC" then all is good.

specification/types.md
```diff
@@ -28,7 +28,7 @@ Structured data, presented however is idiomatic in the implementation language,
 
 ### Datetime
 
-A language primitive for representing a date and time, including timezone information.
+A language primitive for representing a date and time, optionally including timezone information.
```
agentgonzo, line 31:
```suggestion
A language primitive for representing a date and time, optionally including timezone information. If no timezone is specified, the date and time will be treated as UTC.
```
@toddbaert , WDYT?

agentgonzo (approved):

beeme1mr (approved):

james-milligan (approved):

PR #158, opened 2022-10-25
docs: typo corrections

DavidPHirsch:
Signed-off-by: David Hirsch <david.hirsch@dynatrace.com>

## This PR
small correction/suggestion


toddbaert:
I'm not so sure about this one. Is pluralization needed here?

beeme1mr:
I agree with Todd that pluralization doesn't seem necessary in this case.

PR #166, opened 2022-12-06
docs: include STATIC and CACHED in provider reasons

### LINKED ISSUE #165: Amend provider's reasons to include static & cached
[A basic caching strategy](https://github.com/open-feature/flagd/issues/216) of flagd revided the need of a `STATIC` reason to determine which values are cacheable.
The `CACHED` reason is needed to report to the client that the value has been retrieved from cache.

skyerus:
## This PR

Revives the `STATIC` reason in order to determine which values are cacheable.
Introduces the CACHED reason in order to report to the client that the value has been retrieved from cache.

### Related Issues

Fixes #165

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 __pycache__
+.idea
--- a/specification.json
+++ b/specification.json
@@ -230,7 +230,7 @@
         {
             "id": "Requirement 2.2.5",
             "machine_id": "requirement_2_2_5",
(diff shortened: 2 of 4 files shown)


agentgonzo (approved):

toddbaert (approved):
Though reasons are free form, I think these additions represent important information. We've seen the need for them come up in a few different implementations.

beeme1mr (approved):

Issue #167, opened 2022-12-08
First class client support in OpenFeature


toddbaert:
This is a high level issue where I'll try to collect various settled and ongoing discussions pertaining to improvements to OpenFeature for use in client applications.

### Required functionality

Below are some features that may need to be specified for effective client usage, many of which are also valuable on the server. I've checked discussions which are mostly settled, and left unchecked topics which I believe are still in contention. I will strike out (like ~this~) features that we don't believe should be implemented.

- [x] events
  - ready events, change events, etc, are useful for client contexts
  - providers may do evaluation "up front" and then signal readiness
  - change events might need to be scoped to prevent huge payloads when configurations change
  - see https://github.com/open-feature/ofep/pulls
  - also valuable on the server
- [x] "shutdown" functionality
  - providers may setup listeners, timers, that need to be closed
  - see https://github.com/open-feature/ofep/pull/30
  - also valuable on the server
- [x] synchronous API (evaluation functions that do not return promises/tasks/futures or take callbacks)
  - given the existence of readiness events, synchronous evaluation will likely make things easier for developers (avoids loaders, spinners, etc)
- [ ] bulk evaluation
  - bulk evaluation may be useful, especially for evaluating flags "up-front" before providers signal readiness
  - see https://github.com/open-feature/ofep/issues/13
- [ ] function(s) to reconcile context changes (a means to tell providers to "re-evaluate" their cached flag values)
  - this is analgous to methods in some vendor SDKs: [LaunchDarkly](https://launchdarkly.github.io/js-client-sdk/interfaces/LDClient.html#identify), [Flagsmisth](https://docs.flagsmith.com/clients/javascript#example-identifying-a-user-after-initialising-the-client)
  - such functions may be necessary if we implement synchronous evaluation

### Open questions

Below are some open questions that will inform, and be informed by the required functionality above.

- should the client implementations use the same APIs as the server-side implementations?
  - if not, should the client implementation somehow "wrap" or be implemented on top of the existing APIs?
  - if we want synchronous APIs, this may prove challenging
- in client usage, do we consider context to be "more static" (associated with a logged in user?) and modified independently of evaluation?
  - should `context` be removed from evaluation functions in this case?
- should we encourage vendors write a single provider for both the the client and server?
- we've been mostly focusing on web, but are there any considerations to take into account specifically for iOS and Android?
  - would we want both sync and async APIs for these platforms?

justinabrahms:
[...]

PR #171, opened 2023-01-10
feat!: add static/dynamic context

toddbaert:
This PR adds new points to support client use-cases. It does this primarily by defining and specifying a new _static-context_ paradigm.

There's a few requirements that now "fork" on the condition that the implementation targets the _dynamic context_ (server) or _static context_ (client) modes. There's no functional changes specified for dynamic context implementations.

For more justification on the above, see the issue [here](https://github.com/open-feature/spec/issues/167), and the OFEPs linked therein.

_**Please note**_ that there is some "number thrashing" with these changes. I tried to minimize it, but I think it's more important to have a coherent and sensible spec than to make sure to maintain the mapping between section numbers and content. I found that RFC2616 changed the content of their numbered sections [revision](https://datatracker.ietf.org/doc/html/draft-ietf-http-v11-spec-rev-00) to [revision](https://datatracker.ietf.org/doc/html/rfc2616) as well, so this isn't unprecedented.


specification/glossary.md
```diff
@@ -161,3 +164,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
...
 
 A rule is some criteria that's used to determine which variant a particular context should be mapped to.
+
+## SDK Paradigms
+
+Feature flag frameworks tend to come in two categories: those designed for use with a single user client application, and those designed for multi-user applications, such as web server applications. Some parts of the OpenFeature specification diverge depending on which paradigm the implementation seeks to adhere to.
+
+### Multi-Context Paradigm
+
+Server-side use cases typically perform flag evaluations on behalf of many users, with each request or event being associated with a particular user or client. For this reason, server frameworks typically operate something like this:
```
skyerus, line 174:
```suggestion
Server-side use cases typically perform flag evaluations on behalf of many users, with each request or event being associated with a particular user or client. For this reason, server frameworks typically operate similarly to this:
```

specification/glossary.md
```diff
@@ -161,3 +164,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
...
+- with each request or event, relevant dynamic context (for example, user session data) is provided to flag evaluations
+
+### Single-Context Paradigm
+
+In contrast with server-side or other service-type applications, client side applications typically operate in the context of a single user. Most feature flagging libraries for these applications have been designed with this in mind. Frequently, Client/web libraries operate something like this:
+
+- an initialization occurs, which fetches evaluated flags in bulk for a given context (user)
+- the evaluated flags are cached in the library
+- flag evaluations take place against this cache, without a need to provide context (context was already used to evaluate flags in bulk)
+- Functions are exposed on the libraries that signal the cache is no longer valid, and must be reconciled based on a context change. This frequently involves a network request or I/O operation.
```
skyerus, line 186: This bullet point uses sentences whereas the previous don't

specification/sections/01-flag-evaluation.md
```diff
@@ -20,9 +20,11 @@ The `evaluation API` allows for the evaluation of feature flag values, independe
 
 It's important that multiple instances of the `API` not be active, so that state stored therein, such as the registered `provider`, static global `evaluation context`, and globally configured `hooks` allow the `API` to behave predictably. This can be difficult in some runtimes or languages, but implementors should make their best effort to ensure that only a single instance of the `API` is used.
 
+### Setting a provider
+
 #### Requirement 1.1.2
 
-> The `API` **MUST** provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
+> The `API` **MUST** provide a `provider mutator`, a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
```
skyerus, line 27:
```suggestion
> The `API` **MUST** expose a `provider mutator`, a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
```

> provide a provider

makes me uneasy

specification/sections/02-providers.md
```diff
@@ -150,3 +150,70 @@ class MyProvider implements Provider {
...
+
+  //...
+}
+```
+
+### 2.5. Provider context reconciliation
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Client-focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
```
skyerus, line 181:
```suggestion
Single-context focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
```
Perhaps I've misunderstood but it seems to me that `single-context` encompasses client?

specification/sections/02-providers.md
```diff
@@ -150,3 +150,70 @@ class MyProvider implements Provider {
...
+
+#### Requirement 2.5.1
+
+> The provider interface **MUST** define an `on context set` handler, which takes a argument for the previous context, and the newly set context, and which can be optionally implemented to reconcile any stored state pertaining to the global evaluation context.
```
skyerus, line 185:
```suggestion
> The provider interface **MUST** define an `on context set` handler, which takes an argument for the previous context, and the newly set context, and which can be optionally implemented to reconcile any stored state pertaining to the global evaluation context.
```

specification/sections/02-providers.md
```diff
@@ -150,3 +150,70 @@ class MyProvider implements Provider {
...
+
+Especially in single-context implementations, providers and underlying SDKs my maintain a cache of evaluated flags for a particular context.
+The `on context set` handler provides a mechanism to update this state, often by re-evaluating flags in bulk with respect to the new context.
+
+```
+  // run the myOnReadyHandler function when the
```
skyerus, line 191: when the ...

specification/sections/03-evaluation-context.md
```diff
@@ -42,15 +42,39 @@ see: [structure](../types.md#structure), [datetime](../types.md#datetime)
...
+
+> The implementation uses the single-context paradigm.
+
+see: [single-context paradigm](../glossary.md#single-context-paradigm)
+
+##### Conditional Requirement 3.2.1.1
+
+> The API **MUST** have a method for supplying `evaluation context`.
+
+API (global) `evaluation context` can be used to supply data to flag evaluation, such as user name, email, or shopping cart items.
```
skyerus, line 57:
```suggestion
API (global) `evaluation context` can be used to supply data to flag evaluation, such as (but not limited to) user name, email, or shopping cart items.
```

specification/sections/05-events.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Overview
+
+`Events` allow consumers (_application integrator_, _application author_, _integration author_) to react to state changes in the provider or underlying flag management system, such as flag defintion changes, provider readiness, or error conditions. A provider may emit events or run a callback indicating that it received a certain event, optionally providing data associated with that event. Handlers registered on the client are then invoked with this data.
+
+### 5.1. Provider events
+
+#### Requirement 5.1.1
+
+> The provider **MUST** define a mechanism for signalling the occurance of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_SHUTDOWN`.
```
skyerus, line 19:
```suggestion
> The provider **MUST** define a mechanism for signalling the occurrence of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_SHUTDOWN`.
```

specification/glossary.md
toddbaert: I'll try to put both these ideas in a single sentence.

specification/sections/02-providers.md
toddbaert: Ya, I think I can change the language.

I'm also wondering if we should make this requirement conditional on single-context or not. Open to thoughts on that.

specification/sections/02-providers.md
toddbaert: This isn't even the correct example, I missed this one!

specification/sections/05-events.md
```diff
@@ -0,0 +1,40 @@
...
+
+> The provider **MUST** define a mechanism for signalling the occurrence of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_SHUTDOWN`.
+
+see: [provider event types](./../types.md#provider-events)
+
+### 5.2. Event handlers
+
+#### Requirement 5.2.1
+
+> The `client` **MUST** provide an `addHandler` function for attaching callbacks to `provider events`, which accepts event type(s) and a function to run when the associated event(s) occur.
```
beeme1mr, line 27: Is the term `callback` language agnostic?

specification/sections/01-flag-evaluation.md
```diff
@@ -20,9 +20,11 @@ The `evaluation API` allows for the evaluation of feature flag values, independe
 
 It's important that multiple instances of the `API` not be active, so that state stored therein, such as the registered `provider`, static global `evaluation context`, and globally configured `hooks` allow the `API` to behave predictably. This can be difficult in some runtimes or languages, but implementors should make their best effort to ensure that only a single instance of the `API` is used.
 
+### Setting a provider
+
 #### Requirement 1.1.2
 
-> The `API` **MUST** provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
+> The `API` **MUST** expose a `provider mutator`, a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
```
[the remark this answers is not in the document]
toddbaert: It's just adding a single searchable term for "the thing that sets the provider".

It's editorial.

specification/sections/01-flag-evaluation.md
```diff
@@ -33,6 +35,12 @@ See [provider](./02-providers.md) for details.
 
 #### Requirement 1.1.3
 
+> The `provider mutator` function **MUST** run the `initialize` function on the newly registered provider.
```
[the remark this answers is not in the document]
toddbaert: It would be an internal example. I've thus far generally avoided example implementations of internal SDK functions.

I did recently add more to the linked provider initialization example.

specification/sections/01-flag-evaluation.md
```diff
@@ -119,32 +161,63 @@ MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: '
...
+
+> The implementation uses the single-context paradigm.
+
+see: [single-context paradigm](../glossary.md#single-context-paradigm)
+
+##### Conditional Requirement 1.4.1.1
+
+> The `client` **MUST** provide methods for detailed flag value evaluation with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), and `evaluation options` (optional), which returns an `evaluation details` structure.
+
+```typescript
```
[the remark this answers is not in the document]
toddbaert: Not really, it's kinda nothing.

If you don't put _some_ language though, you dont get syntax highlighting, which looks... not good.

I'm considering converting most examples to Java because it's very readable, and pretty commonly understood.

specification/sections/01-flag-evaluation.md
```diff
@@ -33,6 +35,12 @@ See [provider](./02-providers.md) for details.
 
 #### Requirement 1.1.3
```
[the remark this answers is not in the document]
toddbaert: hmmmm... I kinda like that idea. So far, we've only used sub-sub-sub sections (x.x.x.x) in conditionals though, so we'd be breaking that pattern.

:thinking:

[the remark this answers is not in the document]
kinyoklion: It does seem very inconvenient for specification numbers to change. In both the testing front, for example unit tests that align with the specification, as well as the documentation/understanding front (referencing implementation back to specification).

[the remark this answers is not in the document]
beeme1mr: > So far, we've only used sub-sub-sub sections (x.x.x.x) in conditionals though, so we'd be breaking that pattern.

Changing the spec numbers of existing requirements seems far more destructive.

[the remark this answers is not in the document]
toddbaert: Ya, I think I agree, I will make this change.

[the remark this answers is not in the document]
toddbaert: I've done this.

specification/sections/01-flag-evaluation.md
```diff
@@ -163,66 +236,84 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
 
 See [hooks](./04-hooks.md) for details.
+
+### 1.6. Shutdown
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+#### Requirement 1.6.1
+
+> The API **MUST** define a `shutdown` or `dispose` function, which, when called, must call the respective `shutdown`/`dispose` method on the active provider.
```
weyert, line 309: Do we want to spec out the naming of this function instead of offering the flexibility of `shutdown` or `dispose`? I think we could just consistently name it `shutdown` instead confusing readers. 

This is how the `shutdown` for Opentelemetry is expressed:
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.15.0/specification/trace/sdk.md#shutdown

specification/sections/01-flag-evaluation.md
```diff
@@ -163,66 +236,84 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
+
+see: [`shutdown`/`dispose`](./02-providers.md#24-shutdown)
+
+#### Requirement 1.6.2
+
+> When the API's `provider mutator` is invoked, the `shutdown`/`dispose` method on the active provider **MUST** run, if defined.
+
+Setting a new provider means the previous provider is no longer in use, and should therefor be disposed of to prevent resource leaks.
```
weyert, line 317: Do we want to be more clear in which circumstances this would be the case? So there is a bit of a checklist implementators can check against? Instead of this vague sentence?

weyert (changes requested):
I have made a few comments.

specification/glossary.md
```diff
@@ -161,3 +164,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
 ### Rule
 
 A rule is some criteria that's used to determine which variant a particular context should be mapped to.
+
+## SDK Paradigms
```
moredip, line 168: FWIW, I really prefer referring to these as two *paradigms*, rather than "modes"/

specification/glossary.md
```diff
@@ -161,3 +164,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
...
+
+Feature flag frameworks tend to come in two categories: those designed for use with a single user client application, and those designed for multi-user applications, such as web server applications. Some parts of the OpenFeature specification diverge depending on which paradigm the implementation seeks to adhere to.
```
moredip, line 170: I wouldn't say that the frameworks come in two categories - many frameworks support both paradigms. I'd say "have SDKs which operate in two distinct paradigms" instead.

specification/glossary.md
```diff
@@ -161,3 +164,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
...
+
+### Multi-Context Paradigm
```
moredip, line 172: I'm not sure multi-context and single-context are the best names. I prefer dynamic-context and static-context instead. IMO that's the core distinction between the two paradigms - the context isn't fixed, and need to be provided each time.

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
 
 > In cases of abnormal execution, the `resolution details` structure's `error message` field **MAY** contain a string containing additional detail about the nature of the error.
+
+### 2.4 Initialization
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+#### Requirement 2.4.1
+
+> The `provider` interface **MUST** define an `initialize` function which accepts the global `evaluation context` as a argument, and which can be optionally implemented to perform initialization logic relevant to the provider.
```
moredip, line 161: Is the evaluation context required, or is it optional. Seems that we shouldn't REQUIRE that something be passed here, given that we're now adding more plumbing for updating the context and informing the provider that that's occured.

Could we make it optional and say that if it's not provided then the intial evaulation context defaults to a special-case "empty context", which we define as having no targeting key and no additional fields?

specification/sections/01-flag-evaluation.md
```diff
@@ -119,32 +161,63 @@ MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: '
...
 
 See [types](../types.md) for details.
 
-##### Requirement 1.3.3
+#### Requirement 1.3.4
 
 > The `client` **SHOULD** guarantee the returned value of any typed flag evaluation method is of the expected type. If the value returned by the underlying provider implementation does not match the expected type, it's to be considered abnormal execution, and the supplied `default value` should be returned.
 
-#### 1.4. Detailed Flag Evaluation
+### 1.4. Detailed Flag Evaluation
```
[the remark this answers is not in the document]
dabeeeenster: I think generally that with SDK's that target client UI's, developers would expect the flag evaluation methods to be async.

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
+
+### 2.5. Provider context reconciliation
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Single-context focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
+
+#### Requirement 2.5.1
+
+> The provider interface **MUST** define an `on context set` handler, which takes an argument for the previous context, and the newly set context, and which can be optionally implemented to reconcile any stored state pertaining to the global evaluation context.
```
moredip, line 192: just personal taste I suppose, but I prefer `on context changed` to `on context set`. Set sounds like something that only happens once, changed seems more like something that can occur multiple times.

moredip: Also, I really like that we're passing the old context in addition to the new context 👍

specification/sections/03-evaluation-context.md
```diff
@@ -42,15 +42,39 @@ see: [structure](../types.md#structure), [datetime](../types.md#datetime)
...
+
+> The API **MUST** have a method for supplying `evaluation context`.
+
+API (global) `evaluation context` can be used to supply data to flag evaluation, such as (but not limited to) user name, email, or shopping cart items.
+
+##### Conditional Requirement 3.2.1.2
+
+> The Client and invocation **MUST NOT** have a method for supplying `evaluation context`.
+
+In the single-context paradigm, context is global. The client and invocation cannot supply evaluation context.
```
moredip, line 63: I'm struggling with this quite a bit. 

On the one hand, I'm concerned that making evaluation context global is going to cause serious issue for sophisticated consumers who target flag evaluation across multiple dimensions. For example, the Spotify use case where some flags vary by artists, and the "current artist" is going to change frequently. If we only have 1 single shared context then every time ANY evaluation context changes then ALL feature flags would need to be re-evaluated. If we had some sort of nesting/merging capability then we could probably avoid this "cache busting" problem.

However, on the other hand, I suspect that some provider's client-side SDKs assume that evaluation context is shared global state, and if we allowed nested state that could make implementing the OF provider API harder.

specification/sections/01-flag-evaluation.md
toddbaert: Maybe - the only reason I hesitated, was especially in the case of the client, we might want to use some standard language interfaces, such as [AutoCloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) in Java or [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-7.0) in .NET. This would give SDK authors the flexibility to implement those if wanted.

These allow things to be "autoclosed" by certain language constructs. I'm interested in @agentgonzo @justinabrahms @kinyoklion @benjiro 's opinoins on this one.

specification/sections/01-flag-evaluation.md
toddbaert: I will add a some more detail, and an link to disposal.

specification/glossary.md
toddbaert: I think I agree. I will work on this wording.

specification/sections/01-flag-evaluation.md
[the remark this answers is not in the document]
toddbaert: I'm not sure how prescriptive we should be here. I think I've outlined how evaluation can be made sync in the "static context" paradigm, but I'm hesitant to actually _specify_ that.

specification/sections/02-providers.md
toddbaert: I think the problem here is that the sentence is ambiguous.  When I said `and which can be optionally implemented` I was saying the `initialize` function can be optionally implemented by provider authors.

specification/sections/03-evaluation-context.md
toddbaert: > On the one hand, I'm concerned that making evaluation context global is going to cause serious issue for sophisticated consumers who target flag evaluation across multiple dimensions. For example, the Spotify use case where some flags vary by artists, and the "current artist" is going to change frequently. If we only have 1 single shared context then every time ANY evaluation context changes then ALL feature flags would need to be re-evaluated. If we had some sort of nesting/merging capability then we could probably avoid this "cache busting" problem.

Especially with the ability to dif the old/new context, could this be something the provider handles? For example, my context could be a complex object with multiple top level objects, and I could check them for changes in `onContextSet` and only re-evaluate what's needed, assuming my provider is intelligent enough for that.

specification/sections/02-providers.md
toddbaert: I've tried to removed the ambiguity. Please check again this makes sense.

specification/glossary.md
toddbaert: Updated.

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
 
 > In cases of abnormal execution, the `resolution details` structure's `error message` field **MAY** contain a string containing additional detail about the nature of the error.
+
+### 2.4 Initialization
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+#### Requirement 2.4.1
+
+> The `provider` interface **MUST** define an optional `initialize` function, which accepts the global `evaluation context` as an argument and performs initialization logic relevant to the provider.
```
agentgonzo, line 161: What's the benefit of forcing this initialize function, rather than doing the initialization implicitly in the constructor?

specification/sections/02-providers.md
agentgonzo: +1 to "on context changed". That is more in line with existing naming conventions across the industry.

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
+
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
+
+### 2.6. Shutdown
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+#### Requirement 2.6.1
+
+> The provider interface **MUST** define a `shutdown` or `dispose` function, which can be optionally implemented to perform whatever cleanup or shutdown is necessary for the provider implementation.
```
agentgonzo, line 218: Why **must**? Why not make it optional? If the provider doesn't need to do anything on shutdown, why not have a default (no-op) method in the interface so the provider doesn't have to do anything?

It's not hugely important, as implementing a no-op method is trivial anyway. I'm just curious as to why it's mandatory.

specification/sections/03-evaluation-context.md
```diff
@@ -42,15 +42,39 @@ see: [structure](../types.md#structure), [datetime](../types.md#datetime)
...
+
+> The API **MUST** have a method for supplying `evaluation context`.
+
+API (global) `evaluation context` can be used to supply data to flag evaluation, such as (but not limited to) user name, email, or shopping cart items.
+
+##### Conditional Requirement 3.2.1.2
+
+> The Client and invocation **MUST NOT** have a method for supplying `evaluation context`.
+
+In the static-context paradigm, context is global. The client and invocation cannot supply evaluation context.
```
agentgonzo, line 63: Interesting. What's the benefit of enforcing that the invocation cannot supply a context? I'm struggling to see the use-case where this is beneficial to the user.

I can understand that you could get the state where two contexts contain different values for the same key in the context, but in my mind it's trivial to say that the invocation-context overrides the global-context for situations where we have a key-clash.

Given that CloudBees Feature Management allows setting a global context and invocation context, it basically means that CBFM will always operate in the evaluation context paradigm. I'm not sure if I'm understanding this correctly though, or if this is an issue.

specification/sections/02-providers.md
toddbaert: The constructor can't have the evaluation context. At initialization, the SDK can provide a context to the provider if one has been set. It makes things a bit more consistent with respect to the `onContextSet` handler: the provider author now has a few well-defined points at which to perform bulk evaluation/caching if that's relevant to them.

specification/sections/03-evaluation-context.md
toddbaert: As you note, some providers (Cloudbees being one of them) do local evaluation in client apps, and therefore can do "dynamic" evaluation (involving an invocation context) without issue. This isn't the case with other vendors. Those SDKs basically have a "bulk evaluate + cache + return-from-cache" paradigm, where a change in context triggers another bulk evaluation (rules are never sent to the client).

Allowing dynamic context in this case is basically a no-go, it's a huge performance footgun. While it's possible for Cloudbees to fit into the static-context paradigm, it's not possible for the bulk-evaluated SDKs to fit into a dynamic context paradigm. We're trying to walk a fine-line here to find an abstraction that works across as many paradigms as possible.

@moredip  did a really good job explaining this in his recent blog post: https://docs.openfeature.dev/blog/catering-to-the-client-side

I'm definitely interested in getting more thoughts from you on this, since Cloudbees is one of the two vendors I'm aware of whose client SDKs works by shipping a ruleset.

toddbaert: This [OFEP](https://github.com/open-feature/ofep/pull/41) is also highly relevant.

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
+
+### 2.5. Provider context reconciliation
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Static-context focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
+
+#### Requirement 2.5.1
+
+> The provider interface **MUST** define an `on context set` handler, which takes an argument for the previous context, and the newly set context, and which can be optionally implemented to reconcile any stored state pertaining to the global evaluation context.
```
justinabrahms, line 192: There's a lot of overly specifying about what the interface must be. This kinda feels wrong to me. Not all languages have interfaces. I would prefer that we talk about what providers may do, and what the sdk must do if they do that.

Instead of saying "the provider interface must define a handler".. I'd rather this read: "Providers may register a context change handler. Separately, if the context changes, the sdk must call that handler".

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
+
+Especially in static-context implementations, providers and underlying SDKs my maintain a cache of evaluated flags for a particular context.
```
justinabrahms, line 194:
```suggestion
Especially in static-context implementations, providers and underlying SDKs may maintain a cache of evaluated flags for a particular context.
```

specification/types.md
```diff
@@ -53,16 +53,17 @@ A structure which contains a subset of the fields defined in the `evaluation det
...
+| DEFAULT         | The resolved value fell back to a pre-configured value (no dynamic evaluation occurred or dynamic evaluation yielded no result). |
+| TARGETING_MATCH | The resolved value was the result of a dynamic evaluation, such as a rule or specific user-targeting.                            |
+| SPLIT           | The resolved value was the result of pseudorandom assignment.                                                                    |
+| CACHED          | The resolved value was retrieved from cache.                                                                                     |
+| DISABLED        | The resolved value was the result of the flag being disabled in the management system.                                           |
+| UNKNOWN         | The reason for the resolved value could not be determined.                                                                       |
+| STALE           | The resolved value is non-authoritative or possible out of date                                                                  |
```
justinabrahms, line 65:
```suggestion
| STALE           | The resolved value is non-authoritative or possibly out of date                                                                  |
```

tools/specification_parser/test_specification.md
```diff
@@ -4,27 +4,27 @@
 
 Some content.
 
-##### Requirement Some requirement
+#### Requirement Some requirement
```
justinabrahms, line 7: I would have expected to see some sort of anchor link in this to show off the problem you mentioned above.

specification/types.md
```diff
@@ -85,3 +86,14 @@ An enumerated error code represented idiomatically in the implementation languag
...
 - hooks (one or more [hooks](./sections/04-hooks.md), optional)
+
+### Provider Events
+
+An enumeration of provider events.
+
+| Event                          | Explanation                                                                                 |
+| ------------------------------ | ------------------------------------------------------------------------------------------- |
+| PROVIDER_READY                 | The provider is ready to perform flag evaluations.                                          |
+| PROVIDER_ERROR                 | The provider signalled an error.                                                            |
```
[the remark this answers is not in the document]
toddbaert: Signalling the provider is ready (PROVIDER_READY) with an event is something many vendors SDKs do: 

https://docs.launchdarkly.com/sdk/client-side/javascript#initializing-the-client
https://help.split.io/hc/en-us/articles/360020448791-JavaScript-SDK#basic-use

Especially in the browser, it's helpful to expose this information in an event because it gives application authors a means to know when they can start evaluating flags. Before this point, you might need to show loaders, spinners, etc.

For PROVIDER_CONFIGURATION_CHANGED, that's a way to _subscribe_ to configuration or flag values changes from the source of truth. This allows use to attach handlers which react to changes in flag values without the user performing any action (in the playground demo, the color of the app changes, for instance, without any refresh or polling). Many SDKs also support things like this:

https://docs.launchdarkly.com/sdk/features/flag-changes
https://docs.flagsmith.com/clients/javascript#api-reference (see `onChange` callback)

It's not really about what information these events provide over the evaluation details, as much as enabling different programming paradigms.

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
+
+  //...
+}
+```
+
+### 2.5. Provider context reconciliation
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Static-context focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
```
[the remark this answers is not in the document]
toddbaert: I went back and forth on this myself. I can go either way. I'll leave this comment unresolved so people can agree/disagree and I'll close it when we have a consensus.

specification/sections/01-flag-evaluation.md
```diff
@@ -163,66 +236,84 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
+### Evaluation Options
 
-##### Requirement 1.5.1
+#### Requirement 1.5.1
 
 > The `evaluation options` structure's `hooks` field denotes an ordered collection of hooks that the client **MUST** execute for the respective flag evaluation, in addition to those already configured.
 
 See [hooks](./04-hooks.md) for details.
+
+### 1.6. Shutdown
```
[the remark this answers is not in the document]
toddbaert: I can break these all up if you'd like. I considered shutdown+events+static-context all required for proper client support, which is why I opted to create this PR as one thing.

tools/specification_parser/test_specification.md
toddbaert: Here is an example of what was breaking before the parser fix: https://github.com/open-feature/spec/pull/171/files#diff-300118ed122b34e94bd16e8714f09eb91547e48b20e087d4b5ad2847698581d3R112

specification/sections/05-events.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Overview
+
+`Events` allow consumers (_application integrator_, _application author_, _integration author_) to react to state changes in the provider or underlying flag management system, such as flag defintion changes, provider readiness, or error conditions. A provider may emit events or run a callback indicating that it received a certain event, optionally providing data associated with that event. Handlers registered on the client are then invoked with this data.
+
+### 5.1. Provider events
+
+#### Requirement 5.1.1
+
+> The provider **MUST** define a mechanism for signalling the occurrence of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_SHUTDOWN`.
```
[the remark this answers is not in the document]
toddbaert: I guess the language needs to be made more clear here. This is a requirement on the provider interface(s) - the SDK - not on the actual implementation. The SDK has to allow the provider-author the ability to implement events, but the provider author need not actually implement them if it's not relevant to them.

In the experimental web-sdk, I've made the events completely optional.

toddbaert, line 19:
```suggestion
> The provider **MAY** define a mechanism for signalling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_SHUTDOWN`.
```

specification/sections/02-providers.md
toddbaert: Thinking about it more, there's a few places where I've used **MUST** to describe optional fields/methods. I think generally, that you're right, and this can be converted to **MAY**, mentioning the specific constrcut. For example

`The provider interface MUST have optional field X which does Y` 

becomes:

`The provider MAY have X which does Y`.

I think this is generally better and and I've applied it in a few places.

specification/sections/02-providers.md
toddbaert, line 161:
```suggestion
> The `provider` **MAY** define an optional `initialize` function, which accepts the global `evaluation context` as an argument and performs initialization logic relevant to the provider.
```

specification/sections/02-providers.md
```diff
@@ -150,3 +151,85 @@ class MyProvider implements Provider {
...
+
+#### Requirement 2.5.1
+
+> The provider interface **MUST** define an `on context set` handler, which takes an argument for the previous context, and the newly set context, and which can be optionally implemented to reconcile any stored state pertaining to the global evaluation context.
```
toddbaert, line 192:
```suggestion
> The provider **MAY** define an `on context set` handler, which takes an argument for the previous context, and the newly set context, and reconciles any stored state pertaining to the global evaluation context.
```

specification/types.md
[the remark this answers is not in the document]
toddbaert: > Ready and change make sense. I don't understand how someone would use error.

The use case I had in mind was particularly relevant for some of the SDKs/providers that maintain real-time connections. For example, flagd maintains a gRPC connection that could fail, and [Flagsmith has an "onError" hanlder](https://docs.flagsmith.com/clients/javascript#api-reference), and split has an event called `client.Event.SDK_READY_TIMED_OUT`. I was thinking it could be useful to have an error event which these sorts of thigns could be mapped to.

specification/sections/05-events.md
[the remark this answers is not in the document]
toddbaert: I've updated this language.

specification/sections/02-providers.md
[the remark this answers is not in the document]
beeme1mr: I prefer "context change" in this case.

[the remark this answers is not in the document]
jonathannorris: Yea I agree with "context change" here, as the initial context which the provider will fetch its results for is set during `initialize()`.

specification/sections/03-evaluation-context.md
```diff
@@ -42,15 +42,39 @@ see: [structure](../types.md#structure), [datetime](../types.md#datetime)
...
 
-### 3.2 Merging Context
+### 3.2 Context levels and merging
 
-#### Requirement 3.2.1
+#### Condition 3.2.1
+
+> The implementation uses the static-context paradigm.
+
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
```
jonathannorris, line 51: Not sure where the correct part in the spec is to ask about this, but what about providing a callback mechanism at the Variable level for variable value updates when the context is changed?

From this [blog post](https://openfeature.dev/blog/catering-to-the-client-side/) it seems like the intention was to use events like this: 
```
client.addHandler(ProviderEvents.FlagValuesChanged, () => {
  // this would trigger a re-render
  setUseFormalSalutation(client.getBooleanValue('use-formal-salutation', false));
});
```

I would recommend that a flag / variable level callback handler would be more useful for users and make integrations into React Hooks and other similar implementations easier.

```typescript
openFeatureClient.getBooleanValue('boolean-flag', false)
    .onUpdate((value: boolean, details: EvaluationDetails<boolean>) => {

    })
```

specification/glossary.md
toddbaert: Addressed - dynamic/static

specification/sections/03-evaluation-context.md
toddbaert: added

specification/glossary.md
```diff
@@ -163,3 +166,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
 ### Rule
 
 A rule is some criteria that's used to determine which variant a particular context should be mapped to.
+
+## SDK Paradigms
+
+Feature flag frameworks have SDKs which operate in two distinct paradigms: those designed for use with a single user client application (e.g. mobile phones), and those designed for multi-user applications, such as web server applications. Some parts of the OpenFeature specification diverge depending on the paradigm.
```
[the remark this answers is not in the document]
jonathannorris: Yea without this my brain connected `web server applications` to `web applications`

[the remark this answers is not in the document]
jonathannorris: There is certainly a language challenge here where using `user` is more likely to be easily understood initially by developers new to Feature Flags, where `context` is the more technically correct term. So it really depends on the target audience we are trying to speak to with this document.

specification/glossary.md
```diff
@@ -163,3 +166,25 @@ Pseudorandomly resolve flag values using a context property, such as a targeting
...
+
+### Dynamic-Context Paradigm
+
+Server-side use cases typically perform flag evaluations on behalf of many users, with each request or event being associated with a particular user or client. For this reason, server frameworks typically operate similarly to this:
+
+- the application is initialized with some static context (geography, service name, hostname, etc)
+- with each request or event, relevant dynamic context (for example, user session data) is provided to flag evaluations
```
jonathannorris, line 179:
```suggestion
- with each request or event, relevant dynamic context (for example, user session data, unique user identifiers) is provided to flag evaluations
```

specification/sections/02-providers.md
```diff
@@ -244,6 +244,30 @@ class MyProvider implements Provider, AutoDisposable {
   void dispose() {
     // close connections, terminate threads or timers, etc...
   }
+```
+
+### 2.6. Provider context reconciliation
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Static-context focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
```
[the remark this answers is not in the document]
jonathannorris: +1 for `on context changed`

specification/sections/02-providers.md
```diff
@@ -244,6 +244,30 @@ class MyProvider implements Provider, AutoDisposable {
...
+
+#### Requirement 2.6.1
+
+> The provider **MAY** define an `on context changed` handler, which takes an argument for the previous context and the newly set context, and reconciles any stored state pertaining to the global evaluation context.
+
+Especially in static-context implementations, providers and underlying SDKs may maintain a cache of evaluated flags for a particular context.
```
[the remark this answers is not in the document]
jonathannorris: why `local state` over `local cache`, is describing it as a `cache` too specific? I think describing it as a `cache` is likely easier to understand.

specification/sections/03-evaluation-context.md
jonathannorris: I'm not certain if this is part of the implementation (going to test it out soon), but does the OpenFeature SDK have a mechanism of being told when the underlying provider has updated its feature cache? if so then this could be handled within the OpenFeature SDKs. But you are likely correct that this is likely too specific of an implementation detail for this spec right now. However, as we get into implementing more reactive / hook-based client-side SDKs this almost becomes necessary to propagate flag updates.

specification/sections/02-providers.md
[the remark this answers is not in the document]
toddbaert: Ya - it was "Set" and I updated it to "changed" after some feedback, but I missed this spot. I'll do a few more searches to make sure there's no other occurrences.

specification/sections/02-providers.md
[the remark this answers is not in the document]
toddbaert: I think I agree with @jonathannorris on this one. Because this is more of an example and not normative text, I think the use of "cache" is more concrete and helpful in communicating the idea.

specification/sections/03-evaluation-context.md
```diff
@@ -44,13 +44,39 @@ The key uniquely identifies a field in the `evaluation context` and it should be
...
+
+##### Conditional Requirement 3.2.2.1
+
+> The API **MUST** have a method for supplying `evaluation context`.
+
+API (global) `evaluation context` can be used to supply data to flag evaluation, such as (but not limited to) user name, email, or shopping cart items.
```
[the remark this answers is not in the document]
toddbaert: I agree. I like "organization membership".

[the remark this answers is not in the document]
toddbaert: updated

specification/sections/02-providers.md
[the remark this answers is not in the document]
toddbaert: fixed.

specification/glossary.md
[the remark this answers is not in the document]
Kavindu-Dodan: @jonathannorris yeah agree with the point. However, I think the spec should be generic enough with the terminology to cover a wide verity of usecases and then give examples to connect with developer scenarios. And this comment is connected with this poll [1]

[1] - https://github.com/open-feature/spec/discussions/191

specification/sections/02-providers.md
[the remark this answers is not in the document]
kinyoklion: I think the "pertaining to flag evaluations" versus "evaluated flags" part may be more important. An SDK may have local state which it uses to perform evaluations, or it may have local state which is fully evaluated, maybe the state that changes isn't directly flags at all, but segmenting information, or metadata used to drive analytics.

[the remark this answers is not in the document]
kinyoklion: "maintain state for a particular context"

Though, it is probably also fine as is, because it isn't normative.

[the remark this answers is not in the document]
toddbaert: :thinking:

specification/glossary.md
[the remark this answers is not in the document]
toddbaert: > There is certainly a language challenge here where using user is more likely to be easily understood initially by developers new to Feature Flags, where context is the more technically correct term that can help describe more diverse use cases. So it really depends on the target audience we are trying to speak to with this document.

My philosophy on this is normative text should prioritize technical correctness, while the non-normative sections lean more toward practical examples and "typical" patterns.

specification/sections/01-flag-evaluation.md
```diff
@@ -190,69 +230,96 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+> The implementation uses the static-context paradigm.
+
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
+
+##### Conditional Requirement 1.4.2.1
+
+> The `client` **MUST** provide methods for detailed flag value evaluation with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), and `evaluation options` (optional), which returns an `evaluation details` structure.
```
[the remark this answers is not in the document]
toddbaert: @moredip do you have more concerns on this one? The web-sdk currently implements this as @beeme1mr describes.

specification/sections/02-providers.md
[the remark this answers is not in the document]
toddbaert: I've gone with `maintain state for a particular context` as @kinyoklion  recommends, which (I think) satisfies @moredip 's original concern and reads a bit better.

specification/sections/03-evaluation-context.md
toddbaert: @jonathannorris have you had a chance to confirm if `PROVIDER_CONFIGURATION_CHANGED` covers the case you mentioned?

kinyoklion (approved):

specification/sections/01-flag-evaluation.md
```diff
@@ -152,10 +158,38 @@ See [evaluation context](./03-evaluation-context.md) for details.
...
+string myString = client.getStringValue('string-flag', 'N/A', options);
+
+// example number flag evaluation
+number myNumber = client.getNumberValue('number-flag', 75);
+
+// example overloaded structure flag evaluation with optional params
+MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: 'N/A', percentage: 75 }, options);
+```
+
+See [evaluation context](./03-evaluation-context.md) for details.
```
Kavindu-Dodan, line 185:
```suggestion
```

This is no longer appliacable

specification/sections/02-providers.md
```diff
@@ -245,6 +245,30 @@ class MyProvider implements Provider, AutoDisposable {
...
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Static-context focused providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on context changed` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context.
+
+#### Requirement 2.6.1
+
+> The provider **MAY** define an `on context changed` handler, which takes an argument for the previous context and the newly set context, in order to respond to an evaluation context change.
+
+Especially in static-context implementations, providers and underlying SDKs may maintain state for a particular context.
+The `on context changed` handler provides a mechanism to update this state, often by re-evaluating flags in bulk with respect to the new context.
```
Kavindu-Dodan, line 261: How will this look in an implementation 🤔  I assume this is called by SDK when global `EvaluationContext` is changed ?

specification/sections/03-evaluation-context.md
```diff
@@ -44,13 +44,39 @@ The key uniquely identifies a field in the `evaluation context` and it should be
...
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+> The implementation uses the static-context paradigm.
+
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
+
+##### Conditional Requirement 3.2.2.1
+
+> The API **MUST** have a method for supplying `evaluation context`.
```
Kavindu-Dodan, line 69:
```suggestion
> The API **MUST** have a method for setting the global `evaluation context`.
```

I think this will be a setter in API level 🤔

Kavindu-Dodan (approved):

beeme1mr (approved):

thomaspoignant (approved):

specification/sections/01-flag-evaluation.md
toddbaert: This was added in a different spec change and it's no longer relevant to this PR.

weyert (approved):

specification/sections/03-evaluation-context.md
```diff
@@ -44,13 +44,39 @@ The key uniquely identifies a field in the `evaluation context` and it should be
...
+> The API **MUST** have a method for setting the global `evaluation context`.
+
+API (global) `evaluation context` can be used to supply data to flag evaluation, such as (but not limited to) user name, email, or user organization membership changes.
+
+##### Conditional Requirement 3.2.2.2
+
+> The Client and invocation **MUST NOT** have a method for supplying `evaluation context`.
+
+In the static-context paradigm, context is global. The client and invocation cannot supply evaluation context. 
+
```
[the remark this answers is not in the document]
lukas-reining: Initially I was just wondering why, and after looking closer I found this conflict.
As I said, not allowing it seems like a tradeoff and I was just wondering if this is the reason.
So yes we can resolve this one 🙂

One thing that came to my mind: do we have any place where we document such explicit decisions. I could imagine questions like this or ones where the reason for the decision is not even so obvious are documented. To me the spec describes the "what" but it does not always answer the "why". 
Maybe we would benefit of a documentation format in similar to [ADRs](https://www.cognitect.com/blog/2011/11/15/documenting-architecture-decisions).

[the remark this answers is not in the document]
lukas-reining: Mhh, my feeling is that they are kind of different.
What I mean are not proposals, but rather decisions that have been made in the context of some constrains, to make it easy to reference it and say "this is why we decided".
OFEPs to me are really valuable, but I feel like these decision records that I mean are just used with a bit different motivation 🤔 But I see the large overlap, also when looking at the current OFEPs

lukas-reining (approved):

specification/sections/03-evaluation-context.md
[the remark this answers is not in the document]
lukas-reining: I mean this is the wrong thread for it.
Maybe I can think a bit more about this and I can bring it back up somewhere else so we can resolve this thread.

fabriziodemaria (approved):

PR #172, opened 2023-01-11
fix: links and link checks

toddbaert:
We we're missing some checks for `markdown-link-check`.

Unfortunately, there's some bugs in it that don't check anchors in other docs, so until that's fixed we might still occasionally get broken links. I've manually fixed all link issues with this PR though (I hope).

There doesn't appear to be many alternatives, so we may want to consider fixing [this bug](https://github.com/tcort/markdown-link-check/issues/212).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ parse: _check_python
 lint: node_modules
 	@python ./tools/specification_parser/lint_json_output.py specification.json
 	./node_modules/.bin/markdownlint --ignore node_modules/ --ignore tools/ **/*.md
-	./node_modules/.bin/markdown-link-check -c .markdown-link-check-config.json README.md specification/*.md
+	./node_modules/.bin/markdown-link-check -c .markdown-link-check-config.json README.md specification/*.md specification/**/*.md
 
 fix: node_modules
 	prettier -w **/*.md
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ parse: _check_python
 lint: node_modules
 	@python ./tools/specification_parser/lint_json_output.py specification.json
 	./node_modules/.bin/markdownlint --ignore node_modules/ --ignore tools/ **/*.md
-	./node_modules/.bin/markdown-link-check -c .markdown-link-check-config.json README.md specification/*.md
+	./node_modules/.bin/markdown-link-check -c .markdown-link-check-config.json README.md specification/*.md specification/**/*.md
```
toddbaert, line 10: Also recursively check spec subfolders.

aepfli: sadly this won't prevent this issue currently - but hopefully in the near future see https://github.com/tcort/markdown-link-check/issues/225

toddbaert: Ya, no problem. This will at least validate the non-anchor parts of links in docs in subfolders, so it's a step in the right direction.

toddbaert:
Force merging this since it's just fixes to links, and no content changes.

PR #179, opened 2023-04-04
feat: initialization and shutdown

toddbaert:
Adds initialization and shutdown discussed in DRAFT client PR, and OFEPs. This is a piece of the (too big) draft PR [here](https://github.com/open-feature/spec/pull/171). The changes are fairly small and non-breaking.

There's a few non-functional changes here (fixes to formatting and minor enhancements to the parser).

Be sure to review `specification.json` for the net changes.

Most of this is implemented in the [web-sdk](https://github.com/open-feature/js-sdk/tree/main/packages/client/src) already. Some conforming (demo) providers can be found in the playground:

- [cloudbees](https://github.com/open-feature/playground/blob/main/packages/web-cloudbees-provider/src/lib/web-cloudbees-provider.ts)
- [flagsmith](https://github.com/open-feature/playground/blob/main/packages/web-flagsmith-provider/src/lib/web-flagsmith-provider.ts)
- [harness](https://github.com/open-feature/playground/blob/main/packages/web-harness-provider/src/lib/web-harness-provider.ts)
- [launchdarkly](https://github.com/open-feature/playground/blob/main/packages/web-launchdarkly-provider/src/lib/web-launchdarkly-provider.ts)
- [split](https://github.com/open-feature/playground/blob/main/packages/web-split-provider/src/lib/web-split-provider.ts)

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -358,20 +358,27 @@
         {
             "id": "Requirement 2.4.2",
             "machine_id": "requirement_2_4_2",
-            "content": "The `provider` MAY define a mutable boolean `ready` field which indicates the readiness of the provider.",
+            "content": "The `provider` MAY define a `status` field/accessor which indicates the readiness of the provider, with possible values `NOT_READY`, `READY`, or `ERROR`.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
         {
             "id": "Requirement 2.4.3",
             "machine_id": "requirement_2_4_3",
-            "content": "The provider MUST set its `ready` field if its `initialize` function terminates normally.",
+            "content": "The provider MUST set its `status` field/accessor to `READY` if its `initialize` function terminates normally.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 2.4.4",
             "machine_id": "requirement_2_4_4",
+            "content": "The provider MUST set its `status` field to `ERROR` if its `initialize` function terminates abnormally.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.4.5",
+            "machine_id": "requirement_2_4_5",
             "content": "The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.",
             "RFC 2119 keyword": "SHOULD",
             "children": []
--- a/specification.json
+++ b/specification.json
@@ -8,9 +8,23 @@
             "children": []
         },
         {
-            "id": "Requirement 1.1.2",
-            "machine_id": "requirement_1_1_2",
-            "content": "The `API` MUST provide a function to set the default `provider`, which accepts an API-conformant `provider` implementation.",
+            "id": "Requirement 1.1.2.1",
+            "machine_id": "requirement_1_1_2_1",
+            "content": "The `API` MUST define a `provider mutator`, a function to set the default `provider`, which accepts an API-conformant `provider` implementation.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 1.1.2.2",
+            "machine_id": "requirement_1_1_2_2",
+            "content": "The `provider mutator` function MUST invoke the `initialize` function on the newly registered provider before using it to resolve flag values.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 1.1.2.3",
+            "machine_id": "requirement_1_1_2_3",
+            "content": "The `provider mutator` function MUST invoke the `shutdown` function on the previously registered provider once it's no longer being used to resolve flag values.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -214,6 +228,13 @@
             "children": []
         },
         {
+            "id": "Requirement 1.6.1",
+            "machine_id": "requirement_1_6_1",
+            "content": "The API MUST define a `shutdown` function, which, when called, must call the respective `shutdown` function on the active provider.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
             "id": "Requirement 2.1.1",
             "machine_id": "requirement_2_1_1",
             "content": "The provider interface MUST define a `metadata` member or accessor, containing a `name` field or accessor of type string, which identifies the provider implementation.",
@@ -289,22 +310,22 @@
                     "content": "The `resolution details` structure SHOULD accept a generic argument (or use an equivalent language feature) which indicates the type of the wrapped `value` field.",
                     "RFC 2119 keyword": "SHOULD",
                     "children": []
+                },
+                {
+                    "id": "Requirement 2.2.9",
+                    "machine_id": "requirement_2_2_9",
+                    "content": "The `provider` SHOULD populate the `resolution details` structure's `flag metadata` field.",
+                    "RFC 2119 keyword": "SHOULD",
+                    "children": []
+                },
+                {
+                    "id": "Requirement 2.2.10",
+                    "machine_id": "requirement_2_2_10",
+                    "content": "`flag metadata` MUST be a structure supporting the definition of arbitrary properties, with keys of type `string`, and values of type `boolean | string | number`.",
+                    "RFC 2119 keyword": "MUST",
+                    "children": []
                 }
             ]
-        },
-        {
-            "id": "Requirement 2.2.9",
-            "machine_id": "requirement_2_2_9",
-            "content": "The `provider` SHOULD populate the `resolution details` structure's `flag metadata` field.",
-            "RFC 2119 keyword": "SHOULD",
-            "children": []
-        },
-        {
-            "id": "Requirement 2.2.10",
-            "machine_id": "requirement_2_2_10",
-            "content": "`flag metadata` MUST be a structure supporting the definition of arbitrary properties, with keys of type `string`, and values of type `boolean | string | number`.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
         },
         {
             "id": "Requirement 2.3.1",
@@ -324,6 +345,20 @@
             "id": "Requirement 2.3.3",
             "machine_id": "requirement_2_3_3",
             "content": "In cases of abnormal execution, the `resolution details` structure's `error message` field MAY contain a string containing additional detail about the nature of the error.",
+            "RFC 2119 keyword": "MAY",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.4.1",
+            "machine_id": "requirement_2_4_1",
+            "content": "The `provider` MAY define an `initialize` function which accepts the global `evaluation context` as an argument and performs initialization logic relevant to the provider.",
+            "RFC 2119 keyword": "MAY",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.5.1",
+            "machine_id": "requirement_2_5_1",
+            "content": "The provider MAY define a `shutdown` function to perform whatever cleanup is necessary for the implementation.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -358,15 +358,22 @@
         {
             "id": "Requirement 2.4.2",
             "machine_id": "requirement_2_4_2",
-            "content": "The `provider` MAY define a mutable boolean `ready` member which indicates the readiness of the provider.",
+            "content": "The `provider` MAY define a mutable boolean `ready` field which indicates the readiness of the provider.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
         {
             "id": "Requirement 2.4.3",
             "machine_id": "requirement_2_4_3",
-            "content": "If the `initialize` function terminates normally, the provider's `ready` member MUST be set to `true`.",
+            "content": "The provider MUST set its `ready` field if its `initialize` function terminates normally.",
             "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.4.4",
+            "machine_id": "requirement_2_4_4",
+            "content": "The provider SHOULD indicate an error if flag resolution is attempted before the provider is ready.",
+            "RFC 2119 keyword": "SHOULD",
             "children": []
         },
         {
--- a/specification.json
+++ b/specification.json
@@ -356,6 +356,20 @@
             "children": []
         },
         {
+            "id": "Requirement 2.4.2",
+            "machine_id": "requirement_2_4_2",
+            "content": "The `provider` MAY define a mutable boolean `ready` member which indicates the readiness of the provider.",
+            "RFC 2119 keyword": "MAY",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.4.3",
+            "machine_id": "requirement_2_4_3",
+            "content": "If the `initialize` function terminates normally, the provider's `ready` member MUST be set to `true`.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
             "id": "Requirement 2.5.1",
             "machine_id": "requirement_2_5_1",
             "content": "The provider MAY define a `shutdown` function to perform whatever cleanup is necessary for the implementation.",
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -39,7 +39,7 @@ See [provider](./02-providers.md) for details.
 
 > The `provider mutator` function **MUST** invoke the `initialize` function on the newly registered provider before using it to resolve flag values.
 
-The `provider's` readiness can state can be determined from it's `status` member/accessor.
+The `provider's` readiness can state can be determined from its `status` member/accessor.
 
 See [provider initialization](./02-providers.md#24-initialization).
 
@@ -47,7 +47,7 @@ See [provider initialization](./02-providers.md#24-initialization).
 
 >  The `provider mutator` function **MUST** invoke the `shutdown` function on the previously registered provider once it's no longer being used to resolve flag values.
 
-Setting a new provider means the previous provider is no longer in use, and should therefore be disposed of using it's `shutdown` function.
+Setting a new provider means the previous provider is no longer in use, and should therefore be disposed of using its `shutdown` function.
 
 see: [shutdown](./02-providers.md#26-shutdown), [setting a provider](#setting-a-provider)
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -39,7 +39,7 @@ See [provider](./02-providers.md) for details.
 
 > The `provider mutator` function **MUST** invoke the `initialize` function on the newly registered provider before using it to resolve flag values.
 
-The `provider's` readiness can state can be maintained in it's `ready` member.
+The `provider's` readiness can state can be determined from it's `ready` member.
 
 See [provider initialization](./02-providers.md#24-initialization).
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -39,6 +39,8 @@ See [provider](./02-providers.md) for details.
 
 > The `provider mutator` function **MUST** invoke the `initialize` function on the newly registered provider before using it to resolve flag values.
 
+The `provider's` readiness can state can be maintained in it's `ready` member.
+
 See [provider initialization](./02-providers.md#24-initialization).
 
 #### Requirement 1.1.2.3
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -20,9 +20,11 @@ The `evaluation API` allows for the evaluation of feature flag values, independe
 
 It's important that multiple instances of the `API` not be active, so that state stored therein, such as the registered `provider`, static global `evaluation context`, and globally configured `hooks` allow the `API` to behave predictably. This can be difficult in some runtimes or languages, but implementors should make their best effort to ensure that only a single instance of the `API` is used.
 
-#### Requirement 1.1.2
+### Setting a provider
 
-> The `API` **MUST** provide a function to set the default `provider`, which accepts an API-conformant `provider` implementation.
+#### Requirement 1.1.2.1
+
+> The `API` **MUST** define a `provider mutator`, a function to set the default `provider`, which accepts an API-conformant `provider` implementation.
 
 ```typescript
 // example provider mutator
@@ -32,6 +34,20 @@ OpenFeature.setProvider(new MyProvider());
 This provider is used if there is not a more specific client name binding. (see later requirements).
 
 See [provider](./02-providers.md) for details.
+
+#### Requirement 1.1.2.2
+
+> The `provider mutator` function **MUST** invoke the `initialize` function on the newly registered provider before using it to resolve flag values.
+
+See [provider initialization](./02-providers.md#24-initialization).
+
+#### Requirement 1.1.2.3
+
+>  The `provider mutator` function **MUST** invoke the `shutdown` function on the previously registered provider once it's no longer being used to resolve flag values.
+
+Setting a new provider means the previous provider is no longer in use, and should therefor be disposed of using it's `shutdown` function.
+
+see: [shutdown](./02-providers.md#26-shutdown), [setting a provider](#setting-a-provider)
 
 #### Requirement 1.1.3
 
@@ -105,11 +121,11 @@ See [hooks](./04-hooks.md) for details.
 client.getMetadata().getName(); // "my-client"
 ```
 
-#### 1.3. Flag Evaluation
+### 1.3. Flag Evaluation
 
 [![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
 
-##### Requirement 1.3.1
+#### Requirement 1.3.1
 
 > The `client` **MUST** provide methods for typed flag evaluation, including boolean, numeric, string, and structure, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required), `evaluation context` (optional), and `evaluation options` (optional), which returns the flag value.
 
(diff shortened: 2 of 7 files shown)


specification.json
```diff
@@ -282,23 +303,23 @@
...
+                    "content": "The `provider` SHOULD populate the `resolution details` structure's `flag metadata` field.",
+                    "RFC 2119 keyword": "SHOULD",
+                    "children": []
+                },
+                {
+                    "id": "Requirement 2.2.10",
+                    "machine_id": "requirement_2_2_10",
+                    "content": "`flag metadata` MUST be a structure supporting the definition of arbitrary properties, with keys of type `string`, and values of type `boolean | string | number`.",
+                    "RFC 2119 keyword": "MUST",
+                    "children": []
```
toddbaert, line 319: This is not new - git is just seeing it as an add/remove.

specification/sections/01-flag-evaluation.md
```diff
@@ -119,32 +135,31 @@ MyStruct myStruct = client.getObjectValue<MyStruct>('structured-flag', { text: '
...
 > The client **SHOULD** provide functions for floating-point numbers and integers, consistent with language idioms.
 
-```go
-// example in GO
-GetIntValue(flag string, defaultValue int64, evalCtx EvaluationContext, options ...EvaluationOption) (int64, error)
+```java
+int getIntValue(String flag, int defaultValue);
 
-GetFloatValue(flag string, defaultValue float64, evalCtx EvaluationContext, options ...EvaluationOption) (float64, error)
+long getFloatValue(String flag, long defaultValue);
```
toddbaert, line 149: I move we try to use Java examples as they are generally well-understood and familiar to most.

specification/sections/01-flag-evaluation.md
```diff
@@ -31,6 +33,20 @@ OpenFeature.setProvider(new MyProvider());
 
 See [provider](./02-providers.md) for details.
 
+#### Requirement 1.1.2.2
```
justinabrahms, line 36: What happens if either `initialize` or `shutdown` fail for some reason?

toddbaert: Good question.

I think the provider stays in a bad state (assuming the initialize is necessary for it to function properly) and probably throws during evaluation, and so the SDK will default for all evaluations.

Long term, my thinking is a failed initialize would fire the `ProviderEvents.ERROR` event as well, which would allow application authors to "do something" when this happens by attaching an event handler. I think that even in this case, the provider would still default until some corrective action is taken.

specification/sections/01-flag-evaluation.md
```diff
@@ -31,6 +33,20 @@ OpenFeature.setProvider(new MyProvider());
...
+
+> The `provider mutator` function **MUST** invoke the `initialize` function on the newly registered provider before using it to resolve flag values.
```
beeme1mr, line 38: Do you think we should specify an initialization timeout? I'm sure this could be handled in a provider or userland but it may be worth explicitly defining the maximum amount of time the initialization process will wait.

specification/sections/01-flag-evaluation.md
toddbaert: I don't think I did a great job explaining my thinking here, so I'll try again :sweat_smile: 

I think that a lot of the things that can "go wrong" during the initialization in the `setProvider` can go wrong outside that context too. Persistent connections can be lost, APIs can become unavailable, etc. Therefor, we already have to handle these situations "out-of-band" of the `setProvider` function anyway, and I think using events is the best approach for this. For this reason, I think it's best to keep things simple here, and basically just log warnings if things go wrong in the `shutdown/initialization` within the `provider mutator`. The provider may attempt to re-establish whatever state is needed to work correctly, or give up if the error is irrecoverable. If it throws, the SDK will default.

That's my line of thinking at the moment.

weyert: I think the `shutdown` is typically only called when the process exists so I think throwing an event at such a moment would not be really helpful. As don't think you can really act on it at least in Node.js. For example, if you get a SIGKILL signal or something similar.

I assume you are talking when you want to call `shutdown` during the life time of the application. I have to admit I personally haven't found a use case for switching providers in a running application but I think logging a warning works to notify that something went wrong and needs attention.

specification/sections/01-flag-evaluation.md
weyert: Do we want to clarify what the expected order is? Do we want to call `shutdown` on the previous provider before calling `initialise` on the new provider or the other way around?

specification/sections/01-flag-evaluation.md
toddbaert: > I think the `shutdown` is typically only called when the process exists so I think throwing an event at such a moment would not be really helpful. As don't think you can really act on it at least in Node.js. For example, if you get a SIGKILL signal or something similar.
> 
> I assume you are talking when you want to call `shutdown` during the life time of the application. I have to admit I personally haven't found a use case for switching providers in a running application but I think logging a warning works to notify that something went wrong and needs attention.

I think we agree. In this model here are 2 places a provider's shutdown is called:

1) when switching away from that provider (a new provider is set)
2) when the API's shutdown is called (this would be used when the application is terminating, as you mention)

Events don't help much in either case for `shutdown`. For errors during/after `initialize`, I think events are important. Events could be important to signal to the `application author` that a connection was lost or some fundamental state of the provider has changed, even long after initialization.

specification/sections/01-flag-evaluation.md
toddbaert: @weyert I've tried to cover that [here](https://github.com/open-feature/spec/pull/179/files#diff-1443054d965453435ba3da2907dabbc282d2f808c3a64800f10e05fab8351b1aR11-R27).

Basically, this means that the system should never use a provider that hasn't been initialized or shutdown.

specification/sections/01-flag-evaluation.md
justinabrahms: So I think this proposal may be blocked until our events spec is landed.

Reading this: https://github.com/open-feature/spec/pull/171/files#diff-daf30d840f7e7b1752899ac2c4b268af214e6c8e7058549c8d4218abd2e86ba3

It would suggest that event handlers are registered on clients.  This would mean our code would look something like:

```java
OpenFeature api = OpenFeature.getInstance();
Client client = api.getClient('client-name')
client.addHook(ProviderEvent.ERROR, ....)
api.setProvider(new MyProvider())
```

Having to get a client before you've registered a provider seems kinda odd to me?

specification/sections/01-flag-evaluation.md
justinabrahms: What should the behavior be if someone tries to resolve flag values before the provider is in a ready state?

beeme1mr: I think the default value should be returned with the [error code](https://docs.openfeature.dev/specification/types#error-code) `PROVIDER_NOT_READY`.

justinabrahms, line 38: I tried to implement this in Java. It turns out that it's more complex than you'd expect b/c the Provider doesn't know if it's ready or not, per the spec. Only the global singleton knows.

This means the API will need to manage the lifecycle which is really awkward. I'd like it if we could add something to the spec which says "providers need to be able to tell us what it's current state is on demand via some accessor function".

specification.json
```diff
@@ -320,6 +341,20 @@
             "RFC 2119 keyword": "MAY",
             "children": []
         },
+        {
```
kinyoklion, line 344: Depending on what we mean by breaking, new methods on an interface may be breaking depending on the language. (For instance the C# SDK uses C# 7.3, which doesn't have default interface implementations, so extending an interface is considered breaking). Though, those languages could likely use two interfaces.

Just highlighting it because it would be easy to accidentally do this in a breaking way.

(Similarly interface extenstion may be seen as a breaking change for testing, because people implement client interfaces and things, and may not have default implementations.)

specification/sections/01-flag-evaluation.md
Kavindu-Dodan: one option is to mandate a `provider state` which is exposed through the accessor. For example, `ready`, `pending` or `error` state, which can be used to react in evaluations or other usages 🤔 

And in case if the provider does not support init and shutdown, the default can be set to `ready`.

toddbaert: > I think the default value should be returned with the [error code](https://docs.openfeature.dev/specification/types#error-code) `PROVIDER_NOT_READY`.

This is my feeling as well.

toddbaert: > I tried to implement this in Java. It turns out that it's more complex than you'd expect b/c the Provider doesn't know if it's ready or not, per the spec. Only the global singleton knows.
> 
> This means the API will need to manage the lifecycle which is really awkward. I'd like it if we could add something to the spec which says "providers need to be able to tell us what it's current state is on demand via some accessor function".



> one option is to mandate a `provider state` which is exposed through the accessor. For example, `ready`, `pending` or `error` state, which can be used to react in evaluations or other usages thinking
> 
> And in case if the provider does not support init and shutdown, the default can be set to `ready`.

I think I agree. I actually implemented this initially, but then removed it for simplicity... but you're right, it means the API has to track it, which IS awkward. I think it makes sense to add this back to the provider.

specification.json
toddbaert: Absolutely. I think SDK implementations will have to be careful in implementing this. As much as possible we'd like to avoid a new major version.

specification/sections/01-flag-evaluation.md
justinabrahms: @toddbaert I was thinking that we'd close out the events spec first, then this one.. not that we'd combine them.

If you expect someone to respond to a provider being in an error state with a log or metric or something, they will need to know about it. What I was attempting to show with the above block of code is that provider registration happens on the API level, but handling the errors that are generated happens at the Client level. This seems incongruous.

specification/sections/01-flag-evaluation.md
justinabrahms: Boolean indicating whether it's ready seems sufficient to implement this.

Kavindu-Dodan: 👍 for a boolean with a ready state

specification/sections/01-flag-evaluation.md
toddbaert: > What I was attempting to show with the above block of code is that provider registration happens on the API level, but handling the errors that are generated happens at the Client level. This seems incongruous.

At first I thought this was an issue as well, but I think that this is a separation we want, TBH. The code setting the provider likely can't really react to such errors, I think it makes sense to have "client type" code react to it... I don't think that (for example) an alternative provider is going to be set because of an error. I think it's more realistic that areas of the app consuming feature flags (using `client` objects) would take some action or revert to some high-level default.

toddbaert: @justinabrahms another possibility would be to add global events as well, and perhaps scope them with the same scope identifiers you're proposing in your multi-provider proposal. I'm not convinced this is a better way to go at this point though.

specification/sections/01-flag-evaluation.md
toddbaert: OK I will add this.

Kavindu-Dodan (approved):
My only concern is regarding the newly added `initialize` & `shutdown` functions and attempting to adopt SDKs without breaking current implementations. I assume this will be possible for some languages (ex:- Java) but for others (ex:- Go) it can be a breaking change. 

Regardless, approval from my end 

____

Addendum - Investigation for **Go** to have non-breaking changes 

Given that Go does not have default methods, a workaround would be to rely on embedding [1]. Consider the playground example - https://go.dev/play/p/DKv-arO2PeZ

- Introduce another interface - let’s call this **Events**
- Embed  **Events** into current **Provider** contract **FeatureProvider**
- SDK provides a noop implementation for **Events** which can be used by provider implementations

[1] - https://go.dev/doc/effective_go#embedding

specification/sections/01-flag-evaluation.md
toddbaert: Added, see [this commit](https://github.com/open-feature/spec/pull/179/commits/164071b5a1ad19553c19156c8a0b68521a9324d9) for the delta. I hope it's clear enough.

@Kavindu-Dodan @justinabrahms

toddbaert: Added, see [this commit](https://github.com/open-feature/spec/pull/179/commits/164071b5a1ad19553c19156c8a0b68521a9324d9) for the delta. I hope it's clear enough.

@Kavindu-Dodan @justinabrahms

specification.json
toddbaert: @Kavindu-Dodan has some good thoughts about this WRT the Go SDK.

specification/sections/01-flag-evaluation.md
Kavindu-Dodan: Fine from my end :)

specification/sections/02-providers.md
```diff
@@ -157,3 +157,68 @@ class MyProvider implements Provider {
...
+### 2.5. Shutdown
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+#### Requirement 2.5.1
+
+> The provider **MAY** define a `shutdown` function to perform whatever cleanup is necessary for the implementation.
+```java
+// MyProvider implementation of the dispose function defined in Provider
+class MyProvider implements Provider, AutoDisposable {
```
beeme1mr, line 212: Why does this provider implement the `AutoDisposable` method? Is `dispose` not on the `Provider' interface?

toddbaert: AutoDisposable is an builtin language interface in Java, like [IDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-7.0) in C#. https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

We may or may not want to use it, that's up to the SDK implementer, but this is the reason why I don't want to be prescriptive about the naming of this function. I want to give SDK implementors the flexibility to call the method what they want so they can make that decision for their language.

beeme1mr (approved):
Looks great! It's a small but important feature.

specification/sections/01-flag-evaluation.md
toddbaert: @Kavindu-Dodan @justinabrahms @beeme1mr

I'm trying to decide exactly what makes more sense here... though it's simple, I'm somewhat unsure.

In the experimental web SDK, the SDK calls the `initialize()` function when the provider is set, and if that succeeds it updates it's own state tracking the provider's readiness (this is the awkwardness @justinabrahms referred to above). I see the point that maintaining this on the provider feels better, but what I'm not sure of is if the SDK should set that value or the provider should. Does it make more sense for the SDK to run the `initialize()` function and then set this field to ready? Or should we leave that field as read-only from outside and rely on the provider author to set it, and specify that the provider should set it during `initialize()`?

Should we specify something like `the intitialize() function will run if the "ready" state is set to false`? I just worry that this will be an easy thing for providers to screw up. I'm wondering if we should actually just maintain the state in the SDK itself after all.

specification/sections/01-flag-evaluation.md
justinabrahms: It seems like we should have some mechanism to detect if there are errors during initialization without requiring a client. But if there is broad disagreement w/ that approach then don't let this be a blocker.

toddbaert: @justinabrahms 

> It seems like we should have some mechanism to detect if there are errors during initialization without requiring a client.

I spent some time thinking about this. I think that in most real-world production cases, it's the [`application author`](https://openfeature.dev/specification/glossary#application-author) doing the feature evaluation and dealing with the client abstraction that needs to handle errors (loss of connection etc). The author of that particular part of the application knows how best to handle errors in that area.

That said, there's 2 use-cases I can think of where an [`application integrator`](https://openfeature.dev/specification/glossary#application-integrator) would need similar error signals:

- when initially configuring a provider, it might be tedious to have to get a client to troubleshoot connection issues, authentication issues, etc
- for global monitoring/telemetry purposes (sending a message to sentry or similar, for instance)

So I think I see your point. I have two ideas, which are not mutually exclusive:

1) cause `setProvider` to throw after a (configurable?) timeout, if there's an error in `initialize()`

This would at least help with the first problem. It doesn't really help too much with the second, since setting the provider is a one-time thing, and subsequent errors with the provider aren't captured without attaching a client. I think this would be easy to implement, and it wouldn't violate our "don't throw" rule since that doesn't apply to configuration. Practically, this might constitute a breaking change for JS implementations, since errors here would be associated with asynchronous code, meaning the `setProvider` would need to return a promise.

2) add "global" events. 

These would be similar to the client events, but not scoped to a single provider, wouldn't require the creation of a client. I think we still want most events to be associated with a client (expecially because in a mutli-provider scenario, we need a way to scope events). I think global events could be useful for adding handlers for monitoring the system in it's entirety, and not so much application-level concerns like flag value changes. Put another way, global events would be useful to  [`application integrators`](https://openfeature.dev/specification/glossary#application-integrator) and client events would be useful to [`application authors`]([`application authors`](https://openfeature.dev/specification/glossary#application-author)). We may want to only support error event types in global events.

specification/sections/01-flag-evaluation.md
lukas-reining: Generally it feels better, to have the state of the provider as a property of the provider.
I would totally agree that lowering the coherence of the provider by pulling provider state out of the provider is somewhat awkward.

>  I just worry that this will be an easy thing for providers to screw up. 

This could be mitigated by the SDK implementations. E.g. the default implementation of that method could just be setting the `ready state` to true, so providers would have to overwrite the method and call the super method.
Also It seems to me that specifying that the `ready flag` has to be set to true after running the initialize, should be enough to test providers for their spec compliance.

lukas-reining: What I was also wondering about is what happens if the client using that provider is doing feature flag evaluation before the `ready flag` is set to true.
I think there is nothing yet defined in the spec right? If thats true I think we should add that to the specification.
I think the spec should mention that in that case a PROVIDER_NOT_READY should be returned.

toddbaert: > This could be mitigated by the SDK implementations. E.g. the default implementation of that method could just be setting the `ready flag` to true, so providers would have to overwrite the method and call the super method.

Not all languages have default implementations, but I think there's equivalents and I agree with your point in principle.

> Also It seems to me that specifying that the `ready flag` has to be set to true after running the initialize, should be enough to test providers for their spec compliance.

To be clear: do you feel the SDK should set this, or the provider should? Is this a requirement on the SDK or the provider author?

> I think the spec should mention that in that case a PROVIDER_NOT_READY should be returned.

I think this might be a good idea. Yes. I'll think about that a bit more.

lukas-reining: > Not all languages have default implementations, but I think there's equivalents and I agree with your point in principle.

Yeah thats true, I just had in mind that in many cases this at least weakens the problem.

> To be clear: do you feel the SDK should set this, or the provider should? Is this a requirement on the SDK or the provider author?

I would say it is a requirement on the provider author, or maybe better the provider implementation.
This also means that if the provider defines that flag, it should always be initially false, which is different from not defining the flag as the spec says that in this case the provider should be considered ready.

> I think this might be a good idea. Yes. I'll think about that a bit more.

Sounds good! 👍

thomaspoignant (approved):
Super nice addition, this is actually something I already need in the GO SDK.

specification/sections/01-flag-evaluation.md
toddbaert: > This also means that if the provider defines that flag, it should always be initially false, which is different from not defining the flag as the spec says that in this case the provider should be considered ready.

I think there are providers that are "immediately ready" such as those that pull from environment variables or otherwise get their configuration synchronously from their environment.

weyert: I think it is reasonable to fallback to the default value of the flag when the provider is not ready yet. And then pass the reason as suggested in the evaluation result.  Then the implementor/consumer of OpenFeature can decide whether the want to postpone execution until the provider is ready or fallback to the default value of the flag?

weyert (approved):

specification/sections/01-flag-evaluation.md
toddbaert: @lukas-reining @weyert thanks!

I've made the changes described above. I've made the language clearer about the fact it's the provider's responsibility to set the ready member, and also added a point about resolution with providers that are not ready. See [delta](https://github.com/open-feature/spec/pull/179/commits/e5ba035ce2f256e8259dd9d9a009123c4d1733c6).

specification/sections/01-flag-evaluation.md
justinabrahms: Global events could be annotated with the provider they reference. This would allow folks to tie into all events at once, which would be useful for the two use-cases mentioned above.

I agree that this is an application integrator concern. I don't have a preference between the two options you have. 2 seems more powerful. 1 seems easier to implement (in java).

specification/sections/01-flag-evaluation.md
```diff
@@ -31,6 +33,22 @@ OpenFeature.setProvider(new MyProvider());
...
+
+The `provider's` readiness can state can be determined from it's `ready` member.
```
justinabrahms, line 40:
```suggestion
The `provider's` readiness can state can be determined from it's `ready` member/accessor.
```
Shouldn't be doing field-level accessing in java-land.

specification/sections/02-providers.md
```diff
@@ -157,3 +157,76 @@ class MyProvider implements Provider {
...
+    this.startPolling();
+  }
+
+  //...
+}
+```
+
+#### Requirement 2.4.2
+
+> The `provider` **MAY** define a mutable boolean `ready` field which indicates the readiness of the provider.
```
justinabrahms, line 193:
```suggestion
> The `provider` **MAY** define a mutable boolean `ready` field/accessor which indicates the readiness of the provider.
```

specification/sections/02-providers.md
```diff
@@ -157,3 +157,76 @@ class MyProvider implements Provider {
...
+
+Providers without this field can be assumed to be ready immediately.
+
+#### Requirement 2.4.3
+
+> The provider **MUST** set its `ready` field if its `initialize` function terminates normally.
```
justinabrahms, line 199: Should we write a corresponding MUST NOT in the case of failure?

specification/sections/01-flag-evaluation.md
lukas-reining: Looks good, this makes it clearer🙂

specification/sections/02-providers.md
toddbaert: I've thought about this and I think it's wise to change our `ready` boolean to an enum [as is often the case](https://wiki.c2.com/?UseEnumsNotBooleans).

This way we can give more information to users (especially integrators). There's also a distinct difference between the states `I've yet to be initialized` and `I tried to initialize but experienced an error`. I don't think we can capture both of these in a `ready` boolean.

Events (again) are very complementary here. Transitions between `NOT_READY`/`READY`/`ERROR` are all probably useful events, and maintaining equivalent states seems appropriate. Thanks for your continued attention to detail. See [delta](https://github.com/open-feature/spec/pull/179/commits/8ac074681ecb4196d4ce20633a0ff987066a2465).

cc @beeme1mr

specification/sections/01-flag-evaluation.md
Kavindu-Dodan: I also see the benefit of **global events** and IMO is the right way to support events. 

When an application starts, there is a setup phase where runtime configurations are set. I see Provider registration to be part of such a setup phase. And the client can be derived at business logic which can happen after some time (seconds to minutes depending on the usage) of the startup/setup phase. And it could be too late to detect events/failures at this point. Consider a K8s app with a startup probe. A developer can decide readiness to be set only if provider initialization is successful (aka `PROVIDER_READY`).

specification/sections/02-providers.md
```diff
@@ -157,3 +157,82 @@ class MyProvider implements Provider {
...
+    this.startPolling();
+  }
+
+  //...
+}
+```
+
+#### Requirement 2.4.2
+
+> The `provider` **MAY** define a `status` field/accessor which indicates the readiness of the provider, with possible values `NOT_READY`, `READY`, or `ERROR`.
```
Kavindu-Dodan, line 193: Optional, may be a mermaid diagram to visualize the state transition 

```mermaid
---
title: Provider State
---
stateDiagram-v2
    direction LR
    [*] --> NOT_READY
    NOT_READY --> READY
    READY --> ERROR
    ERROR --> READY
```

Kavindu-Dodan: And IMO these states needs to be added to `types.md` & later connect to provider events [1] introduced at  https://github.com/open-feature/spec/pull/182 

[1] - https://github.com/open-feature/spec/blob/23c6c229d9202e3a532b595e7eb80e1476334481/specification/types.md#provider-events

specification/sections/02-providers.md
Kavindu-Dodan: Agree with this change (also see comment [1]). This change helps to handle error state which was not supported previously and help to differentiate not ready vs failure clearly. 

[1] - https://github.com/open-feature/spec/pull/179#discussion_r1184382057

specification/sections/02-providers.md
toddbaert: I love this addition.

toddbaert: [Thanks](https://github.com/open-feature/spec/pull/179/commits/6ebf5870de5ddc081d2ee3ad3e14a44f4a9670b6).

specification/types.md
```diff
@@ -93,3 +93,13 @@ A structure containing the following fields:
...
+
+### Provider Status
+
+An enumeration of possible provider states.
+
+| Status    | Explanation                                                                     |
+| --------- | ------------------------------------------------------------------------------- |
+| NOT_READY | The provider has not been initialized.                                          |
+| READY     | The provider has been initialized, and is able to reliably resolve flag values. |
+| ERROR     | The provider is initialed but is not able to reliably resolve flag values.      |
```
lukas-reining, line 105: I think we have a typo here with "initialed".

specification/sections/01-flag-evaluation.md
lukas-reining: _Not sure to which discussion to put this:_
Requirement 2.4.4 describes that transient errors can lead to the status flag changing between `ERROR` and `READY`. 
Should we specify that the change of the status Flag **MUST** lead to an event dispatched? (E.g. PROVIDER_READY and PROVIDER_ERROR)
Because if we only specify that there must be a (global) event on an error in `initialize()`, application integrators and application authors would get an inconsistent view on the state if the error is transient and status changes back to ready.

specification/types.md
jonathannorris:
```suggestion
| ERROR     | The provider is initialized but is not able to reliably resolve flag values.      |
```

specification/sections/02-providers.md
```diff
@@ -157,3 +157,98 @@ class MyProvider implements Provider {
...
+They might require the completion of an HTTP request, establishing persistent connections, or starting timers or worker threads.
+The `initialization` function is an ideal place for such logic.
+
+```java
+// MyProvider implementation of the initialize function defined in Provider
+class MyProvider implements Provider {
+  //...
+
+  // the global context is passed to the initialization function
+  void initialize(EvaluationContext initialContext) {
```
jonathannorris, line 179: Does it make sense for `initialize()` to potentially have an optional Promise / Callback mechanism to set the provider's state? That would allow you to add an optional Promise / Callback based initialization to the OpenFeature side allowing customers to require initialization before any flags are evaluated. In some server-side implementations waiting for initialization is preferable to serving defaults before the provider is ready.

toddbaert: You're absolutely right. You may have heard me mention that there's a bit of coupling between the init/shutdown features and [events](https://github.com/open-feature/spec/pull/182). In the events spec changes, we're defining means to subscribe to events, one of them being provider readiness.

specification/sections/01-flag-evaluation.md
toddbaert: > _Not sure to which discussion to put this:_ Requirement 2.4.4 describes that transient errors can lead to the status flag changing between `ERROR` and `READY`. Should we specify that the change of the status Flag **MUST** lead to an event dispatched? (E.g. PROVIDER_READY and PROVIDER_ERROR) Because if we only specify that there must be a (global) event on an error in `initialize()`, application integrators and application authors would get an inconsistent view on the state if the error is transient and status changes back to ready.

I agree in principle, but I think we'd have to put that in the events spec. Like I said in slack, these subjects are becoming a bit intertwined, but I'd like to keep the PRs isolated. I think when we release a new spec version, it should contain both init/shudown and events.

lukas-reining: Okay makes sense, so if we still want to include that, should it be added to the other PR (which is then the same problem, or what do we do to have it in the spec later @toddbaert?

toddbaert: > Okay makes sense, so if we still want to include that, should it be added to the other PR (which is then the same problem that we are talking about something not present there), or what do we do to have it in the spec later @toddbaert? Because it also feels kind of wrong to add it to that PR and talk about `initialize()` which is not present there.

`Initiailize()` is present there, that PR is based on this one. I'm just trying to keep discussion scoped - the feedback we've had before is to try to keep PRs as on-topic as possible

lukas-reining (approved):
Looks good to me now! 🙂

PR #182, opened 2023-04-25
Feat/events

toddbaert:
Adds events discussed in DRAFT client PR, and OFEPs. Note that this PR also contains the [initialization/shutdown changes](https://github.com/open-feature/spec/pull/179). To view only the delta between that and this PR, see [here](https://github.com/open-feature/spec/compare/feat/init-and-shutdown...feat/events).

There's a few non-functional changes here (fixes to formatting and minor enhancements to the parser).

Be sure to review `specification.json` for the net changes.

Most of this is implemented in the [web-sdk](https://github.com/open-feature/js-sdk/tree/main/packages/client/src) already. Some conforming (demo) providers can be found in the playground:

- [cloudbees](https://github.com/open-feature/playground/blob/main/packages/web-cloudbees-provider/src/lib/web-cloudbees-provider.ts)
- [flagsmith](https://github.com/open-feature/playground/blob/main/packages/web-flagsmith-provider/src/lib/web-flagsmith-provider.ts)
- [harness](https://github.com/open-feature/playground/blob/main/packages/web-harness-provider/src/lib/web-harness-provider.ts)
- [launchdarkly](https://github.com/open-feature/playground/blob/main/packages/web-launchdarkly-provider/src/lib/web-launchdarkly-provider.ts)
- [split](https://github.com/open-feature/playground/blob/main/packages/web-split-provider/src/lib/web-split-provider.ts)


specification/sections/01-flag-evaluation.md
```diff
@@ -163,80 +178,93 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
...
+### 1.6. Shutdown
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+#### Requirement 1.6.1
+
+> The API **MUST** define a `shutdown` function, which, when called, must call the respective `shutdown` function on the active provider.
+
+The precise name of this function is not prescribed by this specification.
+Relevant language idioms should be considered when choosing the name for this function, in accordance with the resource-disposal semantics of the language in question.
```
thomaspoignant, line 268: From a provider's point of view, it would be easier to have guidance on how to name it because I am afraid that it could be hard to keep track on all the different names in all languages.

toddbaert: I think this comment might be more valuable on [this PR](https://github.com/open-feature/spec/pull/179). I've based this on those changes, I think they might need to go together. I could combine them without an issue, I just don't want them to get too big. You can see just the diff [here](https://github.com/open-feature/spec/compare/feat/init-and-shutdown...feat/events)

thomaspoignant, line 268: From a provider's point of view, it would be easier to have guidance on how to name it because I am afraid that it could be hard to keep track of all the different names in all languages.

dabeeeenster:
I wonder if there might be need for an event type named something like `PROVIDER_STALE`. 

Flagsmith server-side SDKs can run in a mode (local evaluation mode we call it) where it polls the API for the entire environment flag configuration and then runs the flag evaluation engine locally within the SDK runtime. By default this poll occurs every 60 seconds. If the SDK does not receive a response from the API it will fall back to the most recent configuration. 

At some point this flag configuration could/should be considered stale, which would trigger the `PROVIDER_STALE` event. This would probably then be bubbled up to some logging/alerting system to let administrators know that something isnt right/stale flags are being served. 

This event could also be used on application startup if the SDK is not able to make an initial connection to the API. This is generally a worse state to be in as you would generally be serving hard-coded defaults at this point. 

I think several other providers work on a similar basis; some stream changes over websockets/server sent events.

specification/types.md
```diff
@@ -93,3 +93,13 @@ A structure containing the following fields:
...
+
+### Provider Events
+
+An enumeration of provider events.
+
+| Event                          | Explanation                                                                                 |
+| ------------------------------ | ------------------------------------------------------------------------------------------- |
+| PROVIDER_READY                 | The provider is ready to perform flag evaluations.                                          |
+| PROVIDER_ERROR                 | The provider signalled an error.                                                            |
+| PROVIDER_CONFIGURATION_CHANGED | A change was made to the backend flag configuration.                                        |
```
Kavindu-Dodan, line 105: minor - seems `PROVIDER_SHUTDOWN` is missing

toddbaert: Sorry, I meant to remove `PROVIDER_SHUTDOWN` entirely. I'm not sure it's useful enough. I can't imagine why a provider would permanently shutdown and how that would be different than an error.

specification/sections/01-flag-evaluation.md
toddbaert: @thomaspoignant I think this comment might be more valuable on https://github.com/open-feature/spec/pull/179. I've based this on those changes, I think they might need to go together. I could combine them without an issue, I just don't want them to get too big. You can see just the diff [here](https://github.com/open-feature/spec/compare/feat/init-and-shutdown...feat/events)

specification/types.md
Kavindu-Dodan: Agree with this :) I noticed the usage and this is the reason to comment

specification.json
```diff
@@ -559,6 +594,48 @@
...
+            "RFC 2119 keyword": "MAY",
+            "children": []
+        },
+        {
+            "id": "Requirement 5.2.1",
+            "machine_id": "requirement_5_2_1",
+            "content": "The `client` MUST provide an `addHandler` function for attaching callbacks to `provider events`, which accepts event type(s) and a function to run when the associated event(s) occur.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
```
lukas-reining, line 611: I am wondering if it would make sense to let providers also dispatch "custom" events.
E.g. a provider could dispatch an event like PROVIDER_STALE as @dabeeeenster said or maybe something else. This could make it easy for providers to communicate specific events. 

One problem with this proposal is, that either providers or users of the provider could couple to tightly to these events, even though this risk is small if they are only treated as "could occur" as it is for example with the PROVIDER_STALE event.

specification/sections/05-events.md
```diff
@@ -0,0 +1,60 @@
...
+toc_max_heading_level: 4
+---
+
+# 5. Events
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+## Overview
+
+`Events` allow consumers (_application integrator_, _application author_, _integration author_) to react to state changes in the provider or underlying flag management system, such as flag definition changes, provider readiness, or error conditions. A provider may emit events or run a callback indicating that it received a certain event, optionally providing data associated with that event. Handlers registered on the client are then invoked with this data.
```
lukas-reining, line 13: Is this additional data completely free?
For the evaluation context we have the targetingKey that is required. Also we defined the data types used.
Do we know of any common field or requirement to the data structure of that event data?

lukas-reining: E.g. for the PROVIDER_ERROR there could be an required field ERROR_MESSAGE that may be helpful for the consumer. If a provider would be required to provide this, or at least there was a field like  ERROR_MESSAGE it would be way easier to propagate this error to logging/monitoring.

specification.json
toddbaert: > I am wondering if it would make sense to let providers also dispatch "custom" events.

It's something worth considering. As a counterpoint, we've tended to strictly define and enumerate any identifiers or symbols that application authors might write logic around. This preserves one of OpenFeature's key goals: to allow migration between providers with minimal code changes. If we introduced custom events, when a migration happened, you might have a bunch of events that are no longer relevant.

specification/sections/05-events.md
toddbaert: Thus far I haven't seen enough consistency between underlying systems to make specific fields here practically useful. The data accompanying events varies widely between SDKs and solutions, even more than evaluation context, which as you point out almost always contains some "user identifier" which we've mapped to `targeting key`. Event data between the 5 SDKs I've demoed above don't really share much in their payloads.

toddbaert: I think you're right about this. I should add a section about this data structure and mention some error semantics.

specification/types.md
```diff
@@ -93,3 +93,13 @@ A structure containing the following fields:
 A structure which supports definition of arbitrary properties, with keys of type `string`, and values of type `boolean`, `string`, or `number`.
 
 This structure is populated by a provider for use by an [Application Author](./glossary.md#application-author) (via the [Evaluation API](./glossary.md#evaluation-api)) or an [Application Integrator](./glossary.md#application-integrator) (via [hooks](./sections/04-hooks.md)).
+
+### Provider Events
+
+An enumeration of provider events.
```
Kavindu-Dodan, line 99: Let's emphasize that these are currently accepted/agreed event types.  And to support extensibility as discussed [1], let's include a section to describe how an interested party can make a request to add additional event types through OFEP

[1] - https://github.com/open-feature/spec/pull/182/files#r1178321218

toddbaert: I think the spec should be very tight and concise. I'd like to avoid sections in the spec on how to modify it. I've not seen that sort of thing in other specifications, I think it muddies the waters.

toddbaert:
> > @dabeeeenster I think that the pattern and mode of operation you describe certainly makes sense. Can you explain how you might expect an application author to react to this event type? What sort of handler would I attach for PROVIDER_STALE events that would be distinct from PROVIDER_ERROR?
> 
> PROVIDER_ERROR would generally be very bad, and would normally mean the application is running on either default flag values if provided, or someting like false|0|"" for all flag values. PROVIDER_ERROR would most likely cause immediate "This service is broken" alerts and require immediate operational assistance.
> 
> PROVIDER_STALE would communicate that flags are being served, and the engine is running, but the values themselves might well have drifted from the flag/provider dashboard. I would expect these errors could be ignored if they were on something like the 99.99% percentile.

I think I'm convinced of the value of adding `PROVIDER_STALE` (especially after the community meeting today). I will do that.

specification.json
lukas-reining: > What we could do here is define the first set of events that we agree on (based on vendor feedback and experience). Later on, we can allow OFEPs to add additional events.

Defining the first set seems good. But OFEPs should only bring new events that are "generic enough" right?
If so this is a great thing. But it tackles a bit of a different problem than custom events would I would say.

specification/types.md
Kavindu-Dodan: The idea came from the `IANA Considerations` section and seeing the approach in RFCs (ex: - initial content [1] & registration process [2]) . However, as you said, I am not 100% sure this fully matches our requirements. 

Regardless, I assume there could be new event types of events beyond the initial ones we define here.

[1] - https://www.rfc-editor.org/rfc/rfc6749#section-11.2.2
[2] - https://www.rfc-editor.org/rfc/rfc6749#section-11.3.1

specification/sections/05-events.md
toddbaert: > I think you're right about this. I should add a section about this data structure and mention some error semantics.

I've done this, as well as added some additional constraints and clarity. See delta [here](https://github.com/open-feature/spec/pull/182/commits/50434a84555294565fd36e7dd44396f287b901ae).

specification/types.md
```diff
@@ -94,12 +94,19 @@ A structure which supports definition of arbitrary properties, with keys of type
 
 This structure is populated by a provider for use by an [Application Author](./glossary.md#application-author) (via the [Evaluation API](./glossary.md#evaluation-api)) or an [Application Integrator](./glossary.md#application-integrator) (via [hooks](./sections/04-hooks.md)).
 
+### Event Metadata
+
+A structure supporting the addition of arbitrary event data.
+It supports definition of arbitrary properties, with keys of type `string`, and values of type `boolean`, `string`, or `number`.
+It defines an `error message` field with a value of type `string`.
```
lukas-reining, line 101: Should the `error message` only be a defined field in `PROVIDER_ERROR` events metadata or in any?
Could also be in any, so that if one likes it could also be included in `PROVIDER_STALE` events, but I would rather only _define_ the field for `PROVIDER_ERROR`.

specification/sections/05-events.md
lukas-reining: Great, I just added one more question to the spec of the event data.

specification/types.md
lukas-reining: But not sure if this is important for the spec, just wondering.

specification.json
Kavindu-Dodan: > > Besides, language-specific implementation could be done in a generic way, so that specific vendors can add custom events with custom states.
> 
> Is this in the case the spec would allow for custom events?

After thinking a bit deeper on the implementation aspect, I think spec can only **recommend** supported event types. 

The reason is its provider implementation that exposes the `addHandler` function. Even if we define the `addHandler` contract (inputs and outputs), a specific provider implementation could accept events beyond what the specification says as we have no control over inputs and outputs 🤔

Thinking from Go perspective, a pseudocode would look like the below,

```go
// response as specified by spec
type EventResponse struct {
	state       string  // states - we can recommend through spec on what to support 
	description string
}

// event handler 
type EventHandler func(*EventResponse)

// existing provider 
type Provider struct {
}

// spec compliant handler? 
func (p Provider) addHandler(EventHandler) {
     // provider internals 
}
```

 @toddbaert  may be you have other perspective on this

lukas-reining: That's true, but still to be spec compliant you can only return whatever the spec defines isn't it?
So this "recommendation" is good enough to define what means to be OpenFeature compliant I would say.

Kavindu-Dodan: >  but still to be spec compliant you can only return whatever the spec defines isn't it?

Yes, this is this correct. Adhering only to spec-defined event types makes the provider OpenFeature compliant

toddbaert: > The reason is its provider implementation that exposes the `addHandler` function.

It doesn't, in this proposal. The provider has to expose some means of _emitting_ events. Handlers are added to the client.

In JS, this looks like an "emitter" on the provider, which is used to just signal the SDK that an event occurred. See this [example cloudbees provider](https://github.com/open-feature/playground/blob/main/packages/web-cloudbees-provider/src/lib/web-cloudbees-provider.ts#L56) (and the others linked in the description).

specification/types.md
toddbaert: > Should the error message only be a defined field in PROVIDER_ERROR events metadata or in any event types metadata?

I've added `error event metadata`, see [delta](https://github.com/open-feature/spec/pull/182/commits/0c0dbc111c65dbe418250b22c5b67f651b0124a5).

lukas-reining: Nice thats good!

lukas-reining (approved):

specification/sections/05-events.md
```diff
@@ -0,0 +1,82 @@
...
+
+> The `provider` **MAY** define a mechanism for signalling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with an `event metadata` payload. 
+
+If available, native event-emitter or observable/observer language constructs can be used.
+
+see: [provider event types](../types.md#provider-events), [`event metadata`](../types.md#event-metadata).
+
+#### Requirement 5.1.2
+
+> When the `provider` signals the occurrence of a particular `event`, the associated `client` `event handlers` **MUST** run.
```
justinabrahms, line 27: Can you document what happens if running one of the event handlers fails in some way? Should we be swallowing exceptions? Bubbling them up? Firing additional events?

specification/sections/05-events.md
```diff
@@ -0,0 +1,82 @@
...
+```java
+  // run the myOnReadyHandler function when the PROVIDER_READY event is fired
+  client.addHandler(ProviderEvents.Ready, MyClass::myOnReadyHandler);
+```
+
+see: [provider events](#51-provider-events)
+
+#### Requirement 5.2.2
+
+> The `event handler` function **MAY** accept a `event metadata` or `error event metadata` parameter.
```
justinabrahms, line 56: What's the thinking for MAY here? I would have guessed MUST.

specification/sections/05-events.md
```diff
@@ -0,0 +1,82 @@
...
+
+see: [`event metadata`](../types.md#event-metadata), [`error event metadata`](../types.md#error-event-metadata)
+
+#### Requirement 5.2.3
+
+> If the provider's `initialize` function terminates normally, `PROVIDER_READY` handlers **MUST** run.
```
justinabrahms, line 62: Can someone control the ordering in which these run?

specification/sections/05-events.md
toddbaert: I was trying to say it was optional, but I agree it was confusing. Updated.

specification/sections/05-events.md
toddbaert: That's an interesting question. Do you think it's necessary to specify?

specification/sections/05-events.md
toddbaert: Great point. I think the important part to capture is probably that all handlers need to run if there's abnormal execution in one. I've added [5.1.4](https://github.com/open-feature/spec/pull/182/commits/bb8fcb4bf8a673fe818a9f60abd1557782145f98). Do you think that's sufficient?

specification/sections/02-providers.md
```diff
@@ -157,3 +157,98 @@ class MyProvider implements Provider {
...
+
+> The provider **MUST** set its `status` field/accessor to `READY` if its `initialize` function terminates normally.
+
+If the provider supports the `status` field/accessor and initialization succeeds, setting the `status` to `READY` indicates that the provider is initialized and flag evaluation is proceeding normally.
+
+#### Requirement 2.4.4
+
+> The provider **MUST** set its `status` field to `ERROR` if its `initialize` function terminates abnormally.
+
+If the provider supports the `status` field/accessor and initialization fails, setting the `status` to `ERROR` indicates the provider is in an error state. If the error is transient in nature (ex: a connectivity failure of some kind) the provider can attempt to resolve this state automatically.
```
lukas-reining, line 223: This describes that transient errors can lead to the status flag changing between ERROR and READY.
I think we should specify that the change of the status Flag MUST lead to an event dispatched. (E.g. PROVIDER_READY and PROVIDER_ERROR)
Because if we only specify that there must be a (global) event on an error in initialize(), application integrators and application authors would get an inconsistent view on the state if the error is transient and status changes back to ready.

toddbaert: I see why this would be valuable.

I think we should define a point like this, but use "SHOULD" language. Beyond initialization, I think the only providers that would go into an error state asynchronously would be providers which maintain some kind of polling or persistent connection. I think that we want to think of this as an opt-in behavior.

lukas-reining: Makes sense.
The problem to this would be that listeners to that event could maybe wait for the READY event. If some providers do not dispatch this, it would be problematic.

On the other hand, if the spec only says should, one should not depend on that. But I fear that this could still happen easily.

specification/sections/05-events.md
beeme1mr: I'm assuming they would run in the order they were registered but I wonder how valuable it would be to explicitly define that behavior. Requirement 5.1.4 also states that all handlers must run even if an error occurs in one of them. I think the only other reason someone would care about the order would be if there was a shared state between multiple handlers, which seems like something we should discourage. @justinabrahms what do you think?

specification/sections/02-providers.md
lukas-reining: Ah right that makes sense.
So I am with you, SHOULD is the right thing 🙂

specification/sections/05-events.md
toddbaert: > Requirement 5.1.4 also states that all handlers must run even if an error occurs in one of them.

Right, IMO this is the important part.

> I think the only other reason someone would care about the order would be if there was a shared state between multiple handlers, which seems like something we should discourage

Definitely agree with that.

specification/sections/05-events.md
```diff
@@ -0,0 +1,86 @@
...
+
+> `PROVIDER_ERROR` events **SHOULD** populate the `error event metadata`'s `error message` field.
+
+The error message field should contain an informative message as to the nature of the error.
+
+See [event metadata](../types.md#error-event-metadata)
+
+#### Requirement 5.1.4
+
+> If a `handler functions` terminates abnormally, other event handlers **MUST** run.
```
justinabrahms, line 43:
```suggestion
> If a `handler function` terminates abnormally, other event handlers **MUST** run.
```

specification/sections/05-events.md
justinabrahms: I don't care as much about this given that we've said that they don't short circuit each other in 5.1.4.

justinabrahms (approved):

specification.json
```diff
@@ -559,6 +622,76 @@
...
+            "id": "Requirement 5.2.1",
+            "machine_id": "requirement_5_2_1",
+            "content": "The `client` MUST provide an `addHandler` function for attaching callbacks to `provider events`, which accepts event type(s) and a `event handler function`.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 5.2.2",
+            "machine_id": "requirement_5_2_2",
+            "content": "The `event handler` function MUST accept a `event metadata` or `error event metadata` parameter.",
```
Kavindu-Dodan, line 664:
```suggestion
            "content": "The `event handler` function MUST accept an `event metadata` or `error event metadata` parameter.",
```

specification/sections/01-flag-evaluation.md
```diff
@@ -33,6 +35,22 @@ This provider is used if there is not a more specific client name binding. (see
...
+
+The `provider's` readiness can state can be determined from it's `status` member/accessor.
+
+See [provider initialization](./02-providers.md#24-initialization).
+
+#### Requirement 1.1.2.3
+
+>  The `provider mutator` function **MUST** invoke the `shutdown` function on the previously registered provider once it's no longer being used to resolve flag values.
+
+Setting a new provider means the previous provider is no longer in use, and should therefore be disposed of using it's `shutdown` function.
```
beeme1mr, line 50:
```suggestion
Setting a new provider means the previous provider is no longer in use, and should therefore be disposed of using its `shutdown` function.
```

toddbaert: fixed upstream.

specification/sections/05-events.md
```diff
@@ -0,0 +1,86 @@
...
+> If the provider's `initialize` function terminates abnormally, `PROVIDER_ERROR` handlers **MUST** run.
+
+See [provider initialization](./02-providers.md#24-initialization) and [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+
+#### Requirement 5.2.5
+
+> `PROVIDER_READY` handlers added after the provider is already in a ready state **MUST** run immediately.
+
+See [provider initialization](./02-providers.md#24-initialization) and [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+
```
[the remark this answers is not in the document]
toddbaert: IMHO [5.1.1](https://github.com/open-feature/spec/compare/feat/init-and-shutdown...feat/events#diff-daf30d840f7e7b1752899ac2c4b268af214e6c8e7058549c8d4218abd2e86ba3R19) combined with the [Provider Events Type](https://github.com/open-feature/spec/compare/feat/init-and-shutdown...feat/events#diff-a9f4ee44499a09a8f2ae9d6f873888a4bcf1d6d3816da16519e5943a163e6b11R144-R151) table covers this.

specification.json
```diff
@@ -629,6 +629,90 @@
...
+            "id": "Requirement 5.1.3",
+            "machine_id": "requirement_5_1_3",
+            "content": "`PROVIDER_ERROR` events SHOULD populate the `provider event details`'s `error message` field.",
+            "RFC 2119 keyword": "SHOULD",
+            "children": []
+        },
+        {
+            "id": "Requirement 5.2.1",
+            "machine_id": "requirement_5_2_1",
+            "content": "The `client` MUST provide an `addHandler` function for associating callbacks with `provider events`, which accepts an `event type` and a `event handler function`.",
```
Kavindu-Dodan, line 657: Does this mean `addHandler` is exposed from client as well ? 

for example (Java),

```Java
Client client = api.getClient();
client.addHandler(Handler.A, Handlerimpl::A())
```

Kavindu-Dodan: ignore - found the example in events.md

specification/sections/05-events.md
```diff
@@ -0,0 +1,102 @@
...
+#### Requirement 5.2.3
+
+> The `event details` **MUST** contain the `client name` associated with the event.
+
+The `client name` indicates the client/provider with which the event is associated.
+This is particularly relevant for `event handler functions` which are attached to the `API`, not a particular client.
+
+#### Requirement 5.2.4
+
+> The `event handler` function **MUST** accept a `event details` or `error event metadata` parameter.
```
Kavindu-Dodan, line 72: The usage of `or` is confusing here. Taking Java as an example, in minimal form, the handler can be a functional interface.

```
public interface EventHandler {
    void handle(EventDetails details);
}
```

Here, I expect `EventDetails` to be common for any event type. Otherwise, I believe language-specific restrictions will make it difficult to expose a single `addHandler(EventHandler)` method as the method signature will change.

specification/sections/05-events.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Overview
+
+`Events` allow consumers (_application integrator_, _application author_, _integration author_) to react to state changes in the provider or underlying flag management system, such as flag definition changes, provider readiness, or error conditions. A provider may emit events or run a callback indicating that it received a certain event, optionally providing data associated with that event. Handlers registered on the client are then invoked with this data.
+
+### 5.1. Provider events
+
+#### Requirement 5.1.1
+
+> The `provider` **MAY** define a mechanism for signalling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with an `event metadata` payload.
```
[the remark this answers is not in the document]
toddbaert: fixed

moredip (approved):

moredip (approved):

specification/types.md
```diff
@@ -109,29 +109,16 @@ An enumeration of possible provider states.
 A structure defining a provider event payload, including:
 
 - flags changed (string[], optional)
+- message (string, optional)
```
toddbaert, line 112: error and non-error message can go here now.

specification/sections/05-events.md
toddbaert: Hey @Kavindu-Dodan , you're right. The thing I was trying to do here was to note that events could contain the `error event details` or the `event details`... but I've consolidated both of these into a single type which has a message. The message could be useful in both error and non-error scenarios, so I think the distinction isn't worth the complexity.

Please see delta [here](https://github.com/open-feature/spec/pull/182/commits/e09a20ae9906a05e1aed8ca265ca7ceb4e3d7e21).

specification/types.md
```diff
@@ -103,3 +103,49 @@ An enumeration of possible provider states.
...
+
+### Provider Events
+
+An enumeration of provider events.
+
+| Event                          | Explanation                                                                                         |
+| ------------------------------ | --------------------------------------------------------------------------------------------------- |
+| PROVIDER_READY                 | The provider is ready to perform flag evaluations.                                                  |
+| PROVIDER_ERROR                 | The provider signalled an error.                                                                    |
+| PROVIDER_CONFIGURATION_CHANGED | A change was made to the backend flag configuration.                                                |
```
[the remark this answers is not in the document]
toddbaert: I've added a bunch of non-normative clarifying text that helps with this, along with other things, I think.

See delta [here](https://github.com/open-feature/spec/pull/182/commits/cf4932fb42f129e809ac504ba33b95d6471b9419).

specification/sections/05-events.md
```diff
@@ -75,28 +77,37 @@ see: [`event details`](../types.md#event-details), [`error event metadata`](../t
...
 
 #### Requirement 5.2.6
 
-> If the provider's `initialize` function terminates abnormally, `PROVIDER_ERROR` handlers **MUST** run.
+> Event handlers **MUST** persist across `provider` changes.
 
-See [provider initialization](./02-providers.md#24-initialization) and [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+Behavior of event handlers should be independent of the order of handler addition and provider configuration.
+
+### Event handlers and initialization
```
toddbaert, line 88: No new spec content here, I just moved these into their own section

specification/sections/05-events.md
Kavindu-Dodan: Thanks I am having another look into this :)

specification.json
```diff
@@ -230,7 +230,14 @@
...
             "machine_id": "requirement_1_6_1",
-            "content": "The API MUST define a `shutdown` function, which, when called, must call the respective `shutdown` function on the active provider.",
+            "content": "The API MUST define a `shutdown` function which, when called, must call the respective `shutdown` function on the active provider.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 1.6.2",
+            "machine_id": "requirement_1_6_2",
+            "content": "The client MUST define a `shutdown` function which, when called, removes all the event handlers associated with that client.",
```
Kavindu-Dodan, line 240: I think we should also define the client's behavior after the shutdown. I assume the client will act as backed by a noop provider.

specification.json
```diff
@@ -629,6 +636,97 @@
             "content": "The hook MUST NOT alter the `hook hints` structure.",
             "RFC 2119 keyword": "MUST NOT",
             "children": []
+        },
+        {
+            "id": "Requirement 5.1.1",
+            "machine_id": "requirement_5_1_1",
+            "content": "The `provider` MAY define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with an `provider event details` payload.",
```
Kavindu-Dodan, line 643: Discussion - Doubt about the wording `MAY` & the contract agreed between a Provider and the API. 

For events to be subscribed and unsubscribed, I see the need of Provider to expose a method to add event subscriptions (ex:- `providerA.addHandler(Event.TypeA, HandlerImpl)`). I think spec must enforce this, so that API can work correctly.

specification/sections/05-events.md
```diff
@@ -0,0 +1,129 @@
...
+The data that providers supply in event payloads may include a list of `flag keys` changed, error messages, and possibly updated flag values.
+
+```mermaid
+graph
+    P(Provider) -->|emit event| A[API]
+    A -->|run handlers| AH("API (global) event handlers")
+    A --> C[Client]
+    C -->|run handlers| CH(Client event handlers)
+```
+
```
Kavindu-Dodan, line 24: Thank you for the diagram. It cleared my doubts  🙌

specification.json
toddbaert: The idea is that events are still optional. If there's nothing for the SDK internals to subscribe to, there will simply be no asynchronous events (such as `FLAG_CONFIGURATION_CHANGED`) fired.

The only events that are guaranteed to fire in any case, are the `PROVIDER_READY` and `PROVIDER_ERROR` events, if the `initialization` succeeds or fails, respectively.

See here: https://github.com/open-feature/spec/pull/182/files#diff-daf30d840f7e7b1752899ac2c4b268af214e6c8e7058549c8d4218abd2e86ba3R104-R107

specification/sections/05-events.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Events
```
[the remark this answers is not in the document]
toddbaert: Ya, though I'd love to see all SDKs implement this, I think an implementation could work _well enough_ without it.

Kavindu-Dodan (approved):
I think I'm convince on the completeness of this addition. Left one more concern about client shutdown [1] , but the rest looks good. Thank you @toddbaert for taking time on explaining core concepts and clearing doubts

[1] - https://github.com/open-feature/spec/pull/182#discussion_r1188866376

specification.json
toddbaert: It's a minor point, but I'm a bit unsure about this now. I think we have 3 options:

- keep this as is, and add specific directions about "closed clients"
- keep this as is, and say that the behavior of closed client is undefined otherwise
- remove the client close point, and instead describe that handlers can be removed

Kavindu-Dodan: I am leaning towards the `remove the client close point, and instead describe that handlers can be removed` option. IMO it's too complex and unnecessary to spec out `closed clients`. But I am open for opinion of others here :)

toddbaert: @Kavindu-Dodan I've made such a change. See [delta](https://github.com/open-feature/spec/pull/182/commits/a4e70ab3e89bdb23e145cd73e14bdce11022417c). I've kept it pretty vague since different languages might have different patterns for removing listeners/observers. The important part is they need to be able to be removed/deactivated, IMO.

Kavindu-Dodan: I verified the delta and I am satisfied with the change. Yes, agree with the fact that implementation details can vary, but spec must define the requirement :) 

Will resolve this conversation.

specification/sections/05-events.md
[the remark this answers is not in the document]
Kavindu-Dodan: @toddbaert  agree. I like the `extension` approach to making events optional. Providers can opt-in if they support events and if they do, API & Clients will handle them in the way the spec defines.

specification/sections/05-events.md
beeme1mr, line 1: Should we consider adding `remove handlers` methods? This seems valuable at least on the global API because there's no way to create a new instance of it.

### Remove all ready handlers on the global API

```java
OpenFeature.removeHandlers(ProviderEvents.Ready);
```

### Remove all handlers on the global API

```java
OpenFeature.removeHandlers();
```

This same pattern could be applied to the client as well.

specification.json
```diff
@@ -45,7 +45,7 @@
         {
             "id": "Requirement 1.1.5",
             "machine_id": "requirement_1_1_5",
-            "content": "The API MUST provide a function for retrieving the metadata field of the configured `provider`.",
+            "content": "The `API` MUST provide a function for retrieving the metadata field of the configured `provider`.",
```
tcarrio, line 48: If we are going to monospace this it should be consistent across all entries. It's probably not necessary to make that change though

specification/sections/05-events.md
Kavindu-Dodan: I think this is covered by the new delta  - https://github.com/open-feature/spec/pull/182/commits/a4e70ab3e89bdb23e145cd73e14bdce11022417c

beeme1mr: Ah, you're right. Perhaps including an example in the non-normative section would be useful but not required.

kinyoklion (approved):

specification.json
toddbaert: I agree. I noticed a few small inconsistencies like that. I'll create a separate PR that addresses some of the style inconsistencies, and maybe even see if we can do some basic custom linting.

specification/sections/05-events.md
toddbaert: @beeme1mr  see [this change](https://github.com/open-feature/spec/pull/182/commits/5f4c75b16027d9c81bd26d8f149c8d8fe8f09215).

lukas-reining (approved):

PR #183, opened 2023-04-27
Spec change for named client -> provider mappings

justinabrahms:
## This PR

Adds support for mapping a named client to a specific provider.

### Related Issues
Refs https://github.com/open-feature/ofep/pull/56

### Notes

Example of this implemented in Java: https://github.com/open-feature/java-sdk/pull/388

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -17,7 +17,7 @@
         {
             "id": "Requirement 1.1.3",
             "machine_id": "requirement_1_1_3",
-            "content": "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s.",
+            "content": "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -10,34 +10,41 @@
         {
             "id": "Requirement 1.1.2",
             "machine_id": "requirement_1_1_2",
-            "content": "The `API` MUST provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.",
+            "content": "The `API` MUST provide a function to set the default `provider`, which accepts an API-conformant `provider` implementation.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.1.3",
             "machine_id": "requirement_1_1_3",
-            "content": "The `API` MUST provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
+            "content": "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.1.4",
             "machine_id": "requirement_1_1_4",
-            "content": "The API MUST provide a function for retrieving the metadata field of the configured `provider`.",
+            "content": "The `API` MUST provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.1.5",
             "machine_id": "requirement_1_1_5",
-            "content": "The `API` MUST provide a function for creating a `client` which accepts the following options:  - name (optional): A logical string identifier for the client.",
+            "content": "The API MUST provide a function for retrieving the metadata field of the configured `provider`.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.1.6",
             "machine_id": "requirement_1_1_6",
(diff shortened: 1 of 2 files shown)


specification/sections/01-flag-evaluation.md
```diff
@@ -33,6 +33,14 @@ See [provider](./02-providers.md) for details.
...
 #### Requirement 1.1.3
 
+> The `API` **MUST** provide a function to bind a given `provider` to one or more client `name`s.
+
+```java
+OpenFeature.setProvider("client-name", new MyProvider());
+```
+
+#### Requirement 1.1.4
+
```
toddbaert, line 43: I'm interested in why you chose to go with the "number shifting" rather than add this to the end of this section (I think this is fine, I'm just wondering if you came to any conclusions).

Also, have you given any more thought to non-numeric requirements?

specification/sections/01-flag-evaluation.md
```diff
@@ -22,7 +22,7 @@ It's important that multiple instances of the `API` not be active, so that state
 
 #### Requirement 1.1.2
 
-> The `API` **MUST** provide a function to set the global `provider` singleton, which accepts an API-conformant `provider` implementation.
+> The `API` **MUST** provide a function to set the default `provider`, which accepts an API-conformant `provider` implementation.
```
toddbaert, line 25: I think it might be nice to add some non-normative text below this about how the `default` provider is different from `client-name` bound providers.

toddbaert:
I think we might need additional point(s) about retrieving a named client bound to a non-default provider. I'm also wondering if we need to spec that clients retrieved with a given name return the same object instance (or do they just generate a new client associated with the provider bound to that name?)

specification.json
```diff
@@ -10,34 +10,41 @@
...
-            "content": "The `API` MUST provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
+            "content": "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.1.4",
             "machine_id": "requirement_1_1_4",
-            "content": "The API MUST provide a function for retrieving the metadata field of the configured `provider`.",
+            "content": "The `API` MUST provide a function to add `hooks` which accepts one or more API-conformant `hooks`, and appends them to the collection of any previously added hooks. When new hooks are added, previously added hooks are not removed.",
```
lukas-reining, line 27: Should it be possible to add different hooks to different providers?

specification.json
```diff
@@ -10,34 +10,41 @@
...
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.1.6",
             "machine_id": "requirement_1_1_6",
+            "content": "The `API` MUST provide a function for creating a `client` which accepts the following options:  - name (optional): A logical string identifier for the client.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
```
lukas-reining, line 44: What happens on a name collision here?
1.1.7 says that client creation MUST NOT throw or terminate abnormally. 
This results in either having to overwrite the previous client or the operation being a No-Op right?

toddbaert: > This results in either having to overwrite the previous client or the operation being a No-Op right?

Ya, I think so.

I think it makes sense to overwrite it, and this should be specified. Somewhat related is [this](https://github.com/open-feature/spec/pull/183#issuecomment-1527941473) point.

lukas-reining: Yeah overwriting sounds good and should be documented!

specification/sections/01-flag-evaluation.md
beeme1mr: I would like to avoid shifting spec version numbers if possible. @justinabrahms, would it be possible to add the changes to new sections? If not, we'll want to make sure SDK maintainers update test names based on the new requirement numbers.

justinabrahms: non-numeric requirements are out of scope for this change. I added this where I did so that it's numerically in order and located along side the other "set the provider" spec stuff. I could also name this `1.1.3-b` or similar, if you prefer?

specification.json
justinabrahms: Generating a client will never overwrite, nor would it ever throw. Not sure where the confusion is on that one.

If there's a name collision on the provider registration, it would be the same behavior as a map type (e.g. overwrite). I'll update the document.

specification/sections/01-flag-evaluation.md
beeme1mr: No, that's ok. Your reasoning makes sense to me. Thanks for clarifying!

toddbaert (approved):
I think this is a good addition in principle, and the requirements is succinct enough. I'm still slightly concerned about [this](https://github.com/open-feature/spec/pull/183#issuecomment-1527941473), but I'm not convinced it's something we _need_ to specify.

lukas-reining (approved):
Spec seems fine like this and the use case is totally there.

thomaspoignant (approved):
This addition will make the switch between providers even simpler 👏

beeme1mr (approved):

kinyoklion (approved):

Issue #192, opened 2023-05-31
[bug] Multiple providers bound to one "name", and associated issues


benjiro:
Spec: open-feature/spec@a4ffec3
See: https://github.com/open-feature/dotnet-sdk/issues/126

The spec allows a single instance of a provider to be bound to multiple names (even if it didn't allow this, any naive implementation would probably allow it to happen). We either need to prevent it or specify how it should work. We may want to specify that providers that are still bound to any client name are not shut down.  - @toddbaert 

I consider this a spec bug personally, because it constitutes a problematic ambiguity in the spec. - @toddbaert 

cc @justinabrahms 

Quote from @kinyoklion :+1: 

> Basic case.
> ```
> var provider1 = new TestProvider();
> 
> openFeature.SetProvider("a",  provider1);
> 
> // Call init on provider1?
> 
> openFeature.SetProvider("a",  new TestProvider());
> 
> // Call shutdown on provider 1?
> 
> // Call init on new provider.
> 
> ```
> 
> Two Names Scenario 1.
> ```
> var provider = new TestProvider();
> 
> openFeature.SetProvider("a", provider);
> 
> // Call init on provider?
> 
> openFeature.SetProvider("b", provider);
> 
> // Provider already has init called. Do we call it again?
> 
> ```
> 
> Two Names Scenario 2.
> ```
> var provider1 = new TestProvider();
> 
> openFeature.SetProvider("a", provider1);
> openFeature.SetProvider("b", provider1);
> 
> openFeature.SetProvider("b",  new TestProvider());
> 
> // We don't want to shutdown provider1, because it is still registered to "b".
> 
> ```

toddbaert:
cc @lopitz @lukas-reining

toddbaert:
@benjiro FYI I moved this from the dotnet-sdk repo.

lukas-reining:
I totally agree.
Also found that and implemented it in the JS SDK the way that this does not happen for the case between named and unnamed providers.
But scenario 2 is not handled by the JS SDK now, I will fix that @toddbaert.

toddbaert:
> I totally agree. Also found that and implemented it in the JS SDK the way that this does not happen for the case between named and unnamed providers. But scenario 2 is not handled by the JS SDK now, I will fix that @toddbaert.

Let's wait for now to build some consensus, just so that you dont implement something that you have to change. Do you agree with the solution I've vaguely proposed?

lukas-reining:
[...]

Issue #189, opened 2023-05-24
[Question] `API.shutdown()` required in Go?


toddbaert:
[1.6.1](https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#requirement-161) requires an `API.shutdown()` function. After a few discussions with contributors I'm not 100% sure this is necessary, particularly for Go.

The purpose of the `API.shutdown()` function is for providing a means to initiate a graceful shutdown. For example, in Java, you'd catch a SIGINT in your Java app, and then that handler would call `API.Shutdown()` to cleanup all the polling, DB connections, workers, and flush telemetry or whatever other cleanup providers need done.

However in Go, the Context provides some of this functionality. A better pattern for Go _might_ be to create a provider with a NewMyProvider function (which the provider author defines) which takes `Context`. That provider listens for ctx.done() and handles their own shutdown. Application authors simply emit the cancellation event and everything is cleaned up.

We may want to change `1.6.1` to a "MAY" instead of a "MUST", and add some non-normative text like: 

```
Implementations can choose to leverage language idioms such as auto-disposable interfaces and other means of cancellation signal propagation to allow for graceful shutdown.
```

weyert:
I think that's reasonable

thomaspoignant:
Both solution are fine for me, but from the provider point of view it is always a bit tricky to have specificity for one language. I would prefer to have the same pattern everywhere.

If we want to go with the context approach I think we need to be extra cautious to make it explicit in the documentation how we expect providers to handle this.

Kavindu-Dodan:
I am also fine with language-specific implementations to handle the shutdown (context VS shutdown method). 

After going through the spec, I think we should keep the keyword `MUST` and redefine the requirement to align with providers. Example suggestion,

> The API MUST define a mechanism to propagate shutdown requests to active providers

Also, requirement's `...must call the respective shutdown function on the active provider` section mismatch with the provider shutdown requirement [1], which is uses keyword `MAY` (contradict with **must** call as the provider might skip this)

> The provider MAY define a shutdown function.... [1]
 
 ----
Altogether, if we agree on this, then we should correct both API & Provider spec sections to align. 



[1] - https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#25-shutdown

toddbaert:
> I am also fine with language-specific implementations to handle the shutdown (context VS shutdown method).
> 
[...]

PR #193, opened 2023-05-31
fix: various clarifications

toddbaert:
* loosen shutdown requirements to support things like go ctx https://github.com/open-feature/spec/issues/189
* clarify how to handle init/shutdown of providers bound to multiple names https://github.com/open-feature/spec/issues/192
* ~clarify init function should terminate before provider used https://github.com/open-feature/java-sdk/pull/456#discussion_r1210725372~
* clarify that authors can await the READY event during init https://github.com/open-feature/java-sdk/pull/456#discussion_r1210725372

cc @open-feature/technical-steering-committee @lopitz @lukas-reining @kinyoklion @Kavindu-Dodan 

Definitely let me know anyone thinks something is missing or could be improved here.

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -17,7 +17,7 @@
         {
             "id": "Requirement 1.1.2.2",
             "machine_id": "requirement_1_1_2_2",
-            "content": "The `provider mutator` function MUST await the termination of the `initialize` function on the newly registered provider before using it to resolve flag values.",
+            "content": "The `provider mutator` function MUST invoke the `initialize` function on the newly registered provider before using it to resolve flag values.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -17,7 +17,7 @@
         {
             "id": "Requirement 1.1.2.2",
             "machine_id": "requirement_1_1_2_2",
-            "content": "The `provider mutator` function MUST invoke the `initialize` function on the newly registered provider before using it to resolve flag values.",
+            "content": "The `provider mutator` function MUST await the termination `initialize` function on the newly registered provider before using it to resolve flag values.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -230,7 +230,7 @@
         {
             "id": "Requirement 1.6.1",
             "machine_id": "requirement_1_6_1",
-            "content": "The API MUST define a `shutdown` function which, when called, must call the respective `shutdown` function on the active provider.",
+            "content": "The API MUST define a mechanism to propagate a shutdown request to active providers.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -386,7 +386,7 @@
         {
             "id": "Requirement 2.5.1",
             "machine_id": "requirement_2_5_1",
-            "content": "The provider MAY define a `shutdown` function to perform whatever cleanup is necessary for the implementation.",
+            "content": "The provider MAY define a mechanism to gracefully shutdown and dispose of resources.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -17,7 +17,7 @@
         {
             "id": "Requirement 1.1.2.2",
             "machine_id": "requirement_1_1_2_2",
-            "content": "The `provider mutator` function MUST await the termination `initialize` function on the newly registered provider before using it to resolve flag values.",
+            "content": "The `provider mutator` function MUST await the termination of the `initialize` function on the newly registered provider before using it to resolve flag values.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -37,7 +37,7 @@ See [provider](./02-providers.md) for details.
 
 #### Requirement 1.1.2.2
 
-> The `provider mutator` function **MUST** await the termination `initialize` function on the newly registered provider before using it to resolve flag values.
+> The `provider mutator` function **MUST** await the termination of the `initialize` function on the newly registered provider before using it to resolve flag values.
 
 The `provider's` readiness can state can be determined from its `status` member/accessor.
 Provider instances which are already active (because they have been bound to other `names` or otherwise) need not be initialized again.
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -37,12 +37,13 @@ See [provider](./02-providers.md) for details.
 
 #### Requirement 1.1.2.2
 
-> The `provider mutator` function **MUST** await the termination of the `initialize` function on the newly registered provider before using it to resolve flag values.
+> The `provider mutator` function **MUST** invoke the `initialize` function on the newly registered provider before using it to resolve flag values.
 
-The `provider's` readiness can state can be determined from its `status` member/accessor.
+Application authors can await the newly set `provider's` readiness using the `PROVIDER_READY` event.
 Provider instances which are already active (because they have been bound to other `names` or otherwise) need not be initialized again.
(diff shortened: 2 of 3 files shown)


specification.json
```diff
@@ -17,7 +17,7 @@
         {
             "id": "Requirement 1.1.2.2",
             "machine_id": "requirement_1_1_2_2",
-            "content": "The `provider mutator` function MUST invoke the `initialize` function on the newly registered provider before using it to resolve flag values.",
+            "content": "The `provider mutator` function MUST await the termination `initialize` function on the newly registered provider before using it to resolve flag values.",
```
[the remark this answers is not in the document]
toddbaert: That's a typo. It should be:

"MUST await the termination (of the) `initialize` function"

It's funny how your brain just inserts words that it wants to read sometimes :sweat_smile:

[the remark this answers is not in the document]
toddbaert: Updated. I hope that helps.

kinyoklion (approved):

specification.json
```diff
@@ -17,7 +17,7 @@
         {
             "id": "Requirement 1.1.2.2",
             "machine_id": "requirement_1_1_2_2",
-            "content": "The `provider mutator` function MUST invoke the `initialize` function on the newly registered provider before using it to resolve flag values.",
+            "content": "The `provider mutator` function MUST await the termination of the `initialize` function on the newly registered provider before using it to resolve flag values.",
```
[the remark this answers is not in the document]
Kavindu-Dodan: I think this explanation fits well into non-normative section [1]

[1] - https://github.com/open-feature/spec/pull/193/files#diff-300118ed122b34e94bd16e8714f09eb91547e48b20e087d4b5ad2847698581d3R42-R43

Kavindu-Dodan (approved):
Approving - Open discussions can be finalized on agreements and personaly I don't see them to conflict a lot

specification.json
lukas-reining, line 20: What happens if the initialize fails here?
With the current spec, I still set it and only emitted the `PROVIDER_ERROR` event.
We specified that for transient errors, the provider might "recover" and change its status to `PROVIDER_READY`.
For an error on initialize that would indicate to me that we still set the provider.

To me it is not 100% clear why we have to await the init, as we have the mechanisms to wait for and check readiness anyways. It could also be confusing, setting a provider that only eventually is set and used and until that the old provider is still resolving the flags.

benjiro (approved):

specification.json
toddbaert: > With the current spec, I still set it and only emitted the PROVIDER_ERROR event.
We specified that for transient errors, the provider might "recover" and change its status to PROVIDER_READY.

Yes, I think this is the best implementation. The provider can recover as you say.

> To me it is not 100% clear why we have to await the init, as we have the mechanisms to wait for and check readiness anyways. It could also be confusing, setting a provider that only eventually is set and used and until that the old provider is still resolving the flags.

My goal with this is to avoid the situation where the newly set provider is used before ready when switched. Waiting for initialization means there's no period of unavailability. If a provider is switched and we don't wait for init, evaluations all start defaulting for some small period while the new provider is starting. Maybe that's something we're OK with?

Switching providers is _probably_ not something that we should consider a "first class" use case, so I'm open to simplifying things here.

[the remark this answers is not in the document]
toddbaert: @lopitz see the discussion [here](https://github.com/open-feature/spec/pull/193#discussion_r1212930484), I'd like your thoughts on it.

lukas-reining: >  If a provider is switched and we don't wait for init, evaluations all start defaulting for some small period while the new provider is starting. Maybe that's something we're OK with?

Yes, my feeling is that this would be better. But no hard opinion on that iI see pros and cons of both solutions.

liran2000: Question: wait for initialization to finish: 
is there a spec for getStatus() method in SDKs or something like that ?

[spec section](https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#requirement-242)

toddbaert: > Question: wait for initialization to finish: is there a spec for getStatus() method in SDKs or something like that ?
> 
> [spec section](https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#requirement-242)

Yes, that's basically what you've linked above. That will indicate the providers status.

kinyoklion: Generally speaking I think that we expect providers to be able to handle calls prior to their initialization, and the evaluation may result in some status indicating the provider wasn't ready, accompanied by a default value. Without initialization/shutdown it would have just been this way when switching providers.

I think, even with the init/shutdown additions, this is likely fine for a couple reasons.

If someone wants to switch providers, and they specifically want to use the first provider until the second provider is ready, then they can utilized named providers. Set the second provider under a different name, wait for it to be ready, and then either switch to using that name, or additionally alias that provider to the first name.

Even if a provider is not ready, you may want to be making calls to it, if those calls might be generating analytics events which are important to the operation of the specific vendor. 

I think there is potentially a nuance in provider switching that actually causes it to happen in most use cases. If the SDK has a no-op provider until a provider is set, then that is a provider switch?

toddbaert: > Generally speaking I think that we expect providers to be able to handle calls prior to their initialization, and the evaluation may result in some status indicating the provider wasn't ready, accompanied by a default value. Without initialization/shutdown it would have just been this way when switching providers.
> 
> I think, even with the init/shutdown additions, this is likely fine for a couple reasons.
> 
> If someone wants to switch providers, and they specifically want to use the first provider until the second provider is ready, then they can utilized named providers. Set the second provider under a different name, wait for it to be ready, and then either switch to using that name, or additionally alias that provider to the first name.
> 
> Even if a provider is not ready, you may want to be making calls to it, if those calls might be generating analytics events which are important to the operation of the specific vendor.

I think this is very well put, and I agree. It also removes the awkward "provider was set but is not in use yet" situation. I will make changes accordingly.

@lukas-reining see above.

> I think there is potentially a nuance in provider switching that actually causes it to happen in most use cases. If the SDK has a no-op provider until a provider is set, then that is a provider switch?

No, and I _think_ that's fairly easily justified: the noop-provider isn't a _real_ provider - it's an implementation detail - a easy way of "defaulting to no-op". If the noop provider was removed from any implementation, it wouldn't violate the spec, as long as the SDK continued to noop when evaluation methods were called.

Though the noop provider exists, it's functionally equivalent to "no provider".

toddbaert: > I think, the least surprise would be to return the default value. It's basically like a communication error with the feature provider. This of course comes with the risk, that features evaluated right in the beginning may always return the default value. In this case, I actually would think, that a ProviderInitialized event makes sense. This event would be emitted by the API, when the initialization is done and flag evaluation with the newly set provider can be started.

I think we generally agree not to await the initialization before setting the new provider then. With respect to the initialization event you describe, I think that the `PROVIDER_READY` event behaves in such a way. I think we have sufficient "levers and knobs" for authors with respect to initialization. I will revert the initialization point in this PR and add a bit more non-normative clarity.

toddbaert: I've revered the change wrt awaiting the init event, and added a tiny bit of non-normative content as well: https://github.com/open-feature/spec/pull/193/commits/9f5a08468eb413ca42677e4a83ba4f5076ffc2e7

lukas-reining (approved):
Looks good to me now and the changes are really good to me!

specification.json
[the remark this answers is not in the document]
toddbaert: I believe this is no longer relevant with the conclusion of https://github.com/open-feature/spec/pull/193#discussion_r1212930484

justinabrahms (approved):

s-sen:
The spec shows that the [initialization](https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#24-initialization) accepts a global context as a parameter. However the java sdk provider interface has the initialization API without any parameters. Is there any discussion regarding that?

PR #196, opened 2023-06-23
feat: STALE state, run handlers for current state immediately, provider name

toddbaert:
* Adds additional clarity regarding provider/client name mapping, as well as intent
* Replaces client-name for provider-name in events-details
* adds "STALE" state
* handlers for ANY state run immediately if provider is in that state

The only functional changes here are small ones related to events, and a new provider state, so particularly interested in @Kavindu-Dodan and @lukas-reining 's opinions on this one.

Relates to https://github.com/open-feature/spec/pull/200, which depends on `STALE` state.

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -31,7 +31,7 @@
         {
             "id": "Requirement 1.1.3",
             "machine_id": "requirement_1_1_3",
-            "content": "The `API` MUST provide a function to bind a given `provider` to a client `name`. If the client-name already has a bound provider, it is overwritten with the new mapping.",
+            "content": "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -31,7 +31,7 @@
         {
             "id": "Requirement 1.1.3",
             "machine_id": "requirement_1_1_3",
-            "content": "The `API` MUST provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.",
+            "content": "The `API` MUST provide a function to bind a given `provider` to a client `name`. If the client-name already has a bound provider, it is overwritten with the new mapping.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -796,7 +796,7 @@
         {
             "id": "Requirement 5.2.3",
             "machine_id": "requirement_5_2_3",
-            "content": "The `event details` MUST contain the `client name` associated with the event.",
+            "content": "The `event details` MUST contain the `provider name` associated with the event.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -845,7 +845,7 @@
         {
             "id": "Requirement 5.3.3",
             "machine_id": "requirement_5_3_3",
-            "content": "`PROVIDER_READY` handlers attached after the provider is already in a ready state MUST run immediately.",
+            "content": "Client `PROVIDER_READY` handlers attached after the provider is in a ready state MUST run immediately.",
             "RFC 2119 keyword": "MUST",
             "children": []
         }
--- a/specification.json
+++ b/specification.json
@@ -845,7 +845,7 @@
         {
             "id": "Requirement 5.3.3",
             "machine_id": "requirement_5_3_3",
-            "content": "`PROVIDER_READY` handlers attached after the provider is in a ready state MUST run immediately.",
+            "content": "Handlers attached after the provider is already in the associated state, MUST run immediately.",
             "RFC 2119 keyword": "MUST",
             "children": []
         }
--- a/specification.json
+++ b/specification.json
@@ -845,7 +845,7 @@
         {
             "id": "Requirement 5.3.3",
             "machine_id": "requirement_5_3_3",
-            "content": "Client `PROVIDER_READY` handlers attached after the provider is in a ready state MUST run immediately.",
+            "content": "`PROVIDER_READY` handlers attached after the provider is in a ready state MUST run immediately.",
             "RFC 2119 keyword": "MUST",
             "children": []
         }
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -98,7 +98,7 @@ See [provider](./02-providers.md) for details.
 
 ```java
 // example client creation and retrieval
-OpenFeature.getClient(name: "my-named-client");
+OpenFeature.getClient("my-named-client");
 ```
 
 The name is a logical identifier for the client which may be associated with a particular provider by the application integrator.
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -101,7 +101,7 @@ See [provider](./02-providers.md) for details.
 OpenFeature.getClient(name: "my-named-client");
 ```
 
-The name is a logical identifier for the client which may be associated with a particular provider.
+The name is a logical identifier for the client which may be associated with a particular provider by the application integrator.
 If a client name is not bound to a particular provider, the client is associated with the default provider.
 
 See [setting a provider](#setting-a-provider) for details.
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -56,7 +56,7 @@ see: [shutdown](./02-providers.md#25-shutdown), [setting a provider](#setting-a-
 
 #### Requirement 1.1.3
 
-> The `API` **MUST** provide a function to bind a given `provider` to a client `name`. If the client-name already has a bound provider, it is overwritten with the new mapping.
+> The `API` **MUST** provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.
 
 ```java
 OpenFeature.setProvider("client-name", new MyProvider());
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -31,9 +31,9 @@ It's important that multiple instances of the `API` not be active, so that state
 OpenFeature.setProvider(new MyProvider());
 ```
 
-This provider is used if there is not a more specific client name binding. (see later requirements).
+This provider is used if a client is not bound to a specific provider through its name.
 
-See [provider](./02-providers.md) for details.
+See [provider](./02-providers.md), [creating clients](#creating-clients).
 
 #### Requirement 1.1.2.2
 
@@ -56,11 +56,15 @@ see: [shutdown](./02-providers.md#25-shutdown), [setting a provider](#setting-a-
 
 #### Requirement 1.1.3
 
-> The `API` **MUST** provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.
+> The `API` **MUST** provide a function to bind a given `provider` to a client `name`. If the client-name already has a bound provider, it is overwritten with the new mapping.
 
 ```java
 OpenFeature.setProvider("client-name", new MyProvider());
 ```
+
+Named clients can be associated with a particular provider by supplying a matching name when the provider is set.
+
+See [creating clients](#creating-clients).
 
 #### Requirement 1.1.4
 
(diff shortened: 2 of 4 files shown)


specification/sections/05-events.md
```diff
@@ -80,10 +80,12 @@ see: [provider events](#51-provider-events), [`provider event types`](../types.m
...
 
-> The `event details` **MUST** contain the `client name` associated with the event.
+> The `event details` **MUST** contain the `provider name` associated with the event.
 
-The `client name` indicates the client/provider pair with which the event is associated.
-This is especially relevant for event handlers which are attached to the `API`, not a particular client.
+The `provider name` indicates the provider from which the event originated.
+This is especially relevant for global event handlers used for general monitoring, such as alerting on provider errors. 
+
+See [setting a provider](./01-flag-evaluation.md#setting-a-provider), [creating clients](./01-flag-evaluation.md#creating-clients).
```
toddbaert, line 88: The previous requirement was problematic because it's possible for the same provider instance to be bound to multiple names... which means that an event would sometimes need to feature mutliple `client names`. I think this would be overly confusing.

The main use-case for this information (as mentioned in the non-normative section) is troubleshooting and monitoring. For this reason, the provider name is probably a better choice.

Events added to clients themselves already provider a better means of scoping events to a client, so the `client name` wasn't helpful for this purpose.

lukas-reining: >  which means that an event would sometimes need to feature mutliple client names

The current implementation of JS emits the provider event through every client that it is bound to. If client 1 and client 2 both use provider 1, then client 1 and 2 will emit that event.

> Events added to clients themselves already provider a better means of scoping events to a client, so the client name wasn't helpful for this purpose.

Think so too, so to me it seems fine to include the provider name and not the client name in the event.

My only concern is, that for global events, knowing which client is affected only works by knowing which clients this provider is bound to.

specification.json
```diff
@@ -675,7 +675,7 @@
         {
             "id": "Requirement 5.2.3",
             "machine_id": "requirement_5_2_3",
-            "content": "The `event details` MUST contain the `client name` associated with the event.",
+            "content": "The `event details` MUST contain the `provider name` associated with the event.",
```
Kavindu-Dodan, line 678: This makes sense. Having this in handler callback helps with correlating events with the provider.

specification/sections/05-events.md
toddbaert: > The current implementation of JS emits the provider event through every client that it is bound to. If client 1 and client 2 both use provider 1, then client 1 and 2 will emit that event.

Yes, and that is right.

But for the corresponding global event, what happens in this case?

> My only concern is, that for global events, knowing which client is affected only works by knowing which clients this provider is bound to.

This is also true - but I think for global events, you're not so concerned with this. If you were, you'd have probably used a client event handler in the first place.

specification/sections/01-flag-evaluation.md
```diff
@@ -56,12 +56,16 @@ see: [shutdown](./02-providers.md#25-shutdown), [setting a provider](#setting-a-
 
 #### Requirement 1.1.3
 
-> The `API` **MUST** provide a function to bind a given `provider` to one or more client `name`s. If the client-name already has a bound provider, it is overwritten with the new mapping.
```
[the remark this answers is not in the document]
toddbaert: The existing wording seems to imply that you must implement a method that takes an array of client names when setting a provider (at least one person asked me about that).

specification.json
```diff
@@ -724,7 +724,7 @@
         {
             "id": "Requirement 5.3.3",
             "machine_id": "requirement_5_3_3",
-            "content": "`PROVIDER_READY` handlers attached after the provider is already in a ready state MUST run immediately.",
+            "content": "Client `PROVIDER_READY` handlers attached after the provider is in a ready state MUST run immediately.",
```
Kavindu-Dodan, line 727: Any reason to bind this to `Client`? 

Note - implementing this requirement require event handler implementation to track provider state change OR have reference to providers and check their status when handlers get registered.

specification/sections/01-flag-evaluation.md
```diff
@@ -56,12 +56,16 @@ see: [shutdown](./02-providers.md#25-shutdown), [setting a provider](#setting-a-
...
+> The `API` **MUST** provide a function to bind a given `provider` to a client `name`. If the client-name already has a bound provider, it is overwritten with the new mapping.
 
 ```java
 OpenFeature.setProvider("client-name", new MyProvider());
 ```
 
+Named clients can be associated with a particular provider by supplying a matching name when the provider is set.
```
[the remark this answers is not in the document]
beeme1mr: Nevermind, this is talking about associating an existing client with a newly registered named provider.

specification/sections/05-events.md
lukas-reining: > But for the corresponding global event, what happens in this case? Either you must fire 2 global events (that seems wrong) or the event should have both client names (which is the problem I was trying to resolve here).

Okay yes that makes sense to me.

> But for the corresponding global event, what happens in this case? Either you must fire 2 global events (that seems wrong) or the event should have both client names (which is the problem I was trying to resolve here).

I think that depends on the view. If the events would be "client events" I would call this right. If the events are "provider events" as we called them I am with you, and you should only get the event once globally.

> This is also true - but I think for global events, you're not so concerned with this. If you were, you'd have probably used a client event handler in the first place.

Makes sense to me :)

specification.json
toddbaert: > Any reason to bind this to Client?

Yes - I added some justification for this in the non-normative section:

>_Application authors_ may attach readiness handlers to be confident that system is ready to evaluate flags.
If such handlers are attached after the provider underlying the client has already been initialized, they should run immediately.
**_API (global) handlers have no such requirement, as they are potentially bound to multiple providers and are used primarily for diagnostic purposes._**

> Note - implementing this requirement require event handler implementation to track provider state change OR have reference to providers and check their status when handlers get registered.

Yes - but just to be clear - this was also the case before this change, right?

specification/sections/05-events.md
Kavindu-Dodan: > > My only concern is, that for global events, knowing which client is affected only works by knowing which clients this provider is bound to.
> 
> This is also true - but I think for global events, you're not so concerned with this. If you were, you'd have probably used a client event handler in the first place.

Related to [1]

To fully support named client to default provider binding as well as invoking ready handlers when registering require event handling mechanism to know about provider status 

[1] - https://github.com/open-feature/spec/pull/196/files#r1239934159

specification.json
Kavindu-Dodan: This is something I missed in go-sdk implementation. But it's connected to tracking providers through event handler mechanism 

And non-normative text expansion helps here to give a clear idea behind this. Hence ➕  for this

specification/sections/05-events.md
toddbaert: >  require event handling mechanism to know about provider status

> While it's fun to implement Eventing spec, implementation details of this is a good example for devil is in the details idiom smile

:100: . Separation of concerns is a real challenge here. In Java, I passed around a lot of callbacks to help with much of that - so that I could maintain the exact details of some behavior to the caller, but the called object would run that delegated behavior at the right time.

The [providerRepository](https://github.com/open-feature/java-sdk/blob/cbf5c9fb581db1997450725fe6e1655a2f065dfa/src/main/java/dev/openfeature/sdk/ProviderRepository.java) for example takes callbacks that it runs on provider ready/error/shutdown etc. It has no idea what is happening within those functions (they mostly emit events, etc).

specification/sections/01-flag-evaluation.md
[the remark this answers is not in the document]
Kavindu-Dodan: > The `API` **MUST** provide a function to bind a given `provider` to one or more client `name`s. 

I assume the original intension here is to allow the following (syntax Java)

```Java
Provider myProvider = new Provider();

API.setProvider("client-A", myProvider);
API.setProvider("client-B", myProvider);
...
```

While implementing the eventing spec requirements, I found this multi-client bound requirement to be too complex to fulfill with decent logic. 

I think we should have some restrictions to make the spec less complex (reader as well as sdk maintainers). But also fulfil valid requirements like provider to client binding through names.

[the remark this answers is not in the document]
toddbaert: > While implementing the eventing spec requirements, I found this multi-client bound requirement to be too complex to fulfill with decent logic.
>
> I think we should have some restrictions to make the spec less complex (reader as well as sdk maintainers). But also fulfil valid requirements like provider-to-client binding through a name.

What did you have in mind?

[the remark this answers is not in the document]
Kavindu-Dodan: > > While implementing the eventing spec requirements, I found this multi-client bound requirement to be too complex to fulfill with decent logic.
> > I think we should have some restrictions to make the spec less complex (reader as well as sdk maintainers). But also fulfil valid requirements like provider-to-client binding through a name.
> 
> What did you have in mind?

I had the impression that a single provider can be associated with multiple clients. Here a single provider means the same instance as in the example given below,

```java
Provider myProvider = new Provider();

API.setProvider("client-A", myProvider);
API.setProvider("client-B", myProvider);
...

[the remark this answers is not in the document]
Comment (toddbaert) [reply specification/sections/01-flag-evaluation.md:59]: I see, OK I think I'll revert this change.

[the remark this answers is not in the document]
Comment (toddbaert) [reply specification/sections/01-flag-evaluation.md]: Reverted

File: specification.json (hunk 4 above)
Comment (Kavindu-Dodan) [reply specification.json:727]: Revisiting this again after the go-sdk implementation. I doubt whether this needs to be restricted to the client (scoped events)

Assume I only register a global provider. In my application, I add this in a startup logic. Alongside, I register an API-level handler for ready state. By the time execution reach handler registration, the provider could be in a ready state (especially if initialization is nonblocking and happens concurrently). If we enforce this spec change, then my global ready handler never gets executed. 

Besides, we want to consider adding another rule `Requirement 5.3.4 - PROVIDER_ERROR handlers attached after the provider is in a ready state MUST run immediately` 

Reason is, these events are connected to provider initialization. And initialization can yield `PROVIDER_READY` & `PROVIDER_ERROR` which IMO needs special handling.

**Review by beeme1mr — APPROVED** [review]

File: specification.json (hunk 2 above)
Comment (Kavindu-Dodan) [reply specification.json:678]: resolving as done

File: specification.json (hunk 4 above)
Comment (toddbaert) [reply specification.json:727]: I agree with you on both counts. I think I will make this change.

File: specification/sections/05-events.md [hunk 6]
```diff
@@ -134,6 +136,10 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
 
 #### Requirement 5.3.3
 
-> `PROVIDER_READY` handlers attached after the provider is already in a ready state **MUST** run immediately.
+> Client `PROVIDER_READY` handlers attached after the provider is in a ready state **MUST** run immediately.
```
toddbaert, line 139:
```suggestion
> `PROVIDER_READY` handlers attached after the provider is in a ready state **MUST** run immediately.
```

specification/sections/05-events.md
```diff
@@ -134,6 +136,10 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
 
 #### Requirement 5.3.3
 
-> `PROVIDER_READY` handlers attached after the provider is already in a ready state **MUST** run immediately.
+> `PROVIDER_READY` handlers attached after the provider is in a ready state **MUST** run immediately.
 
-See [provider initialization](./02-providers.md#24-initialization) and [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+_Application authors_ may attach readiness handlers to be confident that system is ready to evaluate flags.
+If such handlers are attached after the provider underlying the client has already been initialized, they should run immediately.
+API (global) handlers have no such requirement, as they are potentially bound to multiple providers and are used primarily for diagnostic purposes.
```
toddbaert, line 143:
```suggestion
```

specification/sections/05-events.md
```diff
@@ -134,6 +136,9 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
 
 #### Requirement 5.3.3
 
-> `PROVIDER_READY` handlers attached after the provider is already in a ready state **MUST** run immediately.
+> `PROVIDER_READY` handlers attached after the provider is in a ready state **MUST** run immediately.
```
toddbaert, line 139: @thomaspoignant I think you recently faced some challenges here...

As @Kavindu-Dodan pointed out elsewhere, this point should _probably_ be extended to apply to `STALE` and `ERROR` as well - basically, we don't want race conditions between the provider state and the addition of handlers. I was thinking only of the "happy" path when I originally added this. :sweat_smile:

toddbaert: If

Kavindu-Dodan: I support this change. Handler registration can occur in a later stage of an application (ex:- setup phase for provider init and handlers per client in a request scope), so it makes sense to propagate states to new handlers.

toddbaert:
@Kavindu-Dodan @lukas-reining @beeme1mr @Kavindu-Dodan I've updated this PR. Please re-review.

The main change is the addition of the `STALE` provider state, and changes to `5.3.3`: instead of just `READY` events running immediately, now the spec dictates that handlers of any type should run immediately if they are added when the provider is already in the associated state.

lukas-reining (approved):
Looks good to me :)

beeme1mr (approved):

Kavindu-Dodan (approved):

PR #200, opened 2023-07-21
feat: emit events on context change (client-only)

Billlynch:
This PR adds points to specify the behaviour of on context changed with respect to eventing.

- (optionally) emit `PROVIDER_STALE` when on context changed is executed with mismatching contexts
- emitting `PROVIDER_CONTEXT_CHANGED` when on context changed successfully terminates
- emitting `PROVIDER_ERROR` when on context changed terminates abnormally

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -891,6 +891,27 @@
             "content": "Handlers attached after the provider is already in the associated state, MUST run immediately.",
             "RFC 2119 keyword": "MUST",
             "children": []
+        },
+        {
+            "id": "Requirement 5.3.4",
+            "machine_id": "requirement_5_3_4",
+            "content": "When the provider's `on context changed` is called, `PROVIDER_STALE` handlers MUST run.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 5.3.5",
+            "machine_id": "requirement_5_3_5",
+            "content": "If the provider's `on context changed` function terminates successfully, `PROVIDER_READY` handlers MUST run.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 5.3.6",
+            "machine_id": "requirement_5_3_6",
+            "content": "If the provider's `on context changed` function terminates abnormally, `PROVIDER_ERROR` handlers MUST run.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
         }
     ]
 }
--- a/specification.json
+++ b/specification.json
@@ -893,25 +893,33 @@
             "children": []
         },
         {
-            "id": "Requirement 5.3.4",
-            "machine_id": "requirement_5_3_4",
-            "content": "When the provider's `on context changed` is called, `PROVIDER_STALE` handlers MUST run.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 5.3.5",
-            "machine_id": "requirement_5_3_5",
-            "content": "If the provider's `on context changed` function terminates successfully, `PROVIDER_READY` handlers MUST run.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 5.3.6",
-            "machine_id": "requirement_5_3_6",
-            "content": "If the provider's `on context changed` function terminates abnormally, `PROVIDER_ERROR` handlers MUST run.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
+            "id": "Condition 5.3.4",
+            "machine_id": "condition_5_3_4",
+            "content": "The implementation uses the static-context paradigm.",
+            "RFC 2119 keyword": null,
+            "children": [
+                {
+                    "id": "Conditional Requirement 5.3.4.1",
+                    "machine_id": "conditional_requirement_5_3_4_1",
+                    "content": "When the provider's `on context changed` is called, the provider MAY emit the `PROVIDER_STALE` event, and transition to the `STALE` state.",
+                    "RFC 2119 keyword": "MAY",
+                    "children": []
+                },
+                {
+                    "id": "Conditional Requirement 5.3.4.2",
+                    "machine_id": "conditional_requirement_5_3_4_2",
+                    "content": "If the provider's `on context changed` function terminates normally, associated `PROVIDER_CONTEXT_CHANGED` handlers MUST run.",
+                    "RFC 2119 keyword": "MUST",
+                    "children": []
+                },
+                {
+                    "id": "Conditional Requirement 5.3.4.3",
+                    "machine_id": "conditional_requirement_5_3_4_3",
+                    "content": "If the provider's `on context changed` function terminates abnormally, associated `PROVIDER_ERROR` handlers MUST run.",
+                    "RFC 2119 keyword": "MUST",
+                    "children": []
+                }
+            ]
         }
     ]
 }
--- a/specification/sections/05-events.md
+++ b/specification/sections/05-events.md
@@ -158,8 +158,7 @@ title: Provider context reconciliation
 stateDiagram-v2
     direction TB
     READY --> READY:emit(PROVIDER_CONTEXT_CHANGED)
-    READY --> ERROR
-    ERROR --> READY
+    ERROR --> READY:emit(PROVIDER_READY)
     READY --> STALE:emit(PROVIDER_STALE)
     STALE --> READY:emit(PROVIDER_CONTEXT_CHANGED)
     STALE --> ERROR:emit(PROVIDER_ERROR)
--- a/specification/sections/05-events.md
+++ b/specification/sections/05-events.md
@@ -145,20 +145,56 @@ If such handlers are attached after the provider underlying the client has alrea
 
 See [provider initialization](./02-providers.md#24-initialization), [setting a provider](./01-flag-evaluation.md#setting-a-provider).
 
-#### Requirement 5.3.4
+### Event handlers and context reconciliation
 
-> When the provider's `on context changed` is called, `PROVIDER_STALE` handlers **MUST** run.
+Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
+This can be particularly useful for triggering UI repaints in multiple components when one component updates the [evaluation context](./03-evaluation-context.md).
+Optionally, some providers may transition to a the `STALE` state while their associated context is waiting to be reconciled, since this may involve asynchronous operations such as network calls.
 
-See: [provider events](#51-provider-events), [`provider event types`](../types.md#provider-events)
+```mermaid
+---
+title: Provider context reconciliation 
+---
+stateDiagram-v2
+    direction TB
+    READY --> READY:emit(PROVIDER_CONTEXT_CHANGED)
+    READY --> ERROR
+    ERROR --> READY
+    READY --> STALE:emit(PROVIDER_STALE)
+    STALE --> READY:emit(PROVIDER_CONTEXT_CHANGED)
+    STALE --> ERROR:emit(PROVIDER_ERROR)
+```
 
-#### Requirement 5.3.5
+#### Condition 5.3.4
 
-> If the provider's `on context changed` function terminates successfully, `PROVIDER_READY` handlers **MUST** run.
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
 
-See: [provider events](#51-provider-events), [`provider event types`](../types.md#provider-events)
+> The implementation uses the static-context paradigm.
 
-#### Requirement 5.3.6
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
 
-> If the provider's `on context changed` function terminates abnormally, `PROVIDER_ERROR` handlers **MUST** run.
+##### Conditional Requirement 5.3.4.1
 
-See: [provider events](#51-provider-events), [`provider event types`](../types.md#provider-events)
+> When the provider's `on context changed` is called, the provider **MAY** emit the `PROVIDER_STALE` event, and transition to the `STALE` state.
+
+Some providers cache evaluated flags, and re-evaluate them when the context is changed.
+In these cases, the provider may signal its cache is invalid with the `PROVIDER_STALE` event and the `STALE` provider state.
+
+see: [provider event types](../types.md#provider-events), [provider events](#51-provider-events), context, [provider context reconciliation](02-providers.md#26-provider-context-reconciliation)
+
+##### Conditional Requirement 5.3.4.2
+
+> If the provider's `on context changed` function terminates normally, associated `PROVIDER_CONTEXT_CHANGED` handlers **MUST** run.
+
+The implementation must run any `PROVIDER_CONTEXT_CHANGED` handlers associated with the provider after the provider has reconciled its state and returned from the `on context changed` function.
+The `PROVIDER_CONTEXT_CHANGED` is not emitted from the provider itself; the SDK implementation must run the `PROVIDER_CONTEXT_CHANGED` handlers if the `on context changed` function terminates normally.
+
+see: [provider event types](../types.md#provider-events), [provider events](#51-provider-events), context, [provider context reconciliation](02-providers.md#26-provider-context-reconciliation)
+
+##### Conditional Requirement 5.3.4.3
+
+> If the provider's `on context changed` function terminates abnormally, associated `PROVIDER_ERROR` handlers **MUST** run.
+
+The `PROVIDER_ERROR` is not emitted from the provider itself; the SDK implementation must run the `PROVIDER_ERROR` handlers if the `on context changed` throws or otherwise signals an error.
+
+see: [provider event types](../types.md#provider-events), [provider events](#51-provider-events), context, [provider context reconciliation](02-providers.md#26-provider-context-reconciliation)
--- a/specification/sections/05-events.md
+++ b/specification/sections/05-events.md
@@ -149,6 +149,7 @@ See [provider initialization](./02-providers.md#24-initialization), [setting a p
 
 Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
 This can be particularly useful for triggering UI repaints in multiple components when one component updates the [evaluation context](./03-evaluation-context.md).
+SDK implementations automatically fire the the `PROVIDER_CONTEXT_CHANGED` events if the `on context changed` handler terminates normally (and `PROVIDER_ERROR` events otherwise).
 Optionally, some providers may transition to a the `STALE` state while their associated context is waiting to be reconciled, since this may involve asynchronous operations such as network calls.
 
 ```mermaid
--- a/specification/sections/05-events.md
+++ b/specification/sections/05-events.md
@@ -150,7 +150,7 @@ See [provider initialization](./02-providers.md#24-initialization), [setting a p
 Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
 This can be particularly useful for triggering UI repaints in multiple components when one component updates the [evaluation context](./03-evaluation-context.md).
 SDK implementations automatically fire the the `PROVIDER_CONTEXT_CHANGED` events if the `on context changed` handler terminates normally (and `PROVIDER_ERROR` events otherwise).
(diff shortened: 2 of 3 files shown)


Billlynch:
I wonder if we should also add into the provider name into the provider event details?

If you have multiple providers and you change the context. One provider may take more time than the other to emit the ready event. If you've subscribed to the event via the `OpenFeature.addHandler()` function instead of on the provider/client itself you don't know if the provider you're about to evaluate against is ready or not.

toddbaert:
> I wonder if we should also add the provider name into the provider event details?

> If you have multiple providers and you change the context. One provider may take more time than the other to emit the ready event. If you've subscribed to the event via the `OpenFeature.addHandler()` function instead of on the provider/client itself you don't know if the provider you're about to evaluate against is ready or not.

Agreed.

This is one of the changes proposed [here](https://github.com/open-feature/spec/pull/196).

Something worth mentioning is one of the design principles of the SDKs is for [application authors](https://openfeature.dev/specification/glossary#application-author) (the persona most concerned with actually evaluating flags) not to worry about providers at all. Ideally, they just use a client, which is bound to a provider that the [integrator](https://openfeature.dev/specification/glossary#application-integrator) configured. We'd expect event handlers added to the API to be used most for telemetry, error-reporting, etc. We'd discourage application authors to add global handlers, check if their event is associated with "their" provider, and then take some action. Using a client event handler would be a better solution for that.

specification/sections/05-events.md
```diff
@@ -137,3 +137,21 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
...
+
+> If the provider's `on context changed` function successfully updates the cache, `PROVIDER_READY` handlers **MUST** run.
+
+See: [provider events](#51-provider-events), [`provider event types`](../types.md#provider-events)
+
+#### Requirement 5.3.6
+
+> If the provider's `on context changed` function terminates abnormally, `PROVIDER_ERROR` handlers **MUST** run.
+
+See: [provider events](#51-provider-events), [`provider event types`](../types.md#provider-events)
```
toddbaert, line 157: I like this idea - I think that it could prove useful.

I think one thing I'm unsure about is how to implement `5.3.5`. It would be nice to have these events emitted by the SDK itself, but then I think to satisfy `5.3.5` we would have to _define_ what constitutes a changed context (perhaps a hash of the whole object?). So far, we haven't done this - it's up for the provider to decide what properties they care about on the context, and to decide to take action or not.

I'm hesitant to do this... but maybe it's a good idea for the client?

The other option would be to be less prescriptive here, and simple recommend that provided emit these events themselves when the context has been changed according to their definition.

specification/sections/05-events.md
```diff
@@ -137,3 +137,21 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
...
+
+#### Requirement 5.3.4
+
+> If the provider's `on context changed` executes with mismatched contexts, `PROVIDER_STALE` handlers **MUST** run.
+
+See: [provider events](#51-provider-events), [`provider event types`](../types.md#provider-events)
+
+#### Requirement 5.3.5
+
+> If the provider's `on context changed` function successfully updates the cache, `PROVIDER_READY` handlers **MUST** run.
```
beeme1mr, line 149: Could you explain the use case you're solving with requirement `5.3.5`? I'm worried this could lead to many `PROVIDER_READY` events and overload the meaning behind the event. Would it be helpful to add a new event that could be emitted when the cache updates?

lukas-reining: I also like the change, and we just had the case in a client project where this would have been helpful.
But I am with you @beeme1mr, maybe it should be another event. Maybe even CONTEXT_CHANGED or so?

specification/sections/05-events.md
lukas-reining: > I think one thing I'm unsure about is how to implement 5.3.5. It would be nice to have these events emitted by the SDK itself, but then I think to satisfy 5.3.5 we would have to define what constitutes a changed context (perhaps a hash of the whole object?). So far, we haven't done this - it's up for the provider to decide what properties they care about on the context, and to decide to take action or not.
I'm hesitant to do this... but maybe it's a good idea for the client?

@toddbaert I am not sure if I get it right but why do we have to be sure that the new context actually is different different influences the flags? To me it fells good enough to say, that a PROVIDER_READY or maybe CONTEXT_CHANGED or so event is dispatched when the context is set and `on context changed` terminated.

lukas-reining:
>#### Requirement 5.3.4
>If the provider's `on context changed` executes with mismatched contexts, `PROVIDER_STALE` handlers **MUST** run.

What does "mismatched context" mean in this case? Every context that is different from the current one? We would emit this event then, and because of 5.3.5 we would get a `PROVIDER_READY` directly after it right like the diagram shows?

```mermaid
sequenceDiagram
    participant C as Client
    participant P as Provider
    participant E as EventEmitter

    C-->>P: Set context
    note right of C: This context is mismatching
    P-->>E: Dispatch PROVIDER_STALE
    P->>P: run onContextChanged
    P-->>E: Dispatch PROVIDER_READY

```

specification.json
```diff
@@ -848,6 +848,27 @@
             "content": "`PROVIDER_READY` handlers attached after the provider is already in a ready state MUST run immediately.",
             "RFC 2119 keyword": "MUST",
             "children": []
+        },
+        {
+            "id": "Requirement 5.3.4",
+            "machine_id": "requirement_5_3_4",
+            "content": "If the provider's `on context changed` executes with mismatched contexts, `PROVIDER_STALE` handlers MUST run.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
```
lukas-reining, line 857: >#### Requirement 5.3.4
>If the provider's `on context changed` executes with mismatched contexts, `PROVIDER_STALE` handlers **MUST** run.

What does "mismatched context" mean in this case? Every context that is different from the current one? We would emit this event then, and because of 5.3.5 we would get a `PROVIDER_READY` directly after it right like the diagram shows?

```mermaid
sequenceDiagram
    participant C as Client
    participant P as Provider
    participant E as EventEmitter

    C-)P: Set context
    note right of C: This context is mismatching
    P-)E: Dispatch PROVIDER_STALE
    P->>P: run onContextChanged
    P-)E: Dispatch PROVIDER_READY
```

thomaspoignant: I have the same question as @lukas-reining I don't really understand what "mismatched context" means.

specification/sections/05-events.md
thomaspoignant: I like the idea of having another event called `CONTEXT_CHANGED`.

beeme1mr: Would the provider developer be responsible for emitting the `CONTEXT_CHANGED` event? I think it's worth mentioning that it would only apply to the static context paradigm (i.e. client-side) SDKs.

toddbaert: > Would the provider developer be responsible for emitting the `CONTEXT_CHANGED` event? I think it's worth mentioning that it would only apply to the static context paradigm (i.e. client-side) SDKs.

Yep, we'd have to be very clear in the spec this event only applies to to the client.

specification/sections/05-events.md
toddbaert: I think this is a key point, and relates to [this](https://github.com/open-feature/spec/pull/200#discussion_r1275464288).

The language specifically says "mismatched context" which I think means "the context is different from it's previous value". Like I said before, I think this is my biggest concern with this proposal. I agree that it would be easier to simply ignore if the context is different, and run the even regardless. In the client scenario, we expect authors to be mindful of setting the global context... so I think we should just fire the event every time the context is set, regardless of whether it's "different" or not.

specification.json
toddbaert: https://github.com/open-feature/spec/pull/200#discussion_r1280801308

beeme1mr:
Hey @Billlynch, thanks for the PR. When you have a moment, could you please respond to the open threads? It would be great to get some clarification so we can work through this spec enhancement. Thanks!

specification/sections/05-events.md
lukas-reining: Yeah, that sounds good to me @toddbaert.

specification/sections/05-events.md
Billlynch: Agreed, this only makes sense in the static paradigm.

@beeme1mr The original problem I was trying to solve with this was thinking about frontend frameworks, mainly React. So the thought is that an integration might want to trigger a refresh to the page If the flag has changed. If you have the page already rendered, and then you login (or set the context in a button callback for whatever reason) we can emit the STALE event to tell any component trying to read the flags to show a spinner until such times as it has finished (using React.Suspense for example). Then when the ready event is fired we can tell those components to load and no longer show a spinner.

I don't think this is an overload of the `PROVIDER_READY` event, because that event's definition is that the provider is ready for flag evaluations, and if the context change has triggered a refetch of the flags and that fetch hasn't completed yet then it isn't ready for evaluations as we know the cached flags are stale.

specification/sections/05-events.md
toddbaert: > Yeah, I think emitting STALE and READY again even if the context is not mismatched works!

OK great. I think that satisfies this question for me, and we have a consensus. I recommend you update the proposal.

specification/sections/05-events.md
toddbaert: This came up in another context recently... I was helping somebody develop a provider that also goes into a stale state (it downloads a ruleset locally) and when it lost connection with the server, that ruleset could be stale. It was clear the provider should emit `STALE` in this case, but it's a bit unclear what to emit when the provider reconnects and becomes "unstale". This was in a "server-side" context.

If we decide transitioning out of staleness is a valid use of the `READY` event, we should more rigorously specify that in our glossary and in-line docs.

One small _potential_ mismatch here is that provider states include only `NOT_READY` (need to have their `initialize()` called) `READY` (already initialized, or not in need or initialization) or `ERROR` (something went wrong during initialize). See https://openfeature.dev/specification/types#provider-status.

I'm not sure if we should actually add a `STALE` _state_, but we might want to if there's any behavioral difference between a stale/ready provider from a usage perspective. Maybe not though, since I don't think it's relevant to the life-cycle of the provider (init/shutdown).

cc @bacherfl

Billlynch: +1 on the STALE _state_ I think the matching of the state and events is a good concept!

specification/sections/05-events.md
Kavindu-Dodan: Does this means context change (regardless of matching/mismatching ) makes SDK to emit `STALE` status ?

Kavindu-Dodan:
Given that we have considerable state changes, I would like to see an updated visualization of the state changes of a provider.

I see this fits well with our current state visualization here [^1]

In general, my understanding is that the Provider can go from `ready` state to `stale` and back to `ready` state once remediation for context change is done. Further, we have error status that can be reached from `ready` status as well as from `stale` status 

Context change: Ready -> Stale -> Ready 
Context change: Ready -> Stale -> Error .... -> Ready 

[^1]: https://github.com/open-feature/spec/blob/main/specification/sections/02-providers.md#requirement-242

specification.json
```diff
@@ -848,6 +848,27 @@
             "content": "`PROVIDER_READY` handlers attached after the provider is already in a ready state MUST run immediately.",
             "RFC 2119 keyword": "MUST",
             "children": []
+        },
+        {
+            "id": "Requirement 5.3.4",
+            "machine_id": "requirement_5_3_4",
+            "content": "When the provider's `on context changed`  is called, `PROVIDER_STALE` handlers MUST run.",
```
Kavindu-Dodan, line 855:
```suggestion
            "content": "When the provider's `on context changed` is called, `PROVIDER_STALE` handlers MUST run.",
```

specification/sections/05-events.md
Kavindu-Dodan: My concern is resolved with the latest commit - https://github.com/open-feature/spec/pull/200/commits/8d773c0ef61d4bfe82db1f76d13537f3f8b4ade4

specification.json
Kavindu-Dodan: @lukas-reining yes, this is the agreement according to the latest commit - https://github.com/open-feature/spec/pull/200/commits/8d773c0ef61d4bfe82db1f76d13537f3f8b4ade4 . We no longer mention `mismatched contexts` and events should fire for any context change.

lukas-reining: Great🙂

specification.json
beeme1mr: @Kavindu-Dodan this is autogenerated, so please also update the source.

specification/sections/05-events.md
```diff
@@ -137,3 +137,21 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
 > `PROVIDER_READY` handlers attached after the provider is already in a ready state **MUST** run immediately.
 
 See [provider initialization](./02-providers.md#24-initialization) and [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+
+#### Requirement 5.3.4
+
+> When the provider's `on context changed`  is called, `PROVIDER_STALE` handlers **MUST** run.
```
beeme1mr, line 143:
```suggestion
> When the provider's `on context changed` is called, `PROVIDER_STALE` handlers **MUST** run.
```

specification/sections/05-events.md
```diff
@@ -137,3 +137,21 @@ See [provider initialization](./02-providers.md#24-initialization) and [setting
 > `PROVIDER_READY` handlers attached after the provider is already in a ready state **MUST** run immediately.
 
 See [provider initialization](./02-providers.md#24-initialization) and [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+
+#### Requirement 5.3.4
+
+> When the provider's `on context changed` is called, `PROVIDER_STALE` handlers **MUST** run.
```
beeme1mr, line 143: Please update the `Provider Events` table to include the new event.

https://github.com/open-feature/spec/blob/main/specification/types.md#provider-events

specification/sections/05-events.md
beeme1mr: Should this be included in the current PR?

specification/sections/05-events.md
beeme1mr: Whoops, you're right!

specification/sections/05-events.md
beeme1mr: This is my only open question. Should the provider's state be updated to "STALE" while running the `on context changed` method?

toddbaert:
I'm struggling with one corollary to this.

I think we need to add a STALE state to keep things consistent (that's addressed here). But I'm a bit torn on exactly how we want the provider state to be managed. At the moment, providers' state is mutable only from within, it's readonly from the outside. If we implement this proposal as things stand, we'd expect authors to set the state to STALE while their provider's onContextChanged is run. This seems to be incongruent to the fact that the SDK will be emitting the STALE event on the provider's behalf.

I feel like it should be all-or-nothing to be consistent - either the SDK should fire these events and manage the state itself, or the provider should be responsible for both.

I'm wondering if it wouldn't be better to have the state mutable from the outside, and have the SDK set it after init/error/onContextChanged. This would reduce the burden on providers (though we might have to be careful to make sure the change is as "non-breaking" as possible.

Thoughts?

specification/sections/05-events.md
toddbaert: related: https://github.com/open-feature/spec/pull/200#issuecomment-1675221377

Billlynch:
I see your point.

I think a nice clear thing would be for the providers to emit the event `PROVIDER_STALE` and set the state to a new `STALE` state.

This then can actually work a bit more like the original proposal of only emitting if the provider decides the old and new contexts are actually different. 

So the state of the provider can be:
```
1. NOT_READY -> READY
2. NOT_READY -> ERROR
3. ERROR -> READY
4. READY -> STALE
5. STALE -> READY
6. STALE -> ERROR
```

with corresponding events being fired:
```
1. PROVIDER_READY
2. PROVIDER_ERROR
3. PROVIDER_READY
4. PROVIDER_STALE
5. PROVIDER_READY
6. PROVIDER_ERROR
```

This aligns the setting of the status of the provider, and emission of the events under the same logic, and the same domain as that which will actually make a cache update if needed. 

What do you think?

toddbaert:
> I see your point.
> 
> I think a nice clear thing would be for the providers to emit the event `PROVIDER_STALE` and set the state to a new `STALE` state.
> 
> This then can actually work a bit more like the original proposal of only emitting if the provider decides the old and new contexts are actually different.
> 
> So the state of the provider can be:
> 
> ```
> 1. NOT_READY -> READY
> 2. NOT_READY -> ERROR
> 3. ERROR -> READY
> 4. READY -> STALE
> 5. STALE -> READY
> 6. STALE -> ERROR
> ```
> 
> with corresponding events being fired:
> 
> ```
> 1. PROVIDER_READY
> 2. PROVIDER_ERROR
> 3. PROVIDER_READY
> 4. PROVIDER_STALE
> 5. PROVIDER_READY
> 6. PROVIDER_ERROR
> ```
> 
> This aligns the setting of the status of the provider, and emission of the events under the same logic, and the same domain as that which will actually make a cache update if needed.
> 
> What do you think?

I think this is a coherent proposal. We push a a lot of responsibility onto the provider, but I think that's a _safer_ move. On the SDK side, we would need to do our best to rigorously document the purpose of _STALE_ and that is should be used when context is mutated in client implementations.

It would also be useful in server implementations that cache rulesets or flag values, which is a nice in that it maintains semantic consistency between both SDKs.

toddbaert:
@Billlynch I've invited you to the org. Please comment [here](https://github.com/open-feature/community/pull/239) if you'd like to join (no obligation, but this will allow you to be pinged in PRs, issues, etc).

beeme1mr:
Hey @Billlynch, could you please update this proposal based on [this comment](https://github.com/open-feature/spec/pull/200#issuecomment-1677903240)? That should be the last remaining issue because it's approved and merged. 👍

lukas-reining (approved):

toddbaert:
Hey @Billlynch I'm hoping to release a spec v0.7.0 this week. If you have the time to update this PR we can include this content in the release - otherwise we can do another after this is merged.

toddbaert:
@Billlynch 

I've come back to this PR after a lot of thought based on my experience in the React SDK. I have squashed your commits into one, [and then added my own](https://github.com/open-feature/spec/pull/200/commits/342af54b117dd42d0d640b9dd29c3e28abe5958c), which makes the `STALE` transition optional (some providers might cache local rulesets and never go `STALE`, so I think we can leave it up to them) and uses a new `PROVIDER_CONTEXT_CHANGED` event.

I think this new event is needed to specifically differentiate from `READY` events (which don't indicate any flags changed) and `PROVIDER_CONFIGURATION_CHANGED` which specifically imply the change was propagated from the management system. `PROVIDER_CONTEXT_CHANGED` are unique to the client-side, and particularly relevant for prompting repaints in frameworks like React.

I have included an intro to the new section you added, as well as a mermaid diagram. Please take a look, and thanks a lot for your farsightedness here! :heart: 

cc @lukas-reining @thomaspoignant @kinyoklion @jonathannorris @beeme1mr @Kavindu-Dodan

toddbaert (approved):

specification/sections/05-events.md
```diff
@@ -144,3 +144,56 @@ For instance, _application authors_ may attach readiness handlers to be confiden
 If such handlers are attached after the provider underlying the client has already been initialized, they should run immediately.
 
 See [provider initialization](./02-providers.md#24-initialization), [setting a provider](./01-flag-evaluation.md#setting-a-provider).
+
+### Event handlers and context reconciliation
+
+Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
+This can be particularly useful for triggering UI repaints in multiple components when one component updates the [evaluation context](./03-evaluation-context.md).
+Optionally, some providers may transition to a the `STALE` state while their associated context is waiting to be reconciled, since this may involve asynchronous operations such as network calls.
```
toddbaert, line 152:
```suggestion
SDK implementations automatically fire the the `PROVIDER_CONTEXT_CHANGED` events if the `on context changed` handler terminates normally (and `PROVIDER_ERROR` events otherwise).
Optionally, some providers may transition to a the `STALE` state while their associated context is waiting to be reconciled, since this may involve asynchronous operations such as network calls.
```

Kavindu-Dodan (approved):

beeme1mr (approved):

fabriziodemaria (approved):
Semantically this updated set of events look good to me

I am not sure if we agreed which component should fire the events, if the SDK or the Provider (or both?). But this is probably an implementation detail that goes outside the scope of this PR

toddbaert:
> Semantically this updated set of events looks good to me
> 
> I am not sure if we agreed which component should fire the events, if the SDK or the Provider (or both?). But this is probably an implementation detail that goes outside the scope of this PR

In this case I did mention that (see last 2 points): 

> The `PROVIDER_CONTEXT_CHANGED` is not emitted from the provider itself; the SDK implementation must run the `PROVIDER_CONTEXT_CHANGED` handlers if the `on context changed` function terminates normally.

PR #210, opened 2023-10-13
fix: correct `event details` type, include `provider name`

toddbaert:
We made this change in the spec: https://openfeature.dev/specification/sections/events#requirement-523

But neglected to update it in "types"... it's already implemented in most SDKs.

### DIFF
--- a/specification/types.md
+++ b/specification/types.md
@@ -125,7 +125,7 @@ A structure defining a provider event payload, including:
 
 A structure defining an event payload, including:
 
-- client name (string, required)
+- provider name (string, required)
 - flags changed (string[], optional)
 - message (string, optional)
 - event metadata ([event metadata](#event-metadata))


specification/types.md
```diff
@@ -125,7 +125,7 @@ A structure defining a provider event payload, including:
 
 A structure defining an event payload, including:
 
-- client name (string, required)
```
beeme1mr, line 128: Perhaps you could add this as optional since it's already available in some SDKs.

beeme1mr (approved):

specification/types.md
toddbaert: When we say something is "optional" in all other places it means the SDK must define it, but it can be unset. We could add a note or sub-point about it, but TBH I'd rather just leave it unspecified. I think it offers much less value than the provider name.

toddbaert:
Merging since this is small, and a bug (inconsistency).

Issue #213, opened 2023-11-03
Specify provider state after `shutdown()`


toddbaert:
We don't currently specify the state of the provider after a `shutdown()`.

I'm proposing we should explicitly require (MUST) or suggest (SHOULD) the provider transitions back into either `NOT_READY` (implying that it can be re-initialized) or `READY` (for providers not requiring initialization) once it has been shut down. I think this is a fairly sensible approach, and will remove this ambiguity in a clean way. Some providers already work this way.

Essentially we'd be adding another transition in the state diagram [here](https://openfeature.dev/specification/sections/providers#requirement-242) from `READY` to `NOT_READY` via `shutdown`.

kinyoklion:
This effectively disallows providers that would hit some terminal state, they are neither ready, nor could they be initialized again. Potentially this is fine, but I think it is worth an explicit callout.

toddbaert:
> This effectively disallows providers that would hit some terminal state, they are neither ready, nor could they be initialized again. Potentially this is fine, but I think it is worth an explicit callout.

I agree. If we go this route, we can note that. Our state diagrams will also be circular, which will communicate the same thing.

justinabrahms:
+1

weyert:
In my opinion `shutdown` is a final state so no need to recover from. If they want to `recover` from it they can recreate the provider?

lukas-reining:
I would be fine with either a terminal state or the requirement for all providers to be able to be reenitialized which would imply the two states `READY` or `NOT_READY` as you said @toddbaert.
First I was leaning towards the terminal state but I think as every provider has to be initialized anyways, I would go for beeing able to reenitialize, which should be doable in any case (by just running everything from constructor and a full cleanup in the shutdown)

Kavindu-Dodan:
I am voting for going into `NOT_READY` state. When shutdown is called, the provider is meant to clean up its resources (as per spec _The provider MAY define a mechanism to gracefully shutdown and dispose of resources._ ). This clearly make them `NOT_READY` to serve flag evaluations. And the status changes beyond that are out of scope of the OF spec.

toddbaert:
> In my opinion shutdown is a final state so no need to recover from by going into a NOT_READY state. If they want to recover from it they can recreate the provider?

That was certainly the case before. The problem is the lack of any prescription here makes things ambiguous. It seems to me that we'd like to make the provider lifecycle as consistent as possible - it's already one of the trickiest areas for provider and SDK implementors.

[...]

PR #216, opened 2023-11-15
feat: NOT_READY after provider shutdown

toddbaert:
Add requirement clarifying providers requiring initialization return to `NOT_READY` after shutdown, as agreed on [here](https://github.com/open-feature/spec/issues/213).

Fixes: https://github.com/open-feature/spec/issues/213

### DIFF
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -257,7 +257,7 @@ class MyProvider implements Provider, AutoDisposable {
 
 > After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
 
-If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state, so it can be reinitialized.
+If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state. Some providers may allow re-initialization from this state.
 Providers not requiring initialization are assumed to be ready at all times.
 
 see: [initialization](#24-initialization)
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -257,7 +257,7 @@ class MyProvider implements Provider, AutoDisposable {
 
 > After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
 
-If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state. Some providers may allow re-initialization from this state.
+If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state. Some providers may allow reinitialization from this state.
 Providers not requiring initialization are assumed to be ready at all times.
 
 see: [initialization](#24-initialization)
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -257,7 +257,7 @@ class MyProvider implements Provider, AutoDisposable {
 
 > After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
 
-If a provider requires initialization, once it's shut down, it should transition to its initial `NOT_READY` state, so it can be reinitialized.
+If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state, so it can be reinitialized.
 Providers not requiring initialization are assumed to be ready at all times.
 
 see: [initialization](#24-initialization)
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -255,9 +255,9 @@ class MyProvider implements Provider, AutoDisposable {
 
 #### Requirement 2.5.2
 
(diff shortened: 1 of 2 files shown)


specification/sections/02-providers.md
```diff
@@ -203,12 +203,15 @@ title: Provider State
...
-    NOT_READY --> READY
+    NOT_READY --> READY:initialize
     READY --> ERROR
     ERROR --> READY
     READY --> STALE
     STALE --> READY
     STALE --> ERROR
+    READY --> NOT_READY:shutdown
+    STALE --> NOT_READY:shutdown
+    ERROR --> NOT_READY:shutdown
```
toddbaert, line 214: I added names for a few of the transitions because it made the diagram less confusing. Adding them everywhere seemed redundant though.

Kavindu-Dodan: The preview looks good. IMO good improvement to explain _what_ cause the state change.

```mermaid
---
title: Provider State
---
stateDiagram-v2
    direction LR
    [*] --> NOT_READY
    NOT_READY --> READY:initialize
    READY --> ERROR
    ERROR --> READY
    READY --> STALE
    STALE --> READY
    STALE --> ERROR
    READY --> NOT_READY:shutdown
    STALE --> NOT_READY:shutdown
    ERROR --> NOT_READY:shutdown
```

specification/sections/02-providers.md
```diff
@@ -250,6 +253,15 @@ class MyProvider implements Provider, AutoDisposable {
   }
 ```
 
+#### Requirement 2.5.2
+
+> After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
+
+If a provider requires initialization, once it's shut down, it should transition to its initial `NOT_READY` state, so it can be reinitialized.
```
Kavindu-Dodan, line 260: >  ... so it can be reinitialized

I think we should not enforce this behavior through spec 🤔

Kavindu-Dodan (approved):
Left [this](https://github.com/open-feature/spec/pull/216/files#r1394677422) comment, otherwise change looks fine.

specification/sections/02-providers.md
toddbaert: Do you mean we should just remove the "so it can be reinitialized"?

We could remove it, but I think that this is the logical implication of transitioning back to `NOT_READY`. On the other hand, I guess you could make the argument some providers could still be unable to be made `READY` again, and they could throw if they are reinitialized.

specification/sections/02-providers.md
```diff
@@ -250,6 +253,15 @@ class MyProvider implements Provider, AutoDisposable {
   }
 ```
 
+#### Requirement 2.5.2
+
+> After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
```
beeme1mr, line 258: The non-normative section says "should," while the spec says" must".

specification/sections/02-providers.md
toddbaert, line 260:
```suggestion
If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state, so it can be reinitialized.
```

specification/sections/02-providers.md
toddbaert: Good call, fixed.

specification/sections/02-providers.md
Kavindu-Dodan: Yes, originally thought we shouldn't define this through the spec. But with your explanation, I think we should reword it to have **MAY** - ex:-  ... so it **MAY** be reinitialized.

specification/sections/02-providers.md
```diff
@@ -250,6 +253,15 @@ class MyProvider implements Provider, AutoDisposable {
...
+
+If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state, so it can be reinitialized.
```
toddbaert, line 260:
```suggestion
If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state. Some providers may allow re-initialization from this state.
```

specification/sections/02-providers.md
```diff
@@ -250,6 +253,15 @@ class MyProvider implements Provider, AutoDisposable {
   }
 ```
 
+#### Requirement 2.5.2
+
+> After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
+
+If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state. Some providers may allow re-initialization from this state.
```
toddbaert, line 260:
```suggestion
If a provider requires initialization, once it's shut down, it must transition to its initial `NOT_READY` state. Some providers may allow reinitialization from this state.
```

lukas-reining (approved):
This looks good to me.

toddbaert:
I'm going to merge this Friday EOD unless anyone has concerns.

weyert (approved):

PR #217, opened 2023-11-20
fix: add dynamic-context condition to 4.1.4

toddbaert:
I neglected to add this qualifier when we released the updates for client (static-context). This requirement isn't relevant when the context is global and can only be changed via `OpenFeature.setContext`.

Thanks for the [find](https://github.com/open-feature/js-sdk/issues/630#issuecomment-1818743111), @lukas-reining .

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -559,11 +559,19 @@
             "children": []
         },
         {
-            "id": "Requirement 4.1.4",
-            "machine_id": "requirement_4_1_4",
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):

beeme1mr (approved):

lukas-reining (approved):
Looks good!

PR #220, opened 2023-11-29
chore: fix site deploy

Arhell:
## This PR
fix https://github.com/open-feature/openfeature.dev/pull/271


beeme1mr:
I'm okay with this approach but we should confirm that the order in the side nav remains unchanged. That was the reason the sections had the prefix to begin with.

PR #229, opened 2024-01-10
feat: add domain as an openfeature concept

### LINKED ISSUE #228: Redefine named clients as domain
## Overview

The concept of a named client was introduced in the [v0.6.0 release](https://github.com/open-feature/spec/releases/tag/v0.6.0). The goal was to add support for multiple providers so that they could be used within a single application. It worked by mapping clients to providers using the client name. While the functionality has proven to be very useful, the terminology used to describe the feature can be misleading and confusing.

## Proposal

I suggest that we redefine the named client concept as `domain`. I think this term makes it clear that a user can leverage domains to map clients to providers.

## Impact

This change should have little to no impact on the implementation within the SDKs. However, documentation will need to be updated accordingly. It will also require the [SDK compatibility overview](https://openfeature.dev/docs/reference/technologies/sdk-compatibility) to be updated.

beeme1mr:
## This PR

- defines `domain` as a concept
- updates references to client name to use `domain`
- add `domain` as client metadata

### Related Issues

Fixes #228

### Follow-up Tasks

- Update the readme template
- Update the openfeature.dev SDK compatibility table

### DIFF
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -3,8 +3,8 @@ title: Glossary
 description: A list of terms used within the OpenFeature specification.
 sidebar_position: 1
 ---
-<!-- omit from toc -->
-# Glossary 
+
+# Glossary <!-- omit from toc -->
 
 This document defines some terms that are used across this specification.
 
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -13,37 +13,39 @@ This document defines some terms that are used across this specification.
 
 <!-- toc -->
 
-- [Feature Flag](#feature-flag)
-- [User Roles](#user-roles)
-  - [Application Author](#application-author)
-  - [Application Integrator](#application-integrator)
-  - [Provider Author](#provider-author)
-  - [Integration Author](#integration-author)
-  - [Library Author](#library-author)
-- [Common](#common)
-  - [Feature Flag SDK](#feature-flag-sdk)
-  - [Client-Side SDK](#client-side-sdk)
-  - [Server-Side SDK](#server-side-sdk)
-  - [Feature Flag API](#feature-flag-api)
-  - [Evaluation API](#evaluation-api)
-  - [Flag Management System](#flag-management-system)
-  - [Provider](#provider)
-  - [Integration](#integration)
-  - [Evaluation Context](#evaluation-context)
-  - [Evaluating Flag Values](#evaluating-flag-values)
-  - [Resolving Flag Values](#resolving-flag-values)
-- [Flagging specifics](#flagging-specifics)
-  - [Flag](#flag)
-  - [Flag Key](#flag-key)
-  - [Variant](#variant)
-  - [Values](#values)
-  - [Targeting](#targeting)
-  - [Targeting Key](#targeting-key)
-  - [Fractional Evaluation](#fractional-evaluation)
-  - [Rule](#rule)
-- [SDK Paradigms](#sdk-paradigms)
-  - [Dynamic-Context Paradigm](#dynamic-context-paradigm)
-  - [Static-Context Paradigm](#static-context-paradigm)
+- [Glossary](#glossary)
+  - [Feature Flag](#feature-flag)
+  - [User Roles](#user-roles)
+    - [Application Author](#application-author)
+    - [Application Integrator](#application-integrator)
+    - [Provider Author](#provider-author)
+    - [Integration Author](#integration-author)
+    - [Library Author](#library-author)
+  - [Common](#common)
+    - [Feature Flag SDK](#feature-flag-sdk)
+    - [Client-Side SDK](#client-side-sdk)
+    - [Server-Side SDK](#server-side-sdk)
+    - [Feature Flag API](#feature-flag-api)
+    - [Evaluation API](#evaluation-api)
+    - [Flag Management System](#flag-management-system)
+    - [Provider](#provider)
+    - [Namespace](#namespace)
+    - [Integration](#integration)
+    - [Evaluation Context](#evaluation-context)
+    - [Evaluating Flag Values](#evaluating-flag-values)
+    - [Resolving Flag Values](#resolving-flag-values)
+  - [Flagging specifics](#flagging-specifics)
+    - [Flag](#flag)
+    - [Flag Key](#flag-key)
+    - [Variant](#variant)
+    - [Values](#values)
+    - [Targeting](#targeting)
+    - [Targeting Key](#targeting-key)
+    - [Fractional Evaluation](#fractional-evaluation)
+    - [Rule](#rule)
+  - [SDK Paradigms](#sdk-paradigms)
+    - [Dynamic-Context Paradigm](#dynamic-context-paradigm)
+    - [Static-Context Paradigm](#static-context-paradigm)
 
 <!-- tocstop -->
 
@@ -107,7 +109,9 @@ A source-of-truth for flag values and rules. Flag management systems may include
 
 An SDK-compliant implementation which resolves flag values from a particular flag management system, allowing the use of the [Evaluation API](./sections/01-flag-evaluation.md#13-flag-evaluation) as an abstraction for the system in question.
 
-Providers can be used in two ways. Client-specific providers are active for specific clients, based on their name. The default provider is used if there are no client-specific mappings setup.
+### Namespace
+
+An identifier which logically associates clients with providers, allowing for multiple providers to be used simultaneously within a single application.
 
 ### Integration
 
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -3,8 +3,8 @@ title: Glossary
 description: A list of terms used within the OpenFeature specification.
 sidebar_position: 1
 ---
-
-# Glossary
+<!-- omit from toc -->
+# Glossary 
 
 This document defines some terms that are used across this specification.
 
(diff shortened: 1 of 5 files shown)


specification/glossary.md
```diff
@@ -13,37 +13,39 @@ This document defines some terms that are used across this specification.
...
-  - [Variant](#variant)
-  - [Values](#values)
-  - [Targeting](#targeting)
-  - [Targeting Key](#targeting-key)
-  - [Fractional Evaluation](#fractional-evaluation)
-  - [Rule](#rule)
-- [SDK Paradigms](#sdk-paradigms)
-  - [Dynamic-Context Paradigm](#dynamic-context-paradigm)
-  - [Static-Context Paradigm](#static-context-paradigm)
+- [Glossary](#glossary)
```
beeme1mr, line 16: I'll exclude the h1 from the toc. It was automatically added by VSCode

specification/sections/01-flag-evaluation.md
```diff
@@ -104,6 +107,13 @@ See [hooks](./04-hooks.md) for details.
 OpenFeature.getProviderMetadata();
 ```
 
+TODO: Add text explaining that this is for a namespace
```
beeme1mr, line 110: Action the TODO

dabeeeenster:
Would this allow for multiple clients across multiple (or the same) provider?

beeme1mr:
> Would this allow for multiple clients across multiple (or the same) provider?

Yes, exactly. The functionality already exists in many SDKs.

toddbaert:
> > Would this allow for multiple clients across multiple (or the same) provider?
> 
> Yes, exactly. The functionality already exists in many SDKs.

Ya, it exists but is (IMHO) confusingly named, so I like this change.

federicobond:
I agree with the motivation, the current wording around named clients is a bit awkward. I don't think the word namespace makes things clear though. A namespace usually denotes a logical grouping of identifiers or symbols.

I propose we survey some examples of APIs solving similar problems to get a better idea of the alternatives.

As a first data point, the Python `logging` library uses "name" for the parameter to `logging.getLogger` which returns a logger with a specific configuration defined somewhere else.

beeme1mr:
@federicobond I thought namespace would be a good fit because it's a logical grouping of clients to a provider. However, I would happily change the proposal if we can find a more appropriate term.

toddbaert:
> I agree with the motivation, the current wording around named clients is a bit awkward. I don't think the word namespace makes things clear though. A namespace usually denotes a logical grouping of identifiers or symbols.
> 
> I propose we survey some examples of APIs solving similar problems to get a better idea of the alternatives.
> 
> As a first data point, the Python `logging` library uses "name" for the parameter to `logging.getLogger` which returns a logger with a specific configuration defined somewhere else.

I don't think we need to go with "namespace" but I don't think "name" is a good candidate. We've had feedback that name makes it sound like a unique identifier for a client/provider, which this is not.

> A namespace usually denotes a logical grouping of identifiers or symbols.

@federicobond I would argue that's pretty much what we're going for. A "namespace" in this case could group clients/providers/contexts/events in a single scope ("scope" would also possibly be a decent name) associating them all. "ClientName" already does this, but it's a poor descriptor.

federicobond:
I think the awkward part in named clients is not the "named" but the "clients" bit, which as you say makes it sound like a unique identifier for a client/provider.

I think the confusion can be resolved by keeping name but rewording the spec to clarify it refers to a scope.

Looking at how OpenTelemetry handles this, it says:

> name (required): This name SHOULD uniquely identify the [instrumentation scope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope), such as the [instrumentation library](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-library) (e.g. io.opentelemetry.contrib.mongodb), package, module or class name.

https://opentelemetry.io/docs/specs/otel/trace/api/#get-a-tracer

sheepduke:
Now in the Rust SDK "named" is used. We have functions like `create_client` and `create_named_client`. I agree that it is a bit confusing because it sounds like "create a client that has a name", instead of "create a client that binds to a specific provider". 

Personally I think "namespace" maybe too "strong" towards programming side. Maybe something similar to "scope" or "category" feels more business/domain specific. `create_scoped_client` or `create_categorized_client` etc maybe?

justinabrahms:
Instead of thinking about them as "named clients", is it perhaps useful to think of them as "scoped providers"?

sheepduke:
> Instead of thinking about them as "named clients", is it perhaps useful to think of them as "scoped providers"?

I agree. Just wondering if there is a better name than *scope*. The *scope* (or *namespace* etc) are actually defined for providers. The clients are just *bound* to the corresponding provider.

lukas-reining:
> I think the awkward part in named clients is not the "named" but the "clients" bit, which as you say makes it sound like a unique identifier for a client/provider.
> 
> I think the confusion can be resolved by keeping name but rewording the spec to clarify it refers to a scope.
> 
> Looking at how OpenTelemetry handles this, it says:
> 
> > name (required): This name SHOULD uniquely identify the [instrumentation scope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope), such as the [instrumentation library](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-library) (e.g. io.opentelemetry.contrib.mongodb), package, module or class name.
> 
> https://opentelemetry.io/docs/specs/otel/trace/api/#get-a-tracer

It feels that a naming for the concept of such a "named provider registration" would be helpful.
Semantically we do not want to give a name to a client, so giving a `name` to a `getClient` method feels misleading to me.

> Personally I think "namespace" maybe too "strong" towards programming side.

This is also what I feel. To me, the semantics of `namespace` are not exactly what we want to say.

The best thing I can think of is `scope`, but it does not fully capture this "named provider registration".

federicobond:
The key concept here is that named clients allow for *late binding* providers to clients, so it's never the providers you want to name but the scope of the clients. You want the clients to not know anything about the provider that they will end up using.

In particular, this would be a bad practice:

```js
OpenFeature.setProvider("flagd", new FlagDProvider())
...
OpenFeature.getClient("flagd") // do not do this
```

You rather want:

```js
OpenFeature.setProvider("billingsystem", new FlagDProvider())
...
OpenFeature.getClient("billingsystem")
```

A word that may better capture the meaning of what we want to do here and lead users away from naming the individual providers is `domain`. A client then would be initialized with a `domain` and a provider would be bound to a `domain`.

federicobond:
Another important discussion that is out of the scope of this thread but might come up in the future is the possibility of these scopes/domains having hierarchical semantics.

In that were the case, a provider bound to, for example, `billingsystem` could be used if there were no provider bound to `billingsystem.subsystem`.

toddbaert:
> The key concept here is that named clients allow for late binding providers to clients, so it's never the providers you want to name but the scope of the clients. You want the clients to not know anything about the provider that they will end up using.

This is very well-put!. "Late binding" is exactly what we want to support and communicate. :clinking_glasses: 

I don't so much care we do it, but I think we need to communicate this better, both to SDK implementors and API users. Maybe we just need to improve our non-normative text and some inline doc... I would also be open to changing the term to "domain", because I agree with the idea [here](https://github.com/open-feature/spec/pull/229#issuecomment-1891083384); this is fundamentally what we wanted to support, particularly for large monolithic applications which might encompass multiple teams and problem "domains".

beeme1mr:
Thanks for the feedback. I'll bring this up as a topic tomorrow in the community meeting. I think we're close to finding a good term.

maxveldink:
No new thoughts here than what's been said, but I think the term `domain` fits well for the use case our company has for this. We have names such as `pay_flags,` `tax_calculation`, and `legacy_flags`, which bind to different provider implementations. Clarifying the spec to call these `domains` would have helped clarify which names to use.

federicobond:
As a recap, if I understand correctly, the confusion with named clients arises because the name does not affect the client in any way other than to enable it to late bind to a provider that is different from the one configured globally.

This leads some to think that it's the provider that is given a name, but using provider names defeats the original motivation for introducing this functionality: being able to swap the provider from a central location without having to touch all the client code across different modules and files, code that may even belong to third-party dependencies.

Introducing the word `domain` as a concept helps clarify the purpose of this value: A domain is a string identifying a scope of flags that are covered by a single provider.

The word domain is used also in the [gettext](https://www.gnu.org/software/gettext/manual/gettext.html#index-several-domains) translation system, where it serves a very purpose: one can bind a message catalog to a specific domain and have multiple domains within a single software system served by different catalogs.

beeme1mr:
Thanks for the feedback everyone. I've updated the PR to use the term `domain`. Please take a look when you have a moment.

specification/glossary.md
```diff
@@ -107,7 +108,9 @@ A source-of-truth for flag values and rules. Flag management systems may include
 
 An SDK-compliant implementation which resolves flag values from a particular flag management system, allowing the use of the [Evaluation API](./sections/01-flag-evaluation.md#13-flag-evaluation) as an abstraction for the system in question.
 
-Providers can be used in two ways. Client-specific providers are active for specific clients, based on their name. The default provider is used if there are no client-specific mappings setup.
+### Domain
+
```
Kavindu-Dodan, line 112: nit - extra space

specification/sections/01-flag-evaluation.md
```diff
@@ -104,25 +107,38 @@ See [hooks](./04-hooks.md) for details.
 OpenFeature.getProviderMetadata();
 ```
 
-See [provider](./02-providers.md) for details.
+It must be possible to access provider metadata using a `domain`.
+If a provider has not be registered under the requested domain, the default provider metadata is returned.
+
+```typescript
+// example provider accessor
+OpenFeature.getProviderMetadata("domain-1");
```
Kavindu-Dodan, line 115: We might need to check SDK implementations when this spec get merged. Java already expose this [^1] and I created a follow-up for Go SDK [^2]. Needs to check other SDKs 

[^1]: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.java#L53-L54
[^2]: https://github.com/open-feature/go-sdk/issues/247

Kavindu-Dodan (approved):

specification/glossary.md
beeme1mr: This looks correct to me. What extra space are you referring to?

specification/sections/01-flag-evaluation.md
```diff
@@ -104,25 +107,38 @@ See [hooks](./04-hooks.md) for details.
...
 
 ```java
 // example client creation and retrieval
-OpenFeature.getClient("my-named-client");
+OpenFeature.getClient();
 ```
 
-The name is a logical identifier for the client which may be associated with a particular provider by the application integrator.
-If a client name is not bound to a particular provider, the client is associated with the default provider.
+It must be possible to create a client that is associated with a `domain`.
```
toddbaert, line 133:
```suggestion
It's possible to create a client that is associated with a `domain`.
```

```suggestion
It's possible to create a client that is associated with a `domain`.
```

I try to avoid must/should/etc (RFC2119 language) in non-normative sections. I recommend speaking as if things are already implemented in these sections (like the sentence below: "the client will..."

specification/sections/03-evaluation-context.md
```diff
@@ -79,19 +79,19 @@ In the static-context paradigm, context is global. The client and invocation can
...
 
-See [setting a provider](./01-flag-evaluation.md#setting-a-provider) for details.
+See [setting a provider](./01-flag-evaluation.md#setting-a-provider), [domain](../glossary.md#domain) for details.
 
 ##### Conditional Requirement 3.2.2.4
 
-> The API **MUST** have a a mechanism to manage `evaluation context` for an associated name.
+> The API **MUST** have a mechanism to manage `evaluation context` for an associated `domain`.
 
 In the static-context paradigm, it must be possible to create and remove provider-specific context.
```
toddbaert, line 93:
```suggestion
In the static-context paradigm, it's possible to create and remove provider-specific context.
```

specification/sections/01-flag-evaluation.md
```diff
@@ -104,25 +107,38 @@ See [hooks](./04-hooks.md) for details.
 OpenFeature.getProviderMetadata();
 ```
 
-See [provider](./02-providers.md) for details.
+It must be possible to access provider metadata using a `domain`.
```
toddbaert, line 110:
```suggestion
It's possible to access provider metadata using a `domain`.
```

toddbaert (approved):

federicobond (approved):

thomaspoignant (approved):

lukas-reining (approved):

sheepduke (approved):

beeme1mr:
Thanks for the great reviews and feedback. I'll merge this on Wednesday the 24th unless there's an objection.

austindrenski (approved):

dominikhaska (approved):
Looks good - thank you :)

PR #239, opened 2024-01-29
chore: Clarify resolution detail reason.

### LINKED ISSUE #236: Clarification of reason for evaluation detail and evaluation.
In the specification it is detailed that the `reason` is an optional string, and there are a set of pre-defined values, but I do not think it is clear that free-form values are intended to be supported as well. This has resulted in some SDK development which restricts the reason to an enumeration and prevents conveying reasons more specific to the vendor.

A current example of this in the python SDK: https://github.com/open-feature/python-sdk/blob/d1f27e3278964c2d81b780db97deace87d02cae1/openfeature/flag_evaluation.py#L21

(A different issue will be opened for that specific instance.)

kinyoklion:
## This PR

Attempt to clarify the `reason` field of the resolution details.

### Related Issues

Fixes #236

### DIFF
--- a/specification/types.md
+++ b/specification/types.md
@@ -88,10 +88,10 @@ enum Reason {
     Unknown,
     Stale,
     Error,
-    Custom(String)
+    Other(String)
 }
 
-let myReason = Reason::Custom("my-reason".to_string());
+let myReason = Reason::Other("my-reason".to_string());
 ```
 
 > [!NOTE]
--- a/specification/types.md
+++ b/specification/types.md
@@ -73,6 +73,28 @@ A set of pre-defined reasons is enumerated below:
 | ERROR           | The resolved value was the result of an error.                                                                                   |
 
 > [!NOTE]
+> The `reason` should not be limited to the reasons enumerated above. It can be any of the pre-defined reasons, or
+> any string value. Some type systems have features which can increase the ergonomics of `reason`, for instance a union
+of pre-defined types with a string, or a rust-style enumeration which allows for enumerated values to have associated
+content.
+
+```rust
+enum Reason {
+    Static,
+    Default,
+    TargetingMatch,
+    Split,
+    Cached,
+    Unknown,
+    Stale,
+    Error,
+    Custom(String)
+}
+
+let myReason = Reason::Custom("my-reason".to_string());
+```
+
+> [!NOTE]
 > The `resolution details` structure is not exposed to the Application Author.
 > It defines the data which Provider Authors must return when resolving the value of flags.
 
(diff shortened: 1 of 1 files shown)


specification/types.md
```diff
@@ -72,6 +72,28 @@ A set of pre-defined reasons is enumerated below:
 | STALE           | The resolved value is non-authoritative or possibly out of date                                                                  |
 | ERROR           | The resolved value was the result of an error.                                                                                   |
 
+> [!NOTE]
+> The `reason` should not be limited to the reasons enumerated above. It can be any of the pre-defined reasons, or
+> any string value. Some type systems have features which can increase the ergonomics of `reason`, for instance a union
+of pre-defined types with a string, or a rust-style enumeration which allows for enumerated values to have associated
+content.
+
+```rust
```
kinyoklion, line 81: I am not sure if an example here is overkill.

kinyoklion: The solution to this varies widely language to language, so maybe just pointing it out is enough. 

The js-sdk:
```typescript
export type ResolutionReason = keyof typeof StandardResolutionReasons | (string & Record<never, never>);
```

The solution added to the python SDK.
```python
reason: typing.Optional[typing.Union[str, Reason]] = None
```

specification/types.md
```diff
@@ -72,6 +72,28 @@ A set of pre-defined reasons is enumerated below:
...
+enum Reason {
+    Static,
+    Default,
+    TargetingMatch,
+    Split,
+    Cached,
+    Unknown,
+    Stale,
+    Error,
+    Other(String)
```
kinyoklion, line 91: I had `Custom` here, but I looked at the rust SDK implementation. It was using this pattern already, and has `Other` which is nice.

specification/types.md
toddbaert: I like having a rust example. and this is clear to me.

toddbaert (approved):

specification/types.md
```diff
@@ -72,6 +72,28 @@ A set of pre-defined reasons is enumerated below:
 | STALE           | The resolved value is non-authoritative or possibly out of date                                                                  |
 | ERROR           | The resolved value was the result of an error.                                                                                   |
 
+> [!NOTE]
+> The `reason` should not be limited to the reasons enumerated above. It can be any of the pre-defined reasons, or
+> any string value. Some type systems have features which can increase the ergonomics of `reason`, for instance a union
+of pre-defined types with a string, or a rust-style enumeration which allows for enumerated values to have associated
+content.
```
beeme1mr, line 79:
```suggestion
> of pre-defined types with a string, or a rust-style enumeration which allows for enumerated values to have associated
> content.
```

beeme1mr (approved):

thomaspoignant (approved):

Issue #238, opened 2024-01-29
Proposal to move `Provider Status` field from provider to SDK, refine semantics around ContextChange events


toddbaert:
# Background

Currently, provider authors are required to maintain a field indicating their [state](https://openfeature.dev/specification/types#provider-status): (`NOT_READY`/`READY`/`STALE`/`ERROR`).
This burdens provider authors in a few ways that are not particularly ergonomic, and creates potential foot-guns.
This is particularly concerning for SDKs which rely heavily on events (React, for example); it's quite possible for the "suspense" functionality to break and suspend indefinitely if a provider fails to correctly set their READY state after they are initialized. Additionally, we may need additional states/events to support emerging client use-cases.

## Problem 1: State-management at Initialization

The SDK itself calls `initialize()` on any provider in state `NOT_READY`.
The SDK then emits an `READY` or `ERROR` event based on the successful/unsuccessful termination of `initialize()`.
However, it's the provider's responsibility to then set the `state` correctly.
This results in confusion, and frequently provider authors neglect to properly set the providers status (this is often caught in PRs).

## Problem 2: State-management at context change (client paradigm only)

The static context paradigm (client-side paradigm) defines a `context change` method that can be implemented by the provider, which performs whatever operations are needed to reconcile the evaluated flags with the new context.
Similar to problem above, the burden is put on the provider author to correctly update state (as well as emit the STALE event) when this function is called: https://openfeature.dev/specification/sections/events/#condition-534. Additionally, the semantics of PROVIDER_STALE event may not be ideal or sufficient for this use case; STALE was not originally intended to indicate the providers flags needed to be resolved due to a context change.

## Solution:

To resolve the both the above, the provider state should be maintained by the SDK, and become only notional to the application author, and the SDK should emit all events that are part of the provider lifecycle.
This will make the provider lifecycle entirely the domain of the SDK; the provider would need only to implement the optional `initialize` and `shutdown`, and `context change` (client-only) methods, and the SDK should handle the rest.
From a provider-author's perspective, providers would become stateless; the authors job is simply to implement the desired interfaces, while the SDK would manage the entire lifecycle as appropriate, based on the implemented methods.
SDKs would run initialize/onContextChange/shutdown if they are defined, and maintain any state if necessary.
**_Providers would only have to ensure initialize and shutdown are idempotent and re-entrant._**
[...]

PR #241, opened 2024-02-06
feat: internal provider state, new client events

toddbaert:
* moves provider state into SDK (SDK now maintains provider state/lifecycle; provider interface is now "stateless")
* refine semantics around client state when context is pending/reconciled by using new events/states, instead of STALE
* add PROVIDER_FATAL error code

Resolves: https://github.com/open-feature/spec/issues/238

--- 

For background discussion, see https://github.com/open-feature/spec/issues/238

---

This should make provider authorship less error prone, and simpler. Providers essentially become stateless as far as the SDK is concerned. The provider is only obligated to optionally implement the `initialize`, `shutdown` and `onContextChange` functions and maintain is own private state (caches, connection, etc).

This also refines the semantics around context reconciliation, which is particularly important for client SDKs where providers need to reconcile their flags or ruleset(s) when context changes.

I believe these changes can all be implemented by SDKs in a non-breaking fashion.

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -415,6 +415,8 @@ stateDiagram-v2
     [*] --> NOT_READY
     NOT_READY --> READY:initialize()
     NOT_READY --> ERROR:initialize()
+    NOT_READY --> FATAL:initialize()
+    FATAL --> [*]
     READY --> ERROR:*
     ERROR --> READY:*
     READY --> STALE:*
@@ -423,19 +425,23 @@ stateDiagram-v2
     READY --> NOT_READY:shutdown()
     STALE --> NOT_READY:shutdown()
     ERROR --> NOT_READY:shutdown()
-    READY --> CONTEXT_PENDING:::optional:setContext()
-    CONTEXT_PENDING --> READY:*
+    READY --> RECONCILING:::client:setContext()
+    RECONCILING:::client --> READY:*
 
-    classDef optional fill:#999
+    classDef client fill:#555
 ```
 
 \* transitions occurring when associated events are spontaneously emitted from the provider
 
-Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `CONTEXT_PENDING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
+<span style="color:#555">█</span> only defined in static-context (client-side) paradigm
+
+Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `RECONCILING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
 
 #### Requirement 1.7.1
 
-> The `client` **MUST** define a `provider status` accessor which indicates the readiness of the associated provider, with possible values `NOT_READY`, `READY`, `STALE`, or `ERROR`.
+> The `client` **MUST** define a `provider status` accessor which indicates the readiness of the associated provider, with possible values `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`.
+
+The SDK at all times maintains an up-to-date state corresponding to the success/failure of the last lifecycle method (`initialize`, `shutdown`, `on context change`) or emitted event.
 
 see [provider status](../types.md#provider-status)
 
@@ -449,7 +455,7 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
 
 ##### Conditional Requirement 1.7.2.1
 
-> In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `CONTEXT_PENDING`.
+> In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `RECONCILING`.
 
 In the static context paradigm, the implementation must define a `provider status` indicating that a provider is reconciling its internal state due to a context change.
 
@@ -467,8 +473,32 @@ If the provider has failed to initialize, the `provider status` should indicate 
 
 #### Requirement 1.7.5
 
-> The client **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
+> The client's `provider status` accessor **MUST** indicate `FATAL` if the `initialize` function of the associated provider terminates abnormally and indicates `error code` `PROVIDER_FATAL`.
 
-The SDK should return an informative `error code`, such as `PROVIDER_NOT_READY`, if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
+If the provider has failed to initialize, the `provider status` should indicate the provider is in an error state.
+
+#### Requirement 1.7.6
+
+> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.
+
+The client defaults and returns the `PROVIDER_NOT_READY` `error code` if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
+The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
+
+see: [error codes](https://openfeature.dev/specification/types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
+
+#### Requirement 1.7.7
+
+> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `PROVIDER_FATAL`.
+
+TThe client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
+The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
+
+see: [error codes](https://openfeature.dev/specification/types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
+
+#### Requirement 1.7.8
+
+> Implementations **SHOULD** propagate the `error code` returned from any provider lifecycle methods.
+
+The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
 
 see: [error codes](https://openfeature.dev/specification/types#error-code)
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -428,12 +428,12 @@ stateDiagram-v2
     READY --> RECONCILING:::client:setContext()
     RECONCILING:::client --> READY:*
 
-    classDef client fill:#555
+    classDef client fill:#888
 ```
 
 \* transitions occurring when associated events are spontaneously emitted from the provider
 
-<span style="color:#555">█</span> only defined in static-context (client-side) paradigm
+<span style="color:#888">█</span> only defined in static-context (client-side) paradigm
 
 > [!NOTE]
 > Only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `RECONCILING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -450,7 +450,7 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
 
 > In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `CONTEXT_PENDING`.
 
-In the static context paradigm, the implementation must define a `provider status` indicating that a provider is reconciling its internal state with a context change.
+In the static context paradigm, the implementation must define a `provider status` indicating that a provider is reconciling its internal state due to a context change.
 
 #### Requirement 1.7.3
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -501,4 +501,4 @@ see: [error codes](https://openfeature.dev/specification/types#error-code), [fla
 
 The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
 
-see: [error codes](https://openfeature.dev/specification/types#error-code)
+see: [error codes](../types#error-code)
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -435,7 +435,8 @@ stateDiagram-v2
 
 <span style="color:#555">█</span> only defined in static-context (client-side) paradigm
 
-Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `RECONCILING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
+> [!NOTE]
+> Only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `RECONCILING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
 
 #### Requirement 1.7.1
 
@@ -493,7 +494,7 @@ see: [error codes](../types.md#error-code), [flag value resolution](./02-provide
 TThe client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
 The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
 
-see: [error codes](../types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
+see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
 
 #### Requirement 1.7.8
 
@@ -501,4 +502,4 @@ see: [error codes](../types#error-code), [flag value resolution](./02-providers.
 
 The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
 
-see: [error codes](../types#error-code)
+see: [error codes](../types.md#error-code)
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -491,7 +491,7 @@ see: [error codes](../types.md#error-code), [flag value resolution](./02-provide
 
 > The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `PROVIDER_FATAL`.
 
-TThe client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
+The client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
 The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
 
 see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -410,27 +410,28 @@ This state of the provider is exposed on associated `clients`.
 The diagram below illustrates the possible states and transitions of the `state` field for a provider during the provider lifecycle.
 
 ```mermaid
----
-title: Provider State
----
 stateDiagram-v2
     direction LR
     [*] --> NOT_READY
     NOT_READY --> READY:initialize()
     NOT_READY --> ERROR:initialize()
-    READY --> ERROR*
-    ERROR --> READY*
-    READY --> STALE*
-    STALE --> READY*
-    STALE --> ERROR*
+    READY --> ERROR:*
+    ERROR --> READY:*
+    READY --> STALE:*
+    STALE --> READY:*
+    STALE --> ERROR:*
     READY --> NOT_READY:shutdown()
     STALE --> NOT_READY:shutdown()
     ERROR --> NOT_READY:shutdown()
+    READY --> CONTEXT_PENDING:::optional:setContext()
+    CONTEXT_PENDING --> READY:*
+
+    classDef optional fill:#999
 ```
 
 \* transitions occurring when associated events are spontaneously emitted from the provider
 
-Note that SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define additional states to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
+Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `CONTEXT_PENDING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
 
 #### Requirement 1.7.1
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -480,7 +480,7 @@ If the provider has failed to initialize, the `provider status` should indicate 
 
 #### Requirement 1.7.6
 
-> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.
+> The client **MUST** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.
 
 The client defaults and returns the `PROVIDER_NOT_READY` `error code` if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
 The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
@@ -489,7 +489,7 @@ see: [error codes](../types.md#error-code), [flag value resolution](./02-provide
 
 #### Requirement 1.7.7
 
-> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `PROVIDER_FATAL`.
+> The client **MUST** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `PROVIDER_FATAL`.
 
 The client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
 The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -493,7 +493,7 @@ see: [error codes](https://openfeature.dev/specification/types#error-code), [fla
 TThe client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
 The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
 
-see: [error codes](https://openfeature.dev/specification/types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
+see: [error codes](../types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
 
 #### Requirement 1.7.8
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -401,3 +401,73 @@ The global API object might expose a `shutdown` function, which will call the re
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
 
 see: [`shutdown`](./02-providers.md#25-shutdown)
+
+### 1.7. Provider Lifecycle Management
+
+The implementation maintains an internal representation of the state of configured providers, tracking the lifecycle of each provider.
+This state of the provider is exposed on associated `clients`.
+
+The diagram below illustrates the possible states and transitions of the `state` field for a provider during the provider lifecycle.
+
+```mermaid
+---
+title: Provider State
+---
+stateDiagram-v2
+    direction LR
+    [*] --> NOT_READY
+    NOT_READY --> READY:initialize()
+    NOT_READY --> ERROR:initialize()
+    READY --> ERROR*
+    ERROR --> READY*
+    READY --> STALE*
+    STALE --> READY*
+    STALE --> ERROR*
+    READY --> NOT_READY:shutdown()
+    STALE --> NOT_READY:shutdown()
+    ERROR --> NOT_READY:shutdown()
+```
+
+\* transitions occurring when associated events are spontaneously emitted from the provider
+
+Note that SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define additional states to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
+
+#### Requirement 1.7.1
+
+> The `client` **MUST** define a `provider status` accessor which indicates the readiness of the associated provider, with possible values `NOT_READY`, `READY`, `STALE`, or `ERROR`.
+
+see [provider status](../types.md#provider-status)
+
+#### Condition 1.7.2
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+> The implementation uses the static-context paradigm.
+
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
+
+##### Conditional Requirement 1.7.2.1
+
+> In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `CONTEXT_PENDING`.
+
+In the static context paradigm, the implementation must define a `provider status` indicating that a provider is reconciling its internal state with a context change.
+
+#### Requirement 1.7.3
+
+> The client's `provider status` accessor **MUST** indicate `READY` if the `initialize` function of the associated provider terminates normally.
+
+Once the provider has initialized, the `provider status` should indicate the provider is ready to be used to evaluate flags.
+
+#### Requirement 1.7.4
+
+> The client's `provider status` accessor **MUST** indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.
+
+If the provider has failed to initialize, the `provider status` should indicate the provider is in an error state.
+
+#### Requirement 1.7.5
+
+> The client **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
+
+The SDK should return an informative `error code`, such as `PROVIDER_NOT_READY` if evaluation in attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
+
+see: [error codes](https://openfeature.dev/specification/types#error-code)
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -410,6 +410,9 @@ This state of the provider is exposed on associated `clients`.
 The diagram below illustrates the possible states and transitions of the `state` field for a provider during the provider lifecycle.
 
 ```mermaid
+---
+title: Provider lifecycle
+---
 stateDiagram-v2
     direction LR
     [*] --> NOT_READY
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -468,6 +468,6 @@ If the provider has failed to initialize, the `provider status` should indicate 
 
 > The client **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
 
-The SDK should return an informative `error code`, such as `PROVIDER_NOT_READY` if evaluation in attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
+The SDK should return an informative `error code`, such as `PROVIDER_NOT_READY`, if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
 
 see: [error codes](https://openfeature.dev/specification/types#error-code)
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -484,7 +484,7 @@ If the provider has failed to initialize, the `provider status` should indicate 
 The client defaults and returns the `PROVIDER_NOT_READY` `error code` if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
 The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
 
-see: [error codes](https://openfeature.dev/specification/types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
+see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
 
 #### Requirement 1.7.7
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -426,7 +426,8 @@ stateDiagram-v2
     STALE --> NOT_READY:shutdown()
     ERROR --> NOT_READY:shutdown()
     READY --> RECONCILING:::client:setContext()
-    RECONCILING:::client --> READY:*
+    RECONCILING:::client --> READY
+    RECONCILING:::client --> ERROR
 
     classDef client fill:#888
 ```
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -197,7 +197,7 @@ class MyProvider implements Provider {
 > If the provider's `initialize` function fails to render the provider ready to evaluate flags, it **SHOULD** abnormally terminate.
 
 If a provider is unable to start up correctly, it should indicate abnormal execution by throwing an exception, returning an error, or otherwise indicating so by means idiomatic to the implementation language.
-If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) you can use the `PROVIDER_FATAL` error code.
+If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) the `PROVIDER_FATAL` error code should be used.
 
 see: [error codes](../types.md#error-code)
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -197,7 +197,7 @@ class MyProvider implements Provider {
 > If the provider's `initialize` function fails to render the provider ready to evaluate flags, it **SHOULD** abnormally terminate.
 
 If a provider is unable to start up correctly, it should indicate abnormal execution by throwing an exception, returning an error, or otherwise indicating so by means idiomatic to the implementation language.
-If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) you can use the `PROVIDER_STATE_FATAL` error code.
+If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) you can use the `PROVIDER_FATAL` error code.
 
 see: [error codes](../types.md#error-code)
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -188,53 +188,15 @@ class MyProvider implements Provider {
 }
 ```
 
-#### Requirement 2.4.2
+#### Condition 2.4.2
 
-> The `provider` **MAY** define a `status` field/accessor which indicates the readiness of the provider, with possible values `NOT_READY`, `READY`, `STALE`, or `ERROR`.
+> The provider defines an `initialize` function.
 
-Providers without this field can be assumed to be ready immediately.
+##### Conditional Requirement 2.4.2.1
 
-The diagram below illustrates the possible states and transitions of the `status` fields.
+> If the provider's `initialize` function fails to ready the provider to evaluate flags, it **MUST** abnormally terminate.
 
-```mermaid
----
-title: Provider State
----
-stateDiagram-v2
-    direction LR
-    [*] --> NOT_READY
-    NOT_READY --> READY:initialize
-    READY --> ERROR
-    ERROR --> READY
-    READY --> STALE
-    STALE --> READY
-    STALE --> ERROR
-    READY --> NOT_READY:shutdown
-    STALE --> NOT_READY:shutdown
-    ERROR --> NOT_READY:shutdown
-```
-
-see [provider status](../types.md#provider-status)
-
-#### Requirement 2.4.3
-
-> The provider **MUST** set its `status` field/accessor to `READY` if its `initialize` function terminates normally.
-
-If the provider supports the `status` field/accessor and initialization succeeds, setting the `status` to `READY` indicates that the provider is initialized and flag evaluation is proceeding normally.
-
-#### Requirement 2.4.4
-
-> The provider **MUST** set its `status` field to `ERROR` if its `initialize` function terminates abnormally.
-
-If the provider supports the `status` field/accessor and initialization fails, setting the `status` to `ERROR` indicates the provider is in an error state. If the error is transient in nature (ex: a connectivity failure of some kind) the provider can attempt to resolve this state automatically.
-
-#### Requirement 2.4.5
-
-> The provider **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
-
-It's recommended to set an informative `error code`, such as `PROVIDER_NOT_READY` if evaluation in attempted before the provider is initialized.
-
-see: [error codes](https://openfeature.dev/specification/types#error-code)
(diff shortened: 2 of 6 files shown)


specification/sections/01-flag-evaluation.md
```diff
@@ -400,3 +400,73 @@ The global API object might expose a `shutdown` function, which will call the re
...
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+> The implementation uses the static-context paradigm.
+
+see: [static-context paradigm](../glossary.md#static-context-paradigm)
+
+##### Conditional Requirement 1.7.2.1
+
+> In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `CONTEXT_PENDING`.
```
beeme1mr, line 450: I found the status `CONTEXT_PENDING` confusing at first because it sounds like the provider is waiting for context. Perhaps something like `RECONCILING` would make it clear that the provider is attempting to update its internal state.

specification/sections/01-flag-evaluation.md
```diff
@@ -400,3 +400,73 @@ The global API object might expose a `shutdown` function, which will call the re
...
+
+> The client's `provider status` accessor **MUST** indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.
+
+If the provider has failed to initialize, the `provider status` should indicate the provider is in an error state.
+
+#### Requirement 1.7.5
+
+> The client **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
+
+The SDK should return an informative `error code`, such as `PROVIDER_NOT_READY` if evaluation in attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
```
beeme1mr, line 470:
```suggestion
The SDK should return an informative `error code`, such as `PROVIDER_NOT_READY`, if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
```

specification/sections/02-providers.md
```diff
@@ -188,53 +188,15 @@ class MyProvider implements Provider {
...
+#### Condition 2.4.2
 
-> The `provider` **MAY** define a `status` field/accessor which indicates the readiness of the provider, with possible values `NOT_READY`, `READY`, `STALE`, or `ERROR`.
+> The provider defines an `initialize` function.
 
-Providers without this field can be assumed to be ready immediately.
+##### Conditional Requirement 2.4.2.1
 
-The diagram below illustrates the possible states and transitions of the `status` fields.
+> If the provider's `initialize` function fails to ready the provider to evaluate flags, it **MUST** abnormally terminate.
```
beeme1mr, line 197: I'm not sure what this section means. Could you please rephrase?

specification/sections/05-events.md
```diff
@@ -28,9 +28,16 @@ graph
 
 > The `provider` **MAY** define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with a `provider event details` payload. 
 
+Providers cannot emit `PROVIDER_CONTEXT_CHANGED` or `PROVIDER_CONTEXT_PENDING` event. These are emitted only by the SDK during context reconciliation.
```
beeme1mr, line 31:
```suggestion
Providers cannot emit `PROVIDER_CONTEXT_CHANGED` or `PROVIDER_CONTEXT_PENDING` event.
These are emitted only by the SDK during context reconciliation.
```

specification/sections/05-events.md
```diff
@@ -28,9 +28,16 @@ graph
...
+
 If available, native event-emitter or observable/observer language constructs can be used.
 
-see: [provider event types](../types.md#provider-events), [`event details`](../types.md#provider-event-details).
+When a provider is unable to evaluate flags (perhaps due to loss of connection with a remote service) the provider can signal this by emitting a `PROVIDER_ERROR` event.
+If the error state is irrecoverable, the `PROVIDER_STATE_FATAL` error code can be used.
```
beeme1mr, line 36: Is it possible to set `PROVIDER_STATE_FATAL` during initialization? This may happen if the access key is invalid.

specification/sections/05-events.md
```diff
@@ -147,22 +154,25 @@ See [provider initialization](./02-providers.md#24-initialization), [setting a p
 
 ### Event handlers and context reconciliation
 
-Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
+Providers built to conform to the static context paradigm feature two additional events: `PROVIDER_CONTEXT_CHANGE_PENDING` and `PROVIDER_CONTEXT_CHANGED`.
+While the provider is reconciling it's internal state (the `on context changed` function is running and not yet terminated) the SDK emits `PROVIDER_CONTEXT_CHANGE_PENDING` and transitions the provider into state `CONTEXT_PENDING`.
```
beeme1mr, line 158:
```suggestion
While the provider is reconciling its internal state (the `on context changed` function is running and not yet terminated), the SDK emits `PROVIDER_CONTEXT_CHANGE_PENDING` and transitions the provider into state `CONTEXT_PENDING`.
```

specification/sections/05-events.md
```diff
@@ -175,10 +185,10 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
 
 ##### Conditional Requirement 5.3.4.1
 
-> When the provider's `on context changed` is called, the provider **MAY** emit the `PROVIDER_STALE` event, and transition to the `STALE` state.
+> While the provider's `on context changed` function is executing, associated `PROVIDER_CONTEXT_PENDING` handlers **MUST** run.
 
-Some providers cache evaluated flags, and re-evaluate them when the context is changed.
-In these cases, the provider may signal its cache is invalid with the `PROVIDER_STALE` event and the `STALE` provider state.
+The implementation must run any `PROVIDER_CONTEXT_PENDING` handlers associated with the provider while the provider is reconciling its state.
+In languages with asynchronous semantics, the emission of this event can be skipped if the `on context changed` function of the provider in question executes synchronously for a given provider, no other operations can take place while it runs.
```
beeme1mr, line 191: Why are events skipped in `on context changed` is synchronous?

specification/sections/05-events.md
```diff
@@ -197,4 +207,22 @@ see: [provider event types](../types.md#provider-events), [provider events](#51-
...
+The SDK must update it's internal representation of the provider's state accordingly:
+
+| Event                            | Associated Status                                                  |
+| -------------------------------- | ------------------------------------------------------------------ |
+| `PROVIDER_READY`                 | `READY`                                                            |
+| `PROVIDER_STALE`                 | `STALE`                                                            |
+| `PROVIDER_ERROR`                 | `ERROR`                                                            |
+| `PROVIDER_CONFIGURATION_CHANGED` | N/A (provider remains in state `READY`)                            |
+| `PROVIDER_CONTEXT_CHANGED`       | N/A (only emitted by SDK during context reconciliation) |
+| `PROVIDER_CONTEXT_PENDING`       | N/A (only emitted by SDK during context reconciliation) |
```
beeme1mr, line 226: Please format the table.

specification/types.md
```diff
@@ -110,6 +110,7 @@ An enumerated error code represented idiomatically in the implementation languag
 | TYPE_MISMATCH         | The type of the flag value does not match the expected type.                                |
 | TARGETING_KEY_MISSING | The provider requires a targeting key and one was not provided in the `evaluation context`. |
 | INVALID_CONTEXT       | The `evaluation context` does not meet provider requirements.                               |
+| PROVIDER_STATE_FATAL  | The provider has entered a irrecoverable error state.                                       |
```
beeme1mr, line 113:
```suggestion
| PROVIDER_STATE_FATAL  | The provider has entered an irrecoverable error state.                                      |
```

beeme1mr: Do we need `STATE` in the error code?

specification/sections/05-events.md
```diff
@@ -147,22 +154,25 @@ See [provider initialization](./02-providers.md#24-initialization), [setting a p
 
 ### Event handlers and context reconciliation
 
-Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
+Providers built to conform to the static context paradigm feature two additional events: `PROVIDER_CONTEXT_CHANGE_PENDING` and `PROVIDER_CONTEXT_CHANGED`.
+While the provider is reconciling it's internal state (the `on context changed` function is running and not yet terminated) the SDK emits `PROVIDER_CONTEXT_CHANGE_PENDING` and transitions the provider into state `CONTEXT_PENDING`.
+This can be particularly useful for displaying loading indicators while the [evaluation context](./03-evaluation-context.md) is being reconciled.
+
+If the `on context changed` function terminates normally, the SDK emits (`PROVIDER_CONTEXT_CHANGED`) and transitions the provider into the `READY` state, otherwise it emits `PROVIDER_ERROR` and transitions the provider into `ERROR` state.
```
fabriziodemaria, line 161: How do we handle cases where `setEvaluationContext()` is called a second time by the application before the first call has not finished yet? Should the first `onContextChange()` call terminate normally or abnormally, and what are the implications with the Events emitted and the State changes?

specification/sections/01-flag-evaluation.md
toddbaert: I'm not crazy about PENDING either, and I considered `RECONCILING`, but worried it was too vague, but I'm interested in more opinions. Do you think we'd need to include the work `CONTEXT`? Need it be `CONTEXT_RECONCILING`? Or do you think `RECONCILING` is sufficient?

We should also change the event accordingly, I think? So the event would be `PROVIDER_RECONCILING`.

cc @fabriziodemaria

specification/sections/05-events.md
toddbaert: Yes, that's one of the main purposes. I can mention it in the init section as well.

specification/sections/05-events.md
toddbaert: If it's synchronous, there's no point emitting a pending event. This would be for providers that don't need to do any async context reconciliation, they just do some local computation. In languages like JS, while that's happening, nothing else can happen, so there' no point in emitting the event. Put another way, the context was reconciled in "real time", so it was never "pending" to the outside world - we'd just skip right to "changed".

In JS, this would look like the `on context change` function of the provider didn't return a promise but just did some synchronous work (returned void). Events would be useless or even potentially cause a disruptive UI flash.

I just wanted to provide this escape hatch here. I'm hesitant to go into more detail in the spec, but I wanted to give implementations the ability to skip this emission because I think it makes sense.

specification/types.md
toddbaert: I wen't back and forth on this exact point, I can go either way.

specification/sections/05-events.md
toddbaert: I've done this.

specification/sections/05-events.md
toddbaert: That's a great question.

My first thoughts are that we shouldn't terminate abnormally, and should emit events as normal. We may want to consider adding requirements on providers to make sure the function is reentrant, so at least there's no accidental errors if this happens.

We could also guarantee this doesn't happen by queueing up context-changes, but I think that's putting quite a burden on the SDK for a bit of an edge case. Seeing that `setContext` is an async operation in most implementations, I think authors inherently expect that multiple calls might not finish in the order they are started, and so should do what they can to avoid such race conditions.

A middle ground might be to never emit `PROVIDER_CONTEXT_CHANGED` or change the state from `CONTEXT_PENDING` while another reconciliation is in progress, which seems reasonable, so at least if the function is called twice, we wouldn't pop in and out of the pending state.

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,73 @@ The global API object might expose a `shutdown` function, which will call the re
...
+    STALE --> READY*
+    STALE --> ERROR*
+    READY --> NOT_READY:shutdown()
+    STALE --> NOT_READY:shutdown()
+    ERROR --> NOT_READY:shutdown()
+```
+
+\* transitions occurring when associated events are spontaneously emitted from the provider
+
+Note that SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define additional states to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
```
lukas-reining, line 433:
````suggestion
```mermaid
stateDiagram-v2
    direction LR
    [*] --> NOT_READY
    NOT_READY --> READY:initialize()
    NOT_READY --> ERROR:initialize()
    READY --> ERROR:*
    ERROR --> READY:*
    READY --> STALE:*
    STALE --> READY:*
    STALE --> ERROR:*
    READY --> NOT_READY:shutdown()
    STALE --> NOT_READY:shutdown()
    ERROR --> NOT_READY:shutdown()
    READY --> CONTEXT_PENDING:::optional:setContext()
    CONTEXT_PENDING --> READY:*

    classDef optional fill:#999
```

\* transitions occurring when associated events are spontaneously emitted from the provider

Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `CONTEXT_PENDING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
````

I would not make all the `STATE*` states a different state. In the rendered diagram they look like separate, final states. I changed it so the * is on the edges, what do you think @toddbaert?

Additionally I feel like it would be good to have the `CONTEXT_PENDING` so I added it in another color.

The change rendered looks like this: 
```mermaid
stateDiagram-v2
    direction LR
    [*] --> NOT_READY
    NOT_READY --> READY:initialize()
    NOT_READY --> ERROR:initialize()
    READY --> ERROR:*
    ERROR --> READY:*
    READY --> STALE:*
    STALE --> READY:*
    STALE --> ERROR:*
    READY --> NOT_READY:shutdown()
    STALE --> NOT_READY:shutdown()
    ERROR --> NOT_READY:shutdown()
    READY --> CONTEXT_PENDING:::optional:setContext()
    CONTEXT_PENDING --> READY:*

    classDef optional fill:#999
```

specification/sections/01-flag-evaluation.md
lukas-reining: I feel like `CONTEXT_RECONCILING` feels like the context itself reconciles.
Maybe `RECONCILING` would be the best then.

specification/sections/05-events.md
```diff
@@ -147,22 +155,25 @@ See [provider initialization](./02-providers.md#24-initialization), [setting a p
 
 ### Event handlers and context reconciliation
 
-Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
+Providers built to conform to the static context paradigm feature two additional events: `PROVIDER_CONTEXT_CHANGE_PENDING` and `PROVIDER_CONTEXT_CHANGED`.
+While the provider is reconciling its internal state (the `on context changed` function is running and not yet terminated), the SDK emits `PROVIDER_CONTEXT_CHANGE_PENDING` and transitions the provider into state `CONTEXT_PENDING`.
+This can be particularly useful for displaying loading indicators while the [evaluation context](./03-evaluation-context.md) is being reconciled.
+
+If the `on context changed` function terminates normally, the SDK emits (`PROVIDER_CONTEXT_CHANGED`) and transitions the provider into the `READY` state, otherwise it emits `PROVIDER_ERROR` and transitions the provider into `ERROR` state.
+The `PROVIDER_CONTEXT_CHANGED` is used to signal that the associated context has been changed, and flags should be re-evaluated.
```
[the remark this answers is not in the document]
toddbaert: > If we change to `RECONCILED` we should also change to something like `PROVIDER_RECONCILED`.

We could, and I consider that as well, but this would be a breaking change to any of the implementing client providers (including web). I felt that this wasn't worth it, but I could be convinced otherwise.

specification/sections/03-evaluation-context.md
```diff
@@ -120,12 +120,12 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
+The SDK implementation must run the `on context changed` function on all registered provider that use the global `evaluation context` whenever it is mutated.
 
 ##### Conditional Requirement 3.2.4.2
 
-> When the `evaluation context` for a specific provider is set, the `on context changed` handler **MUST** only run on the associated provider.
+> When the `evaluation context` for a specific provider is set, the `on context changed` function **MUST** only run on the associated provider.
 
-The SDK implementation must run the `on context changed` handler only on the provider that is scoped to the mutated `evaluation context`.
\ No newline at end of file
+The SDK implementation must run the `on context changed` function only on the provider that is scoped to the mutated `evaluation context`.
```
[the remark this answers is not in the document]
toddbaert: I only did this because "handlers" are specifically what we call the functions that application-authors attach to events, while the `on context change` that providers implement is something else. I think this just removes ambiguity.

specification/sections/01-flag-evaluation.md
toddbaert: Oh, I didn't notice that, nice catch. It's just because I added the * at the last moment and it rendered as a different state. Accidental!

toddbaert: > Additionally I feel like it would be good to have the CONTEXT_PENDING so I added it in another color. This way we have all information at one place here.

Love it!!!

specification/sections/05-events.md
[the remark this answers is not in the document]
lukas-reining: Ah no sorry I did not consider that this already existed.
I would also leave it then.

specification/sections/01-flag-evaluation.md
toddbaert: OK I think I will make this change then.

lukas-reining (approved):
Changes look good to me, minding the open conversations! :)

beeme1mr:
I think we can reduce the implementation burden on provider developers and ensure greater consistency by leveraging the provider state in a few additional ways. If we add a `FATAL` state to the SDK, we could short circuit evaluations within the SDK itself. Hooks would still run, but the underlying provider would not be called. The SDK could return the value specified by the developer within the evaluation itself. The reason returned to the user would be `ERROR`, the error type would be `FATAL`, and the message would be whatever the provider defined (e.g. 'invalid auth token`). This would be useful because many existing feature flag SDKs will return a default value on error, which may look like a successful evaluation in OpenFeature unless the provider explicitly handles this scenario. I believe this will lead to inconsistent behavior across implementations unless the SDK handles it.

It may be worth performing the same short circuiting within the SDK if the provider isn't ready yet. This wouldn't only affect providers that have implemented an initialization function that hasn't completed prior to a flag evaluation. This would provide many of the same benefits as described above.

Finally, the error events should also include error code. Currently, only the error message is defined in the spec.

What do you think?
@liran2000 @kinyoklion @lukas-reining @thomaspoignant @fabriziodemaria @toddbaert @jonathannorris

liran2000:
> I think we can reduce the implementation burden on provider developers and ensure greater consistency by leveraging the provider state in a few additional ways. If we add a `FATAL` state to the SDK, we could short circuit evaluations within the SDK itself. Hooks would still run, but the underlying provider would not be called. The SDK could return the value specified by the developer within the evaluation itself. The reason returned to the user would be `ERROR`, the error type would be `FATAL`, and the message would be whatever the provider defined (e.g. 'invalid auth token`). This would be useful because many existing feature flag SDKs will return a default value on error, which may look like a successful evaluation in OpenFeature unless the provider explicitly handles this scenario. I believe this will lead to inconsistent behavior across implementations unless the SDK handles it.
> 
> It may be worth performing the same short circuiting within the SDK if the provider isn't ready yet. This wouldn't only affect providers that have implemented an initialization function that hasn't completed prior to a flag evaluation. This would provide many of the same benefits as described above.
> 
> Finally, the error events should also include error code. Currently, only the error message is defined in the spec.
> 
> What do you think? @liran2000 @kinyoklion @lukas-reining @thomaspoignant @fabriziodemaria @toddbaert @jonathannorris

Sounds good to me

specification/sections/03-evaluation-context.md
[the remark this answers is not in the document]
lukas-reining: Okay fine to me.

specification/sections/02-providers.md
```diff
@@ -188,53 +188,18 @@ class MyProvider implements Provider {
...
-
-If the provider supports the `status` field/accessor and initialization fails, setting the `status` to `ERROR` indicates the provider is in an error state. If the error is transient in nature (ex: a connectivity failure of some kind) the provider can attempt to resolve this state automatically.
-
-#### Requirement 2.4.5
-
-> The provider **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
-
-It's recommended to set an informative `error code`, such as `PROVIDER_NOT_READY` if evaluation in attempted before the provider is initialized.
+If a provider is unable to start up correctly, it should indicate abnormal execution by throwing an exception, returning an error, or otherwise indicating so by means idiomatic to the implementation language.
+If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) you can use the `PROVIDER_STATE_FATAL` error code.
```
Kavindu-Dodan, line 200: I assume the implementation here will come in a form of an exception type (ex:-`ProviederStateFatalExpection.java`) ?

specification/types.md
```diff
@@ -128,12 +129,15 @@ This structure is populated by a provider for use by an [Application Author](./g
...
-| ERROR     | The provider is initialized but is not able to reliably resolve flag values.                        |
-| STALE     | The provider's cached state is no longer valid and may not be up-to-date with the source of truth.  |
+| Status           | Explanation                                                                                        |
+| ---------------- | -------------------------------------------------------------------------------------------------- |
+| NOT_READY        | The provider has not been initialized.                                                             |
+| READY            | The provider has been initialized, and is able to reliably resolve flag values.                    |
+| ERROR            | The provider is initialized but is not able to reliably resolve flag values.                       |
+| STALE            | The provider's cached state is no longer valid and may not be up-to-date with the source of truth. |
+| CONTEXT_PENDING* | The provider is reconciling its state with a context change.                                       |
```
fabriziodemaria, line 138: Should `CONTEXT_PENDING` also be an error code for Evaluation API? Once a new context is passed into `setEvaluationContext()` I'd expect the Evaluation APIs to not return values associated to the older context and rather force "default" until reconciliation is complete

maxveldink (approved):
These changes make sense to me and would simplify the providers we've written thus far 👍🏻 

One thing to clarify (I might be missing the convo somewhere): the SDK is maintaining the state of the given provider instance for a domain that can span multiple clients, right? If the same provider instance is given to multiple domains, should those each maintain separate state? Or is that an edge-case/unsupported behavior?

lukas-reining:
> It may be worth performing the same short circuiting within the SDK if the provider isn't ready yet. This wouldn't only affect providers that have implemented an initialization function that hasn't completed prior to a flag evaluation. This would provide many of the same benefits as described above.

This sounds good to me!

JamieSinn:
> I think we can reduce the implementation burden on provider developers and ensure greater consistency by leveraging the provider state in a few additional ways. If we add a `FATAL` state to the SDK, we could short circuit evaluations within the SDK itself. Hooks would still run, but the underlying provider would not be called. The SDK could return the value specified by the developer within the evaluation itself. The reason returned to the user would be `ERROR`, the error type would be `FATAL`, and the message would be whatever the provider defined (e.g. 'invalid auth token`). This would be useful because many existing feature flag SDKs will return a default value on error, which may look like a successful evaluation in OpenFeature unless the provider explicitly handles this scenario. I believe this will lead to inconsistent behavior across implementations unless the SDK handles it.
> 
> It may be worth performing the same short circuiting within the SDK if the provider isn't ready yet. This wouldn't only affect providers that have implemented an initialization function that hasn't completed prior to a flag evaluation. This would provide many of the same benefits as described above.
> 
> Finally, the error events should also include error code. Currently, only the error message is defined in the spec.
> 
> What do you think? @liran2000 @kinyoklion @lukas-reining @thomaspoignant @fabriziodemaria @toddbaert @jonathannorris

This I think covers the cases I was thinking about

toddbaert:
> These changes make sense to me and would simplify the providers we've written thus far 👍🏻
> 
> One thing to clarify (I might be missing the convo somewhere): the SDK is maintaining the state of the given provider instance for a domain that can span multiple clients, right? If the same provider instance is given to multiple domains, should those each maintain separate state? Or is that an edge-case/unsupported behavior?

I _think_ I lean toward leaving this unspecified. Different langauges have different sematics and paradigms WRT references, pointers, etc. I think I'd rather leave the particularities of this up to implementations. We can perhaps make some helpful suggestions in non-normative sections.

specification/types.md
toddbaert: I wouldn't consider this an error in all cases. I think I would defer to the provider on this. They could throw, or return the previous value with a custom reason (we could even add `CONTEXT_PENDING` (or whatever language we end up with) as a pre-defined reason).

Interested in more thoughts on this.

specification/types.md
toddbaert: I've gone with `PROVIDER_FATAL`, no more "STATE".

specification/sections/01-flag-evaluation.md
toddbaert: I've done this.

toddbaert:
> I think we can reduce the implementation burden on provider developers and ensure greater consistency by leveraging the provider state in a few additional ways. If we add a `FATAL` state to the SDK, we could short circuit evaluations within the SDK itself. Hooks would still run, but the underlying provider would not be called. The SDK could return the value specified by the developer within the evaluation itself. The reason returned to the user would be `ERROR`, the error type would be `FATAL`, and the message would be whatever the provider defined (e.g. 'invalid auth token`). This would be useful because many existing feature flag SDKs will return a default value on error, which may look like a successful evaluation in OpenFeature unless the provider explicitly handles this scenario. I believe this will lead to inconsistent behavior across implementations unless the SDK handles it.
> 
> It may be worth performing the same short circuiting within the SDK if the provider isn't ready yet. This wouldn't only affect providers that have implemented an initialization function that hasn't completed prior to a flag evaluation. This would provide many of the same benefits as described above.
> 
> Finally, the error events should also include error code. Currently, only the error message is defined in the spec.
> 
> What do you think? @liran2000 @kinyoklion @lukas-reining @thomaspoignant @fabriziodemaria @toddbaert @jonathannorris

It seems like there's agreement on this, and I think it's a _really nice addition_. I've added this, as well as changed the name of the reconciliation state/event from `CONTEXT_PENDING` to `RECONCILING`.

specification/sections/02-providers.md
toddbaert: > I assume the implementation here will come in the form of an exception/error type (ex:-ProviederStateFatalExpection.java) ?

Yes.

> Besides, does this mean SDK should differentiate normal error vs PROVIDER_STATE_FATAL and attempt retries. I think SDK shouldn't but this could happen at provider eventually making it to ready status isn't it ?

The provider handles it's own business concerning reconnects and such internal. If it emits a `FATAL` error, it would be logically inconsistent for it to attempt to fix that state. If it throws FATAL, it should give up, and the SDK would make that error code known to the application author/integrator as per the new section `1.7.8`.

specification/sections/01-flag-evaluation.md
```diff
@@ -423,19 +425,23 @@ stateDiagram-v2
...
+    RECONCILING:::client --> READY:*
 
-    classDef optional fill:#999
+    classDef client fill:#555
 ```
 
 \* transitions occurring when associated events are spontaneously emitted from the provider
 
-Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `CONTEXT_PENDING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
+<span style="color:#555">█</span> only defined in static-context (client-side) paradigm
```
beeme1mr, line 436: Clever 😄

specification/sections/01-flag-evaluation.md
```diff
@@ -423,19 +425,23 @@ stateDiagram-v2
...
+
+Note that only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `RECONCILING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
```
beeme1mr, line 438: Should we use the note GitHub syntax? It's now supported in the docs as well.

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,104 @@ The global API object might expose a `shutdown` function, which will call the re
...
+If the provider has failed to initialize, the `provider status` should indicate the provider is in an error state.
+
+#### Requirement 1.7.6
+
+> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.
+
+The client defaults and returns the `PROVIDER_NOT_READY` `error code` if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
+The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
+
+see: [error codes](https://openfeature.dev/specification/types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
```
beeme1mr, line 487:
```suggestion
see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
```

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,104 @@ The global API object might expose a `shutdown` function, which will call the re
...
+
+#### Requirement 1.7.7
+
+> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `PROVIDER_FATAL`.
+
+TThe client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
+The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
+
+see: [error codes](https://openfeature.dev/specification/types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
```
beeme1mr, line 496:
```suggestion
see: [error codes](../types#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
```

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,104 @@ The global API object might expose a `shutdown` function, which will call the re
...
+
+#### Requirement 1.7.8
+
+> Implementations **SHOULD** propagate the `error code` returned from any provider lifecycle methods.
+
+The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
+
+see: [error codes](https://openfeature.dev/specification/types#error-code)
```
beeme1mr, line 504:
```suggestion
see: [error codes](../types#error-code)
```

beeme1mr: Should this be in the event section?

specification/sections/02-providers.md
```diff
@@ -188,53 +188,18 @@ class MyProvider implements Provider {
...
-
-If the provider supports the `status` field/accessor and initialization fails, setting the `status` to `ERROR` indicates the provider is in an error state. If the error is transient in nature (ex: a connectivity failure of some kind) the provider can attempt to resolve this state automatically.
-
-#### Requirement 2.4.5
-
-> The provider **SHOULD** indicate an error if flag resolution is attempted before the provider is ready.
-
-It's recommended to set an informative `error code`, such as `PROVIDER_NOT_READY` if evaluation in attempted before the provider is initialized.
+If a provider is unable to start up correctly, it should indicate abnormal execution by throwing an exception, returning an error, or otherwise indicating so by means idiomatic to the implementation language.
+If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) you can use the `PROVIDER_FATAL` error code.
```
beeme1mr, line 200:
```suggestion
If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) the `PROVIDER_FATAL` error code should be used.
```

specification/sections/05-events.md
```diff
@@ -28,9 +28,17 @@ graph
 
 > The `provider` **MAY** define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with a `provider event details` payload. 
 
+Providers cannot emit `PROVIDER_CONTEXT_CHANGED` or `RECONCILING` event.
```
beeme1mr, line 31:
```suggestion
Providers cannot emit `PROVIDER_CONTEXT_CHANGED` or `PROVIDER_RECONCILING` event.
```

At least that's what the event is called in the types doc.

specification/sections/05-events.md
```diff
@@ -28,9 +28,17 @@ graph
...
+These are emitted only by the SDK during context reconciliation.
+
 If available, native event-emitter or observable/observer language constructs can be used.
 
-see: [provider event types](../types.md#provider-events), [`event details`](../types.md#provider-event-details).
+When a provider is unable to evaluate flags (perhaps due to loss of connection with a remote service) the provider can signal this by emitting a `PROVIDER_ERROR` event.
+If the error state is irrecoverable, the `PROVIDER_FATAL` error code can be used.
+When it recovers, it can emit a `PROVIDER_READY` event.
```
beeme1mr, line 38:
```suggestion
When it recovers, it can emit a `PROVIDER_READY` event.
If the error state is irrecoverable, the `PROVIDER_FATAL` error code can be used.
```

I would swap these lines because a `PROVIDER_FATAL` error code isn't recoverable but a `PROVIDER_ERROR` is.

specification/sections/05-events.md
```diff
@@ -197,4 +214,22 @@ see: [provider event types](../types.md#provider-events), [provider events](#51-
 
 The `PROVIDER_ERROR` is not emitted from the provider itself; the SDK implementation must run the `PROVIDER_ERROR` handlers if the `on context changed` throws or otherwise signals an error.
 
-see: [provider event types](../types.md#provider-events), [provider events](#51-provider-events), context, [provider context reconciliation](02-providers.md#26-provider-context-reconciliation)
\ No newline at end of file
+see: [provider event types](../types.md#provider-events), [provider events](#51-provider-events), context, [provider context reconciliation](02-providers.md#26-provider-context-reconciliation)
```
beeme1mr, line 217: Should `context` be a link?

specification/sections/05-events.md
```diff
@@ -197,4 +214,22 @@ see: [provider event types](../types.md#provider-events), [provider events](#51-
...
+| Event                            | Associated Status                                       |
+| -------------------------------- | ------------------------------------------------------- |
+| `PROVIDER_READY`                 | `READY`                                                 |
+| `PROVIDER_STALE`                 | `STALE`                                                 |
+| `PROVIDER_ERROR`                 | `ERROR`                                                 |
+| `PROVIDER_CONFIGURATION_CHANGED` | N/A (provider remains in state `READY`)                 |
+| `PROVIDER_CONTEXT_CHANGED`       | N/A (only emitted by SDK during context reconciliation) |
+| `PROVIDER_RECONCILING`           | N/A (only emitted by SDK during context reconciliation) |
+
+see [provider lifecycle management](01-flag-evaluation.md#17-provider-lifecycle-management)
```
beeme1mr, line 235:
```suggestion
see: [provider lifecycle management](01-flag-evaluation.md#17-provider-lifecycle-management)
```

specification/sections/05-events.md
```diff
@@ -197,4 +214,22 @@ see: [provider event types](../types.md#provider-events), [provider events](#51-
...
+
+Some providers may emit events spontaneously, based on changes in their internal state (connections, caches, etc).
+The SDK must update it's internal representation of the provider's state accordingly:
+
+| Event                            | Associated Status                                       |
+| -------------------------------- | ------------------------------------------------------- |
+| `PROVIDER_READY`                 | `READY`                                                 |
+| `PROVIDER_STALE`                 | `STALE`                                                 |
+| `PROVIDER_ERROR`                 | `ERROR`                                                 |
+| `PROVIDER_CONFIGURATION_CHANGED` | N/A (provider remains in state `READY`)                 |
```
beeme1mr, line 231:
```suggestion
| `PROVIDER_CONFIGURATION_CHANGED` | N/A (provider remains in its current state)                 |
```

I think this address the non-ready states like `PROVIDER_READY` and `PROVIDER_STALE`.

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,104 @@ The global API object might expose a `shutdown` function, which will call the re
...
+
+#### Requirement 1.7.5
+
+> The client's `provider status` accessor **MUST** indicate `FATAL` if the `initialize` function of the associated provider terminates abnormally and indicates `error code` `PROVIDER_FATAL`.
+
+If the provider has failed to initialize, the `provider status` should indicate the provider is in an error state.
+
+#### Requirement 1.7.6
+
+> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.
```
beeme1mr, line 482: Why is this a **SHOULD**? Perhaps **MUST** would be more appropriate.

specification/sections/01-flag-evaluation.md
toddbaert: No, IMO,  think it's actually more relevant for lifecycle methods than events, which makes me feel like here is better.

specification/sections/01-flag-evaluation.md
toddbaert: I debated this. I think we just want to decide how firm this requirement is. It seems like a really nice thing to have, but not 100% required to have a functional SDK. I could be convinced on a MUST though.

specification/sections/05-events.md
toddbaert: I think this is actually a typo, I think just linking to `provider context reconciliation` is enough.

specification/sections/02-providers.md
toddbaert: nice catch.

specification/sections/01-flag-evaluation.md
toddbaert: Ya I think that's a nice touch. Will do.

specification/sections/01-flag-evaluation.md
toddbaert: I thought so. :sunglasses:

specification/sections/01-flag-evaluation.md
toddbaert: done

specification/sections/01-flag-evaluation.md
beeme1mr: It's a bit hard to see, though. Can you try a different color?

image

toddbaert: I'll move it back to 888

maxveldink (approved):

specification/sections/01-flag-evaluation.md
lukas-reining: I was also just trying out any gray colors😅
I like what you did with the comment!

specification/sections/01-flag-evaluation.md
beeme1mr: I think it should be **MUST**. Otherwise, an SDK could be spec-compliant and behave fundamentally differently.

lukas-reining: I feel like it depends on if the client implements hooks. If so, I would assume it must.
As we say that  `The API, Client, Provider, and invocation MUST have a method for registering hooks.` , I would say the error hooks have to run.

lukas-reining (approved):
Looks good to me!

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,105 @@ The global API object might expose a `shutdown` function, which will call the re
...
+The client defaults and returns the `PROVIDER_NOT_READY` `error code` if evaluation is attempted before the provider is initialized (the provider is still in a `NOT_READY` state).
+The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
+
+see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
+
+#### Requirement 1.7.7
+
+> The client **SHOULD** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `PROVIDER_FATAL`.
+
+TThe client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
```
Kavindu-Dodan, line 494:
```suggestion
The client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
```

specification/sections/02-providers.md
Kavindu-Dodan: Yes, concern is resolved with the updated requirement.

Kavindu-Dodan (approved):

specification/sections/01-flag-evaluation.md
toddbaert: I've used 888, which is in the middle and looks OK in both IMO

specification/types.md
fabriziodemaria: I agree on deferring to the Provider on this.
In the cases where the Provider wants to return the default value during reconciliation, we could suggest to use `ERROR` reason with `PROVIDER_NOT_READY` error code (this is how we handle it now in our [Kotlin Confidence Provider](https://github.com/spotify/confidence-openfeature-provider-kotlin/pull/123/files), for example). In this case, we can expand the `PROVIDER_NOT_READY` error code's description:
```
The provider has not been initialized, or a context reconciliation is in progress.
```
Perhaps better to discuss this minor detail on separate PR, and happy to resolve this comment thread.

specification/sections/05-events.md
```diff
@@ -147,22 +161,25 @@ See [provider initialization](./02-providers.md#24-initialization), [setting a p
 
 ### Event handlers and context reconciliation
 
-Providers built to conform to the static context paradigm feature an additional `PROVIDER_CONTEXT_CHANGED` event, which is used to signal that the global context has been changed, and flags should be re-evaluated.
+Providers built to conform to the static context paradigm feature two additional events: `PROVIDER_CONTEXT_CHANGE_PENDING` and `PROVIDER_CONTEXT_CHANGED`.
```
fabriziodemaria, line 164:
```suggestion
Providers built to conform to the static context paradigm feature two additional events: `PROVIDER_RECONCILING` and `PROVIDER_CONTEXT_CHANGED`.
```
This is the naming used in the "types" table. I see a couple more instances of `PROVIDER_CONTEXT_CHANGE_PENDING` throughout the PR, which I think we need to update

toddbaert: Nice catch. I will make sure these are all resolved.

toddbaert:
rebased to resolve a 1-word conflict.

specification/sections/05-events.md
toddbaert: Thanks!

beeme1mr (approved):

toddbaert:
I'd like to merge this early next week if possible. Let me know if you have any concerns!

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,105 @@ The global API object might expose a `shutdown` function, which will call the re
...
+<span style="color:#888">█</span> only defined in static-context (client-side) paradigm
+
+> [!NOTE]
+> Only SDKs implementing the [static context (client-side) paradigm](../glossary.md#static-context-paradigm) define `RECONCILING` to facilitate [context reconciliation](./02-providers.md#26-provider-context-reconciliation).
+
+#### Requirement 1.7.1
+
+> The `client` **MUST** define a `provider status` accessor which indicates the readiness of the associated provider, with possible values `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`.
+
+The SDK at all times maintains an up-to-date state corresponding to the success/failure of the last lifecycle method (`initialize`, `shutdown`, `on context change`) or emitted event.
```
fabriziodemaria, line 445: Since the SDK will anyway have to monitor Events coming from the Providers for certain State transitions, wouldn't it be simpler to have Events from the Provider be the _only_ driver for State transitions? Events can also provide more nuanced information about what is happening in the Provider, rather than being constrained to waiting for a returning/throwing function call

specification/sections/01-flag-evaluation.md
```diff
@@ -401,3 +401,105 @@ The global API object might expose a `shutdown` function, which will call the re
...
+    READY --> ERROR:*
+    ERROR --> READY:*
+    READY --> STALE:*
+    STALE --> READY:*
+    STALE --> ERROR:*
+    READY --> NOT_READY:shutdown()
+    STALE --> NOT_READY:shutdown()
+    ERROR --> NOT_READY:shutdown()
+    READY --> RECONCILING:::client:setContext()
+    RECONCILING:::client --> READY:*
```
justinabrahms, line 429: Reconciling can never error?

specification/sections/01-flag-evaluation.md
toddbaert: > Since the SDK will anyway have to monitor Events coming from the Providers for certain State transitions, wouldn't it be simpler to have Events from the Provider be the _only_ driver for State transitions?

I think the ergonomics here might be confusing, and I think it will lead to some state corruption. For example, if there's an error during a provider's `initialize()` function, it will have to catch it and then emit an error event. I think it's likely authors will not handle all such errors, in which case the event won't be emitted. If instead we pay attention to lifecycle methods like `initialize()`, no matter what happens, our state is preserved, since a thrown error will result in the `error` state.

On the other hand, events that are emitted spontaneously from a provider also need to be handled (for example, a provider suddenly loses a connection or goes STALE and emits a corresponding event).

So I think we need to handle both of these.

> Events can also provide more nuanced information about what is happening in the Provider, rather than being constrained to waiting for a returning/throwing function call

I'm not sure that's entirely the case, since the OpenFeature errors and error event payloads basically contain the same information, after this change

specification/sections/05-events.md
andreas-karlsson: Is this really an edge-case? I would imagine apps call setContext from event handlers (say a dark-mode toggle) which would quickly lead to overlapping reconciliation cycles, and the spec should be clear on how to handle these.

Also, from an experimentation perspective, it's adamant that the state remain in RECONCILING until the last initiated cycle has ended. If the state transitions to READY before the last cycle has ended, flags will return stale values (for example values for another user). If the state transitions to ERROR before the last cycle has ended, flags will return default values. Both of these behaviours will taint the user experience and the experiment.

specification/sections/01-flag-evaluation.md
fabriziodemaria: Good points. If the need arise, we'll have more opportunities to massage the contract between SDK and Provider in smaller iterations moving forward. I suggest we just resolve this comment thread 👍

specification/sections/05-events.md
toddbaert: > Is this really an edge-case? I would imagine apps call setContext from event handlers (say a dark-mode toggle) which would quickly lead to overlapping reconciliation cycles, and the spec should be clear on how to handle these.
> 
> Also, from an experimentation perspective, it's adamant that the state remain in RECONCILING until the last initiated cycle has ended. If the state transitions to READY before the last cycle has ended, flags will return stale values (for example values for another user). If the state transitions to ERROR before the last cycle has ended, flags will return default values. Both of these behaviours will taint the user experience and the experiment.

@andreas-karlsson you make some compelling points.

We could keep things vague and just say SDKs must "handle" this in some way... but I'm thinking it's better to prescribe a pattern to handle it.

Do you think if we spec'd something like what I described [here](https://github.com/open-feature/spec/pull/241#discussion_r1479925311) (specifically the quoted section below) it would be sufficient?

> A middle ground might be to never emit PROVIDER_CONTEXT_CHANGED or change the state from CONTEXT_PENDING while another reconciliation is in progress, which seems reasonable, so at least if the function is called twice, we wouldn't pop in and out of the pending state.

cc @fabriziodemaria

andreas-karlsson: Thanks, that would be great.

toddbaert: @andreas-karlsson @fabriziodemaria I've pushed a commit to address this specific issue: https://github.com/open-feature/spec/pull/241/commits/ad3b33768f9de746b9c766164b186b62d7dcaddc

specification/sections/01-flag-evaluation.md
toddbaert: I've added this.

specification/sections/01-flag-evaluation.md
```diff
@@ -426,7 +426,8 @@ stateDiagram-v2
     STALE --> NOT_READY:shutdown()
     ERROR --> NOT_READY:shutdown()
     READY --> RECONCILING:::client:setContext()
-    RECONCILING:::client --> READY:*
+    RECONCILING:::client --> READY
```
beeme1mr, line 432: Did you intend to remove the asterisk?

toddbaert: Yes this is not emitted spontaneously from the provider, it's part of the lifecycle (triggered by SDK events).

fabriziodemaria (approved):

nicklasl (approved):

toddbaert:
I'll merge this by EOD today unless I hear objections.

PR #245, opened 2024-02-14
feat: adds allFlagMetadata to clients and provider interface

maxveldink:
## This PR
Adds an optional `allFlagMetadata` accessor to clients and the provider interface for retrieving a list of flag keys available in the provider.

- Adds requirements for `allFlagMetadata` accessor
- Defines `Collection` and `Provider Metadata` types in the Types page and adds key information to `Flag Metadata`.

### Related Issues
[Discussion](https://github.com/orgs/open-feature/discussions/338) that predated this proposal.

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -87,7 +87,7 @@
         {
             "id": "Requirement 1.2.3",
             "machine_id": "requirement_1_2_3",
-            "content": "The client interface MUST define a `all flag metadata` member or accessor, returning a collection of flag metadata from the provider.",
+            "content": "The client interface MUST define a `all flag metadata` member or accessor, containing a `flags` field or accessor which is a collection of flag metadata, from the provider.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -368,7 +368,7 @@
         {
             "id": "Requirement 2.1.2",
             "machine_id": "requirement_2_1_2",
-            "content": "The provider MAY define an `all flag metadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.",
+            "content": "The provider MAY define an `all flag metadata` field or accessor, containing a `flags` field or accessor which is a collection of flag metadata, identifying all available flag keys in the provider.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -85,6 +85,13 @@
             "children": []
         },
         {
+            "id": "Requirement 1.2.3",
+            "machine_id": "requirement_1_2_3",
+            "content": "The client interface MUST define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
             "id": "Condition 1.3.1",
             "machine_id": "condition_1_3_1",
             "content": "The implementation uses the dynamic-context paradigm.",
@@ -361,7 +368,7 @@
         {
             "id": "Requirement 2.1.2",
             "machine_id": "requirement_2_1_2",
-            "content": "The provider `metadata` member or accessor MAY define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md",
+            "content": "The provider MAY define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -359,6 +359,28 @@
             "children": []
         },
         {
+            "id": "Requirement 2.1.2",
+            "machine_id": "requirement_2_1_2",
+            "content": "The provider `metadata` member or accessor MAY define an `allFlagKeys` field or accessor of type collection of strings, which identifies all available flag keys in the provider.",
+            "RFC 2119 keyword": "MAY",
+            "children": []
+        },
+        {
+            "id": "Condition 2.1.3",
+            "machine_id": "condition_2_1_3",
+            "content": "The implementing language has an asynchronous support.",
+            "RFC 2119 keyword": null,
+            "children": [
+                {
+                    "id": "Conditional Requirement 2.1.3.1",
+                    "machine_id": "conditional_requirement_2_1_3_1",
+                    "content": "The accessor for `allFlagKeys` SHOULD be asynchronous.",
+                    "RFC 2119 keyword": "SHOULD",
+                    "children": []
+                }
+            ]
+        },
+        {
             "id": "Requirement 2.2.1",
             "machine_id": "requirement_2_2_1",
             "content": "The `feature provider` interface MUST define methods to resolve flag values, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required) and `evaluation context` (optional), which returns a `resolution details` structure.",
--- a/specification.json
+++ b/specification.json
@@ -373,21 +373,6 @@
             "children": []
         },
         {
-            "id": "Condition 2.1.3",
-            "machine_id": "condition_2_1_3",
-            "content": "The implementing language has an asynchronous support.",
-            "RFC 2119 keyword": null,
-            "children": [
-                {
-                    "id": "Conditional Requirement 2.1.3.1",
-                    "machine_id": "conditional_requirement_2_1_3_1",
-                    "content": "The accessor for `allFlagMetadata` SHOULD be asynchronous.",
-                    "RFC 2119 keyword": "SHOULD",
-                    "children": []
-                }
-            ]
-        },
-        {
             "id": "Requirement 2.2.1",
             "machine_id": "requirement_2_2_1",
             "content": "The `feature provider` interface MUST define methods to resolve flag values, with parameters `flag key` (string, required), `default value` (boolean | number | string | structure, required) and `evaluation context` (optional), which returns a `resolution details` structure.",
--- a/specification.json
+++ b/specification.json
@@ -368,7 +368,7 @@
         {
             "id": "Requirement 2.1.2",
             "machine_id": "requirement_2_1_2",
-            "content": "The provider MAY define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md",
+            "content": "The provider MAY define an `allFlagMetadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -361,7 +361,7 @@
         {
             "id": "Requirement 2.1.2",
             "machine_id": "requirement_2_1_2",
-            "content": "The provider `metadata` member or accessor MAY define an `allFlagKeys` field or accessor of type collection of strings, which identifies all available flag keys in the provider.",
+            "content": "The provider `metadata` member or accessor MAY define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
@@ -374,7 +374,7 @@
                 {
                     "id": "Conditional Requirement 2.1.3.1",
                     "machine_id": "conditional_requirement_2_1_3_1",
-                    "content": "The accessor for `allFlagKeys` SHOULD be asynchronous.",
+                    "content": "The accessor for `allFlagMetadata` SHOULD be asynchronous.",
                     "RFC 2119 keyword": "SHOULD",
                     "children": []
                 }
--- a/specification.json
+++ b/specification.json
@@ -87,7 +87,7 @@
         {
             "id": "Requirement 1.2.3",
             "machine_id": "requirement_1_2_3",
-            "content": "The client interface MUST define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.",
+            "content": "The client interface MUST define a `all flag metadata` member or accessor, returning a collection of flag metadata from the provider.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -368,7 +368,7 @@
         {
             "id": "Requirement 2.1.2",
             "machine_id": "requirement_2_1_2",
-            "content": "The provider MAY define an `allFlagMetadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.",
+            "content": "The provider MAY define an `all flag metadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -65,7 +65,7 @@ This function not only sets the provider, but ensures that the provider is ready
 // default provider
 OpenFeatureAPI.getInstance().setProviderAndWait(myprovider); // this method blocks until the provider is ready or in error
 // client uses the default provider
-Client client = OpenFeatureAPI.getInstance().getClient(); 
+Client client = OpenFeatureAPI.getInstance().getClient();
 
 // provider associated with domain-1
 OpenFeatureAPI.getInstance().setProviderAndWait('domain-1', myprovider); // this method blocks until the provider is ready or in error
@@ -170,6 +170,16 @@ client.getMetadata().getDomain(); // "domain-1"
 
 In previous drafts, this property was called `name`.
 For backwards compatibility, implementations should consider `name` an alias to `domain`.
+
+#### Requirement 1.2.3
+
+> The client interface **MUST** define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.
+
+```typescript
+client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
+```
+
+This may return an empty collection if the provider does not implement the `allFlagMetadata` member or accessor.
 
 ### 1.3. Flag Evaluation
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -179,7 +179,7 @@ For backwards compatibility, implementations should consider `name` an alias to 
 client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
 ```
 
-This may return an empty collection if the provider does not implement the `allFlagMetadata` member or accessor.
+This may return a language-idiomatic way of describing the absence of a value if the provider does not implement the `allFlagMetadata` member or accessor.
 
 ### 1.3. Flag Evaluation
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -173,7 +173,7 @@ For backwards compatibility, implementations should consider `name` an alias to 
 
 #### Requirement 1.2.3
 
-> The client interface **MUST** define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.
+> The client interface **MUST** define a `all flag metadata` member or accessor, returning a collection of flag metadata from the provider.
 
 ```typescript
 client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -173,10 +173,10 @@ For backwards compatibility, implementations should consider `name` an alias to 
 
 #### Requirement 1.2.3
 
-> The client interface **MUST** define a `all flag metadata` member or accessor, returning a collection of flag metadata from the provider.
+> The client interface **MUST** define a `all flag metadata` member or accessor, containing a `flags` field or accessor which is a collection of flag metadata, from the provider.
 
 ```typescript
-client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
+client.getAllFlagMetadata() // { flags: [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}] }
 ```
 
 This may return a language-idiomatic way of describing the absence of a value if the provider does not implement the `allFlagMetadata` member or accessor.
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -34,13 +34,7 @@ provider.getMetadata().getName(); // "my-custom-provider"
 provider.getAllFlagMetadata(); // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
 ```
 
-#### Condition 2.1.3
-
-> The implementing language has an asynchronous support.
-
-##### Conditional Requirement 2.1.3.1
-
-> The accessor for `allFlagMetadata` **SHOULD** be asynchronous.
+For some providers, fetching all flags is an expensive operation. If there is language support, consider implementing this as an asynchronous operation.
 
 ### 2.2 Flag Value Resolution
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -28,13 +28,13 @@ provider.getMetadata().getName(); // "my-custom-provider"
 
 #### Requirement 2.1.2
 
-> The provider **MAY** define an `allFlagMetadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.
+> The provider **MAY** define an `all flag metadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.
 
 ```typescript
 provider.getAllFlagMetadata(); // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
 ```
 
-This operation should not be confused with bulk evaluation of all flags.
+This operation should not be confused with a bulk evaluation of all flags.
 This is an informative operation available to clients for understanding which flag keys are currently active from a provider.
 Example usages include building a debug screen based on available flags and detecting stale flag evaluations.
 For some providers, fetching all flags is an expensive operation. If there is language support, consider implementing this as an asynchronous operation.
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -34,6 +34,9 @@ provider.getMetadata().getName(); // "my-custom-provider"
 provider.getAllFlagMetadata(); // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
 ```
 
+This operation should not be confused with bulk evaluation of all flags.
+This is an informative operation available to clients for understanding which flag keys are currently active from a provider.
+Example usages include building a debug screen based on available flags and detecting stale flag evaluations.
 For some providers, fetching all flags is an expensive operation. If there is language support, consider implementing this as an asynchronous operation.
 
 ### 2.2 Flag Value Resolution
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -26,6 +26,22 @@ Providers are the "translator" between the flag evaluation calls made in applica
 provider.getMetadata().getName(); // "my-custom-provider"
 ```
 
+#### Requirement 2.1.2
+
+> The provider `metadata` member or accessor **MAY** define an `allFlagKeys` field or accessor of type collection of strings, which identifies all available flag keys in the provider.
+
+```typescript
+provider.getMetadata().getAllFlagKeys(); // ['featureA', 'featureB']
+```
+
+#### Condition 2.1.3
+
+> The implementing language has an asynchronous support.
+
+##### Conditional Requirement 2.1.3.1
+
+> The accessor for `allFlagKeys` **SHOULD** be asynchronous.
+
 ### 2.2 Flag Value Resolution
 
 `Providers` are implementations of the `feature provider` interface, which may wrap vendor SDKs, REST API clients, or otherwise resolve flag values from the runtime environment.
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -38,6 +38,7 @@ This operation should not be confused with a bulk evaluation of all flags.
 This is an informative operation available to clients for understanding which flag keys are currently active from a provider.
 Example usages include building a debug screen based on available flags and detecting stale flag evaluations.
 For some providers, fetching all flags is an expensive operation. If there is language support, consider implementing this as an asynchronous operation.
+If the provider does not implement this behavior or the flag metadata cannot currently be retrieved due to an error status, the language-idiomatic way of describing the absence of a value can be returned from this method.
 
 ### 2.2 Flag Value Resolution
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -28,10 +28,10 @@ provider.getMetadata().getName(); // "my-custom-provider"
 
 #### Requirement 2.1.2
 
-> The provider `metadata` member or accessor **MAY** define an `allFlagKeys` field or accessor of type collection of strings, which identifies all available flag keys in the provider.
+> The provider `metadata` member or accessor **MAY** define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md#flag-metadata), which identifies all available flag keys in the provider.
 
 ```typescript
-provider.getMetadata().getAllFlagKeys(); // ['featureA', 'featureB']
+provider.getMetadata().getAllFlagMetadata(); // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
 ```
 
 #### Condition 2.1.3
@@ -40,7 +40,7 @@ provider.getMetadata().getAllFlagKeys(); // ['featureA', 'featureB']
 
 ##### Conditional Requirement 2.1.3.1
 
-> The accessor for `allFlagKeys` **SHOULD** be asynchronous.
+> The accessor for `allFlagMetadata` **SHOULD** be asynchronous.
 
 ### 2.2 Flag Value Resolution
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -28,10 +28,10 @@ provider.getMetadata().getName(); // "my-custom-provider"
 
 #### Requirement 2.1.2
 
-> The provider `metadata` member or accessor **MAY** define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md#flag-metadata), which identifies all available flag keys in the provider.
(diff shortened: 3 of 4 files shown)


specification/sections/02-providers.md
```diff
@@ -26,6 +26,22 @@ Providers are the "translator" between the flag evaluation calls made in applica
...
+```
+
+#### Condition 2.1.3
+
+> The implementing language has an asynchronous support.
+
+##### Conditional Requirement 2.1.3.1
+
+> The accessor for `allFlagKeys` **SHOULD** be asynchronous.
+
```
maxveldink, line 44: I am unsure if we need to call this out, but it felt important, given how expensive this call might be. I'm happy to remove this if it's overly restrictive.

specification/types.md
maxveldink, line 1: We didn't have a call out for the Provider Metadata or Collection types. We have other references to collections in the spec, so I took a stab at defining what it is. I am happy for language suggestions here or thoughts on whether I should not define these in this PR.

specification/sections/02-providers.md
weyert: I think it's a good idea to call out. Makes it more obvious and clear imho

specification/sections/02-providers.md
```diff
@@ -26,6 +26,22 @@ Providers are the "translator" between the flag evaluation calls made in applica
 provider.getMetadata().getName(); // "my-custom-provider"
 ```
 
+#### Requirement 2.1.2
+
+> The provider `metadata` member or accessor **MAY** define an `allFlagKeys` field or accessor of type collection of strings, which identifies all available flag keys in the provider.
+
+```typescript
+provider.getMetadata().getAllFlagKeys(); // ['featureA', 'featureB']
```
justinabrahms, line 34: I may be misunderstanding the use-case here, but I believe these also need typing information, else you won't know which method to call to get the value out, right?

kinyoklion: I am also having a bit of a hard time understanding the ergonomics of this. This does satisfy one category of usage, "does the flag I plan on evaluating exist", but it doesn't provide a great means of getting all of the values associated with those keys.

There are a couple use cases I am aware of for evaluating all the flags.

1. Bootstrap. You evaluate all the flags so you can use them for something like SSR. You render the values into the page and you avoid pop-in. Or you store them into a cache when you build your app for the same reason.

2. You evaluate them all and then you pluck values out of the map, versus evaluating them individually. One useful example of this is to ensure that all your flags are consistent with each other for some transactional period, like handling an http request. (The flags could change part way through the request, and if the flags have relationships, then that isn't ideal.)

For me the second one is somewhat of an anti-pattern, or shows an underlying weakness of a system. Once you do that, then you lose insight into what flags are actually being evaluated, which causes problems for tracking exposure for things like an experiment.

kinyoklion: I am not sure how we get from the keys, to either of these use-cases, easily.

maxveldink: Those are great points, especially for how we'd go from keys to the correct resolution method. I'm hoping to avoid adding a complete bulk evaluation right now. Based on the discussion predating this, we want to try a more minimal approach at first.

We could be more explicit about some required fields for flag metadata, including key and type. Would that be sufficient information if we return a collection of flag metadata through the provider metadata?

We still wouldn't meet the bootstrapping use case described by @kinyoklion, but we would be able to

1. Detect stale flags being evaluated and
2. Build a debug screen based on available flags

which are two use cases we've encountered.

specification/types.md
```diff
@@ -122,8 +126,19 @@ A structure containing the following fields:
 
 A structure which supports definition of arbitrary properties, with keys of type `string`, and values of type `boolean`, `string`, or `number`.
 
+`key` is a required property with a value of `string`.
+`type` is a required property with a `string` value of either `boolean`, `string`, `number`, or `structure`.
```
federicobond, line 130: How should languages that disambiguate number types like Go or Python handle `number`-typed flags?

beeme1mr, line 130: Perhaps this should be an optional property. It's nice to have but may not be easily available to all providers.

maxveldink: My thought was that this would indicate which client method would be used to resolve the flag. Should we call out that this value simply needs to map to an available client resolution method?

maxveldink: I'm okay with it being optional 👍🏻

specification/sections/02-providers.md
```diff
@@ -26,6 +26,22 @@ Providers are the "translator" between the flag evaluation calls made in applica
 provider.getMetadata().getName(); // "my-custom-provider"
 ```
 
+#### Requirement 2.1.2
+
+> The provider `metadata` member or accessor **MAY** define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md#flag-metadata), which identifies all available flag keys in the provider.
+
+```typescript
+provider.getMetadata().getAllFlagMetadata(); // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
```
liran2000, line 34: 1. There were some [discussions](https://cloud-native.slack.com/archives/C0344AANLA1/p1698258230035399?thread_ts=1698251286.355879&cid=C0344AANLA1) for fetching all flags with a specific tag and similar actions. Maybe adding an optional/required context parameter which can be empty, so providers can implement it ?
2. Why the method is under metadata, and not directly on the provider ?

beeme1mr: An optional method on the provider may be preferred because the values could be calculated upon request. It would also make it easier to add support for this feature in providers that would need to make an external API call.

It's worth noting that end users are not supposed to interact directly with a provider. An accessor will need to be added to the [Flag Evaluation API](https://openfeature.dev/specification/sections/flag-evaluation) as well.

specification/sections/02-providers.md
```diff
@@ -26,6 +26,22 @@ Providers are the "translator" between the flag evaluation calls made in applica
...
+```
+
+#### Condition 2.1.3
+
+> The implementing language has an asynchronous support.
+
+##### Conditional Requirement 2.1.3.1
+
+> The accessor for `allFlagMetadata` **SHOULD** be asynchronous.
```
beeme1mr, line 43: This requirement would change how we treat metadata, at least in the JS SDK. Currently, metadata is static data that conforms to an interface.

https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/provider/provider.ts#L33-L35

beeme1mr (approved):
Thanks for getting this started and sorry for the delayed review. Please review my comments when you have a moment and let me know if you have any questions. Thanks!

specification/sections/02-providers.md
liran2000: Can it be updated to _getAllFlagMetadata_(context)?

beeme1mr: Why would you need context? This should be metadata about the flags and not actually a bulk evaluation API.

liran2000: For example, to get just flags with a specific tag. This question was raised sometimes. Having context parameter can let the provider an option to support it.

beeme1mr: What's the use case for doing that? The original proposal was to return a list of flags available within the SDK that could be used for debugging and simple admin UIs. I think anything more complex than that would not work consistently across implementations, and using vendor APIs directly may be the better option.

liran2000: Some vendors have attributes for flags, like hidden etc' at their admin UI. Use case can be to get all non-hidden flags. Also to know more about the flag, can consider returning optional metadata per flag, like hidden or not.
These are some cases to provide optional flexibility, if not needed for now, it can be fine without it.

Kavindu-Dodan:
It might be too late to consider, but there was an old OFEP [1] [2] on flag metadata, which is similar to the use cases that attempt to address through this spec change. 

From OFEP pov, `getFlagMetadata(KEY)` & `getAllFlagMetadata()` were new APIs of OF client. 

[1] - https://github.com/open-feature/ofep/pull/44 
[2] - https://github.com/Kavindu-Dodan/ofep/blob/460112b8585e045b2e8fd4bfba83eb7cb22d0990/OFEP-flag-metadata-to-applications.md

specification/sections/02-providers.md
Kavindu-Dodan, line 34: It might be too late to consider, but there was an old OFEP [1] [2] on flag metadata, which is similar to the use cases that attempt to address through this spec change. 

From OFEP pov, `getFlagMetadata(KEY)` & `getAllFlagMetadata()` were new APIs of OF client. 

[1] - https://github.com/open-feature/ofep/pull/44 
[2] - https://github.com/Kavindu-Dodan/ofep/blob/460112b8585e045b2e8fd4bfba83eb7cb22d0990/OFEP-flag-metadata-to-applications.md

specification/sections/02-providers.md
maxveldink: That is a good point; having it on the provider interface directly removes the condition. I wouldn't want to change the static nature of metadata with this change, and I think there's a reasonable expectation that this call would be dynamic for each call.

I'm going to add it instead to the provider interface, as well as expose it to the client and see how that looks.

specification/sections/02-providers.md
maxveldink: Oh cool, thank you for linking these, I hadn't seen them. Yeah, I agree with you and @beeme1mr that we probably should move those to be exposed on the client so that users don't need to interact directly with the provider. Let me make that change and see how it looks.

maxveldink: I would defer passing in any context information that the provider can use to filter flag keys for now. We do allow for arbitrary properties on the flag metadata that is returned from the call, so there could be some client-side filtering that could be done or directly calling the provider in that given use case if the evaluation must be done server-side.

specification.json
```diff
@@ -294,6 +301,28 @@
             "RFC 2119 keyword": "MUST",
             "children": []
         },
+        {
+            "id": "Requirement 2.1.2",
+            "machine_id": "requirement_2_1_2",
+            "content": "The provider MAY define an `allFlagMetadata` field or accessor of type collection of [flag_metadata](../types.md",
```
thomaspoignant, line 307: Not sure if it is a real problem, but the markdown link here is cut in half.

beeme1mr: Good catch @thomaspoignant. Markdown should be avoided in the requirements but work fine in the non-normative sections.

Kavindu-Dodan (approved):

Kavindu-Dodan (commented):
Please consider updating the title to match the recent changes

federicobond:
One thing that worries me a bit about this specification is that many vendors will have paginated APIs for retrieving lists of flags, and there is no guidance on how they should implement this method.

Should they consume the whole collection in N requests in order to ensure correctness?

lukas-reining:
> One thing that worries me a bit about this specification is that many vendors will have paginated APIs for retrieving lists of flags, and there is no guidance on how they should implement this method.
>
> Should they consume the whole collection in N requests in order to ensure correctness?

Good point @federicobond.
On the other hand a provider may just decide to not implement metadata if pagination is too much of an hurdle as cited in the following.
A provider may just decide to fetch all pages or simply not implement `getAllMetadata` if either the information is not available or implementations is just not good e.g. if the endpoints are expensive or just hard to iterate.

> Requirement 1.2.3
> The client interface **MUST** define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.
>```typescript
>client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
>```
>
>This may return an empty collection if the provider does not implement the `allFlagMetadata` member or accessor.

specification/sections/01-flag-evaluation.md
```diff
@@ -171,6 +171,16 @@ client.getMetadata().getDomain(); // "domain-1"
...
 
+#### Requirement 1.2.3
+
+> The client interface **MUST** define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.
+
+```typescript
+client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
+```
+
+This may return an empty collection if the provider does not implement the `allFlagMetadata` member or accessor.
```
lukas-reining, line 182: I am not sure about this one @maxveldink as I would assume something like an optional, null, undefined, or error code if the provider does not implement it.
Having an empty collections would mean that semantically, "no flags available" is the same as "flag metadata" not implemented. 
As a consumer this could be problematic because one might have to implement different behavior if `allFlagsMetadata` is not implemented compared to just no flags being available.

toddbaert: One of the rules we have is that none of the client methods should throw - the client is the primary abstraction the `application author` interfaces with - it's the most likely abstraction to be used in request handlers and other critical paths at runtime.

I think we want to allow implementations a lot of flexibility here, but I would suggest a null or undefined return that's clearly annotated.

lukas-reining: > I think we want to allow implementations a lot of flexibility here, but I would suggest a null or undefined return that's clearly annotated.

Yes thats exactly what I was suggesting @toddbaert. Maybe for the wording it could be something like `a language idiomatic way of describing the absence of a value` to kept the implementation open null, undefined, options, maybes, etc.

specification.json
```diff
@@ -358,6 +365,28 @@
...
+            "children": [
+                {
+                    "id": "Conditional Requirement 2.1.3.1",
+                    "machine_id": "conditional_requirement_2_1_3_1",
+                    "content": "The accessor for `allFlagMetadata` SHOULD be asynchronous.",
+                    "RFC 2119 keyword": "SHOULD",
+                    "children": []
+                }
+            ]
+        },
```
toddbaert, line 389: I'm not 100% sure this is necessary to specify. I think most language implementations will make the right choice here, especially if you mentioned something like this in a non-normative section.

specification/sections/02-providers.md
toddbaert: I think it would be worth adding additional non-normative text explaining the purpose of this feature and some example supported use-cases.

specification.json
maxveldink: I was on the fence about making this a full requirement. I like the idea of doing this as non-normative text 👍🏻

specification/sections/01-flag-evaluation.md
maxveldink: Not throwing makes sense to me. Creating the distinction between a `null` or `undefined` value being returned vs a collection would be enough information for the caller to decide what to do. Unlike flag evaluation, since not all providers will implement this behavior, callers cannot have a hard requirement on the return value (or should always handle the sad path, regardless of the underlying issue).

I've updated the language here to match @lukas-reining 's suggestion. Let me know how that looks!

specification/sections/02-providers.md
maxveldink: I took an initial stab at including more text around this feature. Let me know of any edits/clarifications I can make there.

specification/sections/01-flag-evaluation.md
```diff
@@ -171,6 +171,16 @@ client.getMetadata().getDomain(); // "domain-1"
 In previous drafts, this property was called `name`.
 For backwards compatibility, implementations should consider `name` an alias to `domain`.
 
+#### Requirement 1.2.3
+
+> The client interface **MUST** define a `allFlagMetadata` member or accessor, returning a collection of flag metadata from the provider.
```
toddbaert, line 176:
```suggestion
> The client interface **MUST** define a `all flag metadata` member or accessor, returning a collection of flag metadata from the provider.
```

nit, see style guide: https://github.com/open-feature/spec?tab=readme-ov-file#style-guide

specification/sections/02-providers.md
```diff
@@ -26,6 +26,19 @@ Providers are the "translator" between the flag evaluation calls made in applica
 provider.getMetadata().getName(); // "my-custom-provider"
 ```
 
+#### Requirement 2.1.2
+
+> The provider **MAY** define an `allFlagMetadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.
```
toddbaert, line 31:
```suggestion
> The provider **MAY** define an `all flag metadata` field or accessor of type collection of flag_metadata, which identifies all available flag keys in the provider.
```

nit, see style guide: https://github.com/open-feature/spec?tab=readme-ov-file#style-guide

specification/types.md
```diff
@@ -123,8 +127,18 @@ A structure containing the following fields:
 
 A structure which supports definition of arbitrary properties, with keys of type `string`, and values of type `boolean`, `string`, or `number`.
 
+`key` is a required property with a value of `string`.
```
toddbaert, line 130:
```suggestion
`key` is a required property with a value of type `string`.
```

specification/types.md
```diff
@@ -123,8 +127,18 @@ A structure containing the following fields:
...
+`type` is an optional property with a `string` value indicating which client resolution method should be used to resolve the flag key.
```
toddbaert, line 131:
```suggestion
`type` is an optional property with a value of type `string`, indicating which client resolution method should be used to resolve the flag key.
```

specification/types.md
```diff
@@ -123,8 +127,18 @@ A structure containing the following fields:
...
+
 This structure is populated by a provider for use by an [Application Author](./glossary.md#application-author) via the [Evaluation API](./glossary.md#evaluation-api) or an [Application Integrator](./glossary.md#application-integrator) via [hooks](./sections/04-hooks.md).
 
+### Provider Metadata
+
+A structure which carries information about the provider.
+In addition to the defined metadata fields below, arbitrary information can be stored here unique to the provider.
+
+`name` is a required key with a string value.
```
toddbaert, line 140:
```suggestion
`name` is a required key with a value of type string.
```

toddbaert (commented):
Left minor nits, but largely ready to approve; I'm interested in getting some more feedback, though. @kinyoklion  and @federicobond seem like they might have some reservations.

specification/sections/02-providers.md
```diff
@@ -26,6 +26,19 @@ Providers are the "translator" between the flag evaluation calls made in applica
...
+
+```typescript
+provider.getAllFlagMetadata(); // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
+```
+
+This operation should not be confused with bulk evaluation of all flags.
```
beeme1mr, line 37:
```suggestion
This operation should not be confused with a bulk evaluation of all flags.
```

beeme1mr (approved):

lukas-reining (approved):

specification/sections/01-flag-evaluation.md
```diff
@@ -171,6 +171,16 @@ client.getMetadata().getDomain(); // "domain-1"
 In previous drafts, this property was called `name`.
 For backwards compatibility, implementations should consider `name` an alias to `domain`.
 
+#### Requirement 1.2.3
+
+> The client interface **MUST** define a `all flag metadata` member or accessor, returning a collection of flag metadata from the provider.
+
+```typescript
+client.getAllFlagMetadata() // [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
```
beeme1mr, line 179: I wonder if this should be an object that contains a list of flags.

```ts
{
  flags: [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}]
}
```

That would allow us to easily add new metadata to the response in a non-breaking way. For example, a last updated timestamp and version property would be extremely useful.

maxveldink: Makes sense to me, I'll make that update and see what others think.

beeme1mr:
Let's try and get this merged next week. If anyone has any concerns, please speak up so they can be addressed.

luizgribeiro (approved):

luizgribeiro:
I have a few minor concerns about this.

```
This may return a language-idiomatic way of describing the absence of a value if the provider does not implement the `allFlagMetadata` member or accessor.
```

I get that in some languages it's straight forward, but in others it would be open to the provider to determine how to express that it does not implement this method. Eg: In Java there could be one provider that throws and another one that returns an empty `Collection`. Any idea on how to deal with this?
There's also an issue with empty `Collections` imho. It would be ambiguous because user's apps wouldn't be able to determine if this is not implemented by the provider or there's no flag configured.

My first thought in this situation is to return a struct with flag's metadata and details.

Regarding the provider's backend being overwhelmed  by this sort of request, there could be something like a flag metadata query for `getAllFlagMetadata`. If flags store metadata about itself (like their bounded context) the App could get exactly what it needs. It's probably not in the scope of this proposal though

maxveldink:
Thanks @luizgribeiro ! @lukas-reining has a similar concern [here](https://github.com/open-feature/spec/pull/245#discussion_r1501681193), which I tried to alleviate by having the provider interface optionally implement the behavior. In contrast, the client interface always needs to implement the behavior, with the option of returning `null` if the all flag metadata behavior is unavailable, for any reason.

Is it worthwhile to clarify that the client should handle erroring states and prevent them from bubbling up if something goes wrong in the provider?

luizgribeiro:
> In contrast, the client interface always needs to implement the behavior, with the option of returning null if the all flag metadata behavior is unavailable, for any reason.

That's where I think there might be a way to express provider state clearly to the app.
Using a struct/map it's possible to communicate what happened and the app can handle scenarios as needed. @federicobond's statement [(here)](https://github.com/open-feature/spec/pull/245#discussion_r1502778094 )gave us some details and scenarios in which it would be nice to differentiate the result.

So, using something like the following would be an alternative.

```ts
type AllFlagsMetadataReturn = {
  status: 'OK' | 'REQUEST_ERROR';
  data: Array<{
    flag: string;
    metadata: 'string' | 'bool' | 'int' | 'float'
  }>;
};
```

toddbaert:
> > In contrast, the client interface always needs to implement the behavior, with the option of returning null if the all flag metadata behavior is unavailable, for any reason.
> 
> That's where I think there might be a way to express provider state clearly to the app. Using a struct/map it's possible to communicate what happened and the app can handle scenarios as needed. @federicobond's statement [(here)](https://github.com/open-feature/spec/pull/245#discussion_r1502778094)gave us some details and scenarios in which it would be nice to differentiate the result.
> 
> So, using something like the following would be an alternative.
> 
> ```ts
> type AllFlagsMetadataReturn = {
>   status: 'OK' | 'REQUEST_ERROR';
>   data: Array<{
>     flag: string;
>     metadata: 'string' | 'bool' | 'int' | 'float'
>   }>;
> };
> ```

@luizgribeiro in the latest (unreleased) spec version, there's an independent provider state accessor that must be available, see [1.7.1](https://github.com/open-feature/spec/blob/37cf68b0d68b6814514bcded521b9e199efcead3/specification.json#L291-L295) and proceeding points.

kinyoklion:
> Left minor nits, but largely ready to approve; I'm interested in getting some more feedback, though. @kinyoklion and @federicobond seem like they might have some reservations.

I can probably accept this specification, in that I cannot see the immediate harm that it causes, but I am still not confident in use cases.

specification/types.md
```diff
@@ -26,6 +26,10 @@ A numeric value of unspecified type or size. Implementation languages may furthe
 
 Structured data, presented however is idiomatic in the implementation language, such as JSON or YAML.
 
+### Collection
+
+A list or series of similar data types, presented however is idiomatic in the implementation language.
```
justinabrahms, line 31: Is this an ordered thing? Is it worth calling it out?

specification/types.md
```diff
@@ -123,8 +127,18 @@ A structure containing the following fields:
 
 A structure which supports definition of arbitrary properties, with keys of type `string`, and values of type `boolean`, `string`, or `number`.
 
+`key` is a required property with a value of type `string`.
+`type` is an optional property with a value of type `string`, indicating which client resolution method should be used to resolve the flag key.
```
justinabrahms, line 131: I don't think this is spelled out, but we might want to.   `key` and `type` are reserved properties on the Flag Metadata object.

specification/types.md
maxveldink: This is a fairly generic definition since I didn't want to place too many constraints on the collection type returned per language. If order matters for a collection, I would think we would call that out in the requirement/section where it was applicable versus the blanket unordered callout. But I can add that if it's important to state here!

maxveldink:
@luizgribeiro, I see what you're saying about the return type. I worry that we'd be adding too much complexity to the provider method if we introduced a new type to model potential error situations from the provider. I think allowing the provider method to return `null`, for now, would give enough information to the client to know if the flags returned are empty versus unavailable.

I added one more callout on the provider spec to clarify that more.

maxveldink:
I believe I've addressed all feedback on this PR now. I'll wait until tomorrow morning and merge this if there is no other feedback!

lukas-reining (approved):

specification/sections/02-providers.md
```diff
@@ -26,6 +26,20 @@ Providers are the "translator" between the flag evaluation calls made in applica
...
+#### Requirement 2.1.2
+
+> The provider **MAY** define an `all flag metadata` field or accessor, containing a `flags` field or accessor which is a collection of flag metadata, identifying all available flag keys in the provider.
+
+```typescript
+provider.getAllFlagMetadata(); // { flags: [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}] }
+```
+
+This operation should not be confused with a bulk evaluation of all flags.
+This is an informative operation available to clients for understanding which flag keys are currently active from a provider.
```
dferber90, line 38: The method is planned to be named `getAllFlagMetadata` but the description mentions this returns "currently active" flags only. Is "currently active" defined further somewhere?

weyert: Good point, worth clarifying this. My personal understanding is that it refers to the enabled feature flags in the provider

maxveldink: Thanks for catching that! I know `active` is semantically significant for most vendors. I tried to clarify this by changing it to `available`, which can be interpreted different per provider.

weyert: @dferber90 I hope it clarifies it. What do you think of this enhancement?

jonathannorris:
Hey @maxveldink what's the motivation behind having the `type` in the data model? 

I think we might have a better chance of constraining the usage of this feature if the data model was simpler and more restrictive:

for example instead of: 
```{ flags: [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}] }```

A simpler mapping of just `flag key` -> `feature key` would less likely be used beyond just as metadata:

```{ "key_1": "featureA", "key_2": "featureB"}```

**Comment by dferber90** [comment]
The way our [Feature Flag toolbar integration](https://vercel.com/blog/toolbar-feature-flags) works is that we query the various providers and map their API responses into this shape

```ts
interface FlagOptionType {
  value: any;
  label?: string;
}

interface FlagDefinitionType {
  options?: FlagOptionType[];
  /**
   * The URL where the feature flag can be managed.
   */
  origin?: string;
  description?: string;
}
 
type ProviderData = {
  definitions: Record<string, FlagDefinitionType>;;
};
```

Being able to see the available options & description lets us build a rich UI for overwriting flags, where we can display the available flags, their descriptions, their options and the label of each option if present.

image

toddbaert:
> Hey @maxveldink, what's the motivation behind having the `type` in the data model?
> 
> I think we might have a better chance of constraining the usage of this to its intended use-case if the data model was simpler and more restrictive:
> 
> for example, instead of: `{ flags: [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}] }`
> 
> A simpler mapping of just `flag key` -> `feature key` would less likely be used beyond just as metadata:
> 
> `{ "key_1": "featureA", "key_2": "featureB"}`

My take is generally "less is more" but I'm not sure if anyone's envisioned use-case requires the type info. If not, I think we should drop it, but I'm wondering it if could be helpful. Some backends don't strictly type flags. cc @maxveldink @jonathannorris  @dferber90

toddbaert:
> The way our [Feature Flag toolbar integration](https://vercel.com/blog/toolbar-feature-flags) works is that we query the various providers and map their API responses into this shape
> 
> ```ts
> interface FlagOptionType {
>   value: any;
>   label?: string;
> }
> 
> interface FlagDefinitionType {
>   options?: FlagOptionType[];
>   /**
>    * The URL where the feature flag can be managed.
>    */
>   origin?: string;
>   description?: string;
> }
>  
> type ProviderData = {
>   definitions: Record<string, FlagDefinitionType>;;
> };
> ```
> 
> Being able to see the available options & description lets us build a rich UI for overwriting flags, where we can display the available flags, their descriptions, their options and the label of each option if present.
> 
> image

I guess "label" would correspond to our `key` prop? `description` could be something supplied loosely in the metadata if I'm understanding correctly.

Is the proposed API sufficient then? Or is there anything we're missing? What is `value`? We don't evaluate these flags in our case so if it's the flag value I'm not sure we can satisfy that.

maxveldink:
> Hey @maxveldink, what's the motivation behind having the `type` in the data model?
> 
> I think we might have a better chance of constraining the usage of this to its intended use-case if the data model was simpler and more restrictive:
> 
> for example, instead of: `{ flags: [{"key": "featureA", "type": "boolean"}, {"key": "featureB", "type": "string"}] }`
> 
> A simpler mapping of just `flag key` -> `feature key` would less likely be used beyond just as metadata:
> 
> `{ "key_1": "featureA", "key_2": "featureB"}`

I originally had a more straightforward collection like this, just really a list of keys. The need for a type comes from client evaluation. Callers need to know which client evaluation method to call for a given flag key, so we need some type information there to understand which client method to call.

maxveldink:
@toddbaert @dferber90 Yep, `description` could be returned in the `flag metadata` type that we have defined. That can largely be left up to the provider/underyling vendor

jonathannorris:
> ## This PR
> Adds an optional `allFlagMetadata` accessor to clients and the provider interface for retrieving a list of flag keys available in the provider.
> 
> * Adds requirements for `allFlagMetadata` accessor
> * Defines `Collection` and `Provider Metadata` types in the Types page and adds key information to `Flag Metadata`.
> 
> ### Related Issues
> [Discussion](https://github.com/orgs/open-feature/discussions/338) that predated this proposal.

I'm weary about bringing this up and don't want to block this PR further, but it sounds like you would be better served by a bulk flag evaluation method. That is generally what our customers would use for this use case. I usually imagine "metadata" like this should primarily be used for analytics/tracking purposes, not really for having enough information to properly make a request for an individual flag evaluation.

weyert:
Bulk evaluation costs money (e.g. 10x normal single flag evaluation) while fetching a list of flags available does not

kinyoklion:
> Bulk evaluation costs money (e.g. 10x normal single flag evaluation) while fetching a list of flags available does not

Currently the only way to provide this data for a LaunchDarkly provider would be via a bulk-evaluation, which is part of the reason I am dubious about use cases. For the LaunchDarkly case the primary cost will be in the time the evaluations takes.

toddbaert:
I think there's some value in, to use the language of the PR:

> understanding which flag keys are currently available from a provider

I can imagine situations where this could be useful for debugging, administrative UIs, etc; generally for tangential use-cases not typical flag evaluation as we normally consider it. I appreciate how some SDKs/providers simply don't support an equivalent right now.

kinyoklion:
> I think there's some value in, to use the language of the PR:
> 
> > understanding which flag keys are currently available from a provider
> 
> I can imagine situations where this could be useful for debugging, administrative UIs, etc; generally for tangential use-cases not typical flag evaluation as we normally consider it. I appreciate how some SDKs/providers simply don't support an equivalent right now.
> 
> I'm still not sure if @dferber90 's use case is satisfied if we don't return actual evaluated flag values though.

I do see use of that information from a debugging perspective, but I don't know about this specific way of exposing it.

If your application is evaluating flags, then hooks provide the means to know what flags are evaluated, and what values they evaluated to.

The resolution details should contain information about flags that are evaluated that don't exist.

Provider configuration change events, when they can be supported, also provide that list of flags, but potentially in an incremental way.

I don't like bulk-evaluation, but it does solve these use cases, and I suspect flag meta data will just be used to make less efficient versions of bulk-evaluation (which is worse than just bulk-evaluation).

toddbaert:
OK... I hate to throw more into the mix here, but I really don't want us to merge something we're not sure about.

I have a alternate path I'd like to propose which perhaps satisfies the original use-case brought up by @maxveldink and others, _and_ dodges the valid pitfalls @kinyoklion mentions...

What if instead of adding a method to the `client`, we put a very similar payload in the `READY` event payload (and also enhance the `PROVIDER_CONFIGURATION_CHANGED` payload to feature the same). This means that _if_ your provider support it, you could attach a `READY` handler that would get all the flags and their metadata.

This has the advantage of being isomorphic with the `PROVIDER_CONFIGURATION_CHANGED` events, improving consistency. Plus, it makes it implicitly clear that this set may change - and may need to be updated with change events; this can be easily done by adding a similar handler to those.

maxveldink:
The current proposal seems to invite misuse, and we haven't reached an entire agreement. I'm cool with closing this attempt (with gratitude for the great conversation) so we can focus the discussion on using some other affordances in OpenFeature, such as hooks and events.

The Ruby SDK has yet to implement hooks and events, so I'm not as versed in them from an implementation perspective. Can we open a new issue/discussion for @toddbaert 's proposal?

dferber90:
Seems our use case was misunderstood a bit. To us an API would be useful which answers the questions
- "Which flags does the provider offer, and what are their descriptions?"
- "For each flag the provider offers, which values can it resolve to, and what are those called if they have a name (label)?"

Imagine a payload like

```
{
  "myFlag": {
    "description": "A feature flag",
    "origin": "https://example.com/myFlag",
    "options": [
      { "value": false, "label": "off" },
      { "value": true, "label": "on" }
    ]
  }
}

```

> What is value? We don't evaluate these flags in our case so if it's the flag value I'm not sure we can satisfy that.

Value is not the value the flag resolved to, but rather what it _could_ resolve to. We turn this list of possible values into a dropdown where customers can select one of the possible values to override a feature flag.

kinyoklion:
> Seems our use case was misunderstood a bit. To us an API would be useful which answers the questions
> 
> * "Which flags does the provider offer, and what are their descriptions?"
> * "For each flag the provider offers, which values can it resolve to, and what are those called if they have a name (label)?"
> 
> Imagine a payload like
> 
> ```json
> {
>   "myFlag": {
>     "description": "A feature flag",
>     "origin": "https://example.com/myFlag",
>     "options": [
>       { "value": false, "label": "off" },
>       { "value": true, "label": "on" }
>     ]
>   }
> }
> ```
> 
> > What is value? We don't evaluate these flags in our case so if it's the flag value I'm not sure we can satisfy that.
> 
> Value is not the value the flag resolved to, but rather what it _could_ resolve to. We turn this list of possible values into a dropdown where customers can select one of the possible values to override a feature flag.
> 
> Loading this kind of information usually requires a different API key, which needs to be kept secret. I'm not familiar enough with OpenFeature yet to know whether this would be in scope of OpenFeature, or whether OpenFeature is purely concerned with the "resolving feature flags" part.

This seems like a much greater level of detail than a provider would have in most cases. (Especially for single context paradigm where the available options are often not communicated to the client at all.)

I don't know what the requirements would be across many vendors, but for LaunchDarkly I know that it would require a different type of credential to get this kind of information than the credentials by SDKs.

It also seems like a debug level of feature, not a production feature? Something that maybe would lend itself to having some configuration to allow usage that could be easily disabled in production?

maxveldink:
As mentioned above, closing for now. Thanks all for the conversation!

PR #260, opened 2024-05-09
feat: tracking

toddbaert:
- Adds section 6: tracking
  - no normative sections
  - overview
  - goals/non-goals (temporary)
  
  This is not final! Please review goals/non-goals especially, and add any questions you might have.

### DIFF
--- a/specification/sections/06-tracking.md
+++ b/specification/sections/06-tracking.md
@@ -20,6 +20,7 @@ Vendors and custom solutions often support a _tracking_ or _goal measuring_ API 
 - Specify a flexible API widely compatible with basic vendor functionality
   - Define tracking event payload
   - Define tracking event identifier
+  - Support A/B testing and experimentation use-cases
   - Support client and server paradigms
   - Provide recommendations around: 
     - Async vs sync
@@ -30,9 +31,8 @@ Vendors and custom solutions often support a _tracking_ or _goal measuring_ API 
 
 - Creating an experimentation platform
 - Covering every user-tracking use case
-  - Focus on A/B testing, and experimentation based on feature flags
-  - Focus on "facts" not "metrics"
-    - Simple point in time data, no aggregation
+  - We will not define any data aggregation mechanisms
+  - We will not focus on "metrics", but instead, "facts"
 
 ### Design Principles
 
--- a/specification/sections/06-tracking.md
+++ b/specification/sections/06-tracking.md
@@ -1,0 +1,40 @@
+---
+title: Tracking
+description: Specification defining tracking API
+toc_max_heading_level: 4
+---
+
+# 6. Tracking
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+## Overview
+
+Experimentation is a primary use case for feature flags.
+In practice, this often means flags variants are assigned to users at random or in accordance to a business rule, while the impact of the assigned variant on some business objective is measured.
+Vendors and custom solutions often support a _tracking_ or _goal measuring_ API to facilitate the measurement of these business objectives.
+
+### Goals
+
+- Develop official terminology to support consistent implementation
+- Specify a flexible API widely compatible with basic vendor functionality
+  - Define tracking event payload
+  - Define tracking event identifier
+  - Support client and server paradigms
+
+### Non-goals
+
+- Creating an experimentation platform
+- Covering every user-tracking use case
+  - Focus on A/B testing, and experimentation based on feature flags
+  - Focus on "facts" not "metrics"
+    - Simple point in time data, no aggregation
+
+### Design Principles
+
+We value the following:
+
+- Adherence to, and compatibility with OpenFeature semantics
+- Maximum compatibility and ease-of-adoption for existing solutions
+- Minimum traffic and payload size
+- Ease-of-use for application authors, integrators, and provider authors (in that order)
--- a/specification/sections/06-tracking.md
+++ b/specification/sections/06-tracking.md
@@ -21,6 +21,10 @@ Vendors and custom solutions often support a _tracking_ or _goal measuring_ API 
   - Define tracking event payload
   - Define tracking event identifier
   - Support client and server paradigms
+  - Provide recommendations around: 
+    - Async vs sync
+    - Flushing mechanisms
+    - Event batching
 
 ### Non-goals
 
--- a/specification/sections/06-tracking.md
+++ b/specification/sections/06-tracking.md
@@ -11,7 +11,7 @@ toc_max_heading_level: 4
 ## Overview
 
 Experimentation is a primary use case for feature flags.
-In practice, this often means flags variants are assigned to users at random or in accordance to a business rule, while the impact of the assigned variant on some business objective is measured.
+In practice, this often means flag variants are assigned to users at random or in accordance with a business rule, while the impact of the assigned variant on some business objective is measured.
 Vendors and custom solutions often support a _tracking_ or _goal measuring_ API to facilitate the measurement of these business objectives.
 
 ### Goals


specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+---
+
+# 6. Tracking
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+## Overview
+
+Experimentation is a primary use case for feature flags.
+In practice, this often means flags variants are assigned to users at random or in accordance to a business rule, while the impact of the assigned variant on some business objective is measured.
```
beeme1mr, line 14:
```suggestion
In practice, this often means flag variants are assigned to users at random or in accordance with a business rule, while the impact of the assigned variant on some business objective is measured.
```

specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+- Develop official terminology to support consistent implementation
+- Specify a flexible API widely compatible with basic vendor functionality
+  - Define tracking event payload
+  - Define tracking event identifier
+  - Support client and server paradigms
+
+### Non-goals
+
+- Creating an experimentation platform
+- Covering every user-tracking use case
```
weyert, line 28: Considering adding as non-goal anything that relies on `rrweb` (https://www.npmjs.com/package/rrweb) such as record and replay functionality

specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+In practice, this often means flag variants are assigned to users at random or in accordance with a business rule, while the impact of the assigned variant on some business objective is measured.
+Vendors and custom solutions often support a _tracking_ or _goal measuring_ API to facilitate the measurement of these business objectives.
+
+### Goals
+
+- Develop official terminology to support consistent implementation
+- Specify a flexible API widely compatible with basic vendor functionality
+  - Define tracking event payload
+  - Define tracking event identifier
+  - Support client and server paradigms
```
federicobond, line 23: Should this include recommendations and/or SDK support for asynchronous events flushing? Most vendors support async events.

weyert: You mean able to trigger event flushing manually, alongside it being part of the `shutdown`? That make sense I think

federicobond: That, but also how to approach async in general, perhaps the SDKs could bundle a robust implementation of message batching so that individual providers are easier to write.

federicobond: My sense is that SDKs and framework integrations will have a better context to handle these than individual providers.

liran-p: async can come together with batch aggregations to reduce network requests and possibly improve performance for high rate, and also via a hook, like done in [Go Feature Flag provider](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/hook/DataCollectorHook.java).
it was described at the requirements of this example by @thomaspoignant:

>
- The data collection should be activated by default, but it should be possible to deactivate it in the configuration.
- We should not call the ... API ... for every evaluation, but call it in bulk with a periodic threshold.
- The threshold should be configurable by the user of the provider.

All these are related to implementation, so not sure whether or how to include mentioning it in spec.

federicobond: Agree, just pointing out that these choices impact the design of the provider API.

liran2000: \* fixing user

async can come together with batch aggregations to reduce network requests and possibly improve performance for high rate, and also via a hook, like done in [Go Feature Flag provider](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/go-feature-flag/src/main/java/dev/openfeature/contrib/providers/gofeatureflag/hook/DataCollectorHook.java).
it was described at the requirements of this example by @thomaspoignant:

>
- The data collection should be activated by default, but it should be possible to deactivate it in the configuration.
- We should not call the ... API ... for every evaluation, but call it in bulk with a periodic threshold.
- The threshold should be configurable by the user of the provider.

All these are related to implementation, so not sure whether or how to include mentioning it in spec.

> Agree, just pointing out that these choices impact the design of the provider API and thus the spec

specification/sections/06-tracking.md
moredip: I think it's implied

specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+title: Tracking
+description: Specification defining tracking API
+toc_max_heading_level: 4
+---
+
+# 6. Tracking
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+## Overview
```
moredip, line 11: There are two other important use cases which could be worth calling out

- flag usage analysis (for detecting obsolete flags)
- controlled rollout: making sure a flag change has no unexpected impact on metrics (which typically includes more technical than business metrics)

Not sure how much these would actually affect how the spec is defined, but as it stands the scope seems a bit too narrow

specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+- Specify a flexible API widely compatible with basic vendor functionality
+  - Define tracking event payload
+  - Define tracking event identifier
+  - Support client and server paradigms
+
+### Non-goals
+
+- Creating an experimentation platform
+- Covering every user-tracking use case
+  - Focus on A/B testing, and experimentation based on feature flags
```
moredip, line 29: I got confused whether focus on A/B testing is a non-goal

specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+  - Focus on "facts" not "metrics"
+    - Simple point in time data, no aggregation
+
+### Design Principles
+
+We value the following:
+
+- Adherence to, and compatibility with OpenFeature semantics
+- Maximum compatibility and ease-of-adoption for existing solutions
```
moredip, line 38: 💯 good to call out

specification/sections/06-tracking.md
federicobond: I believe those use cases fall under the observability initiative https://github.com/open-telemetry/community/pull/2042, but it would not hurt to check for overlap with tracking.

nicklasl: I'm totally on board with the "controlled rollout" use case but please explain a bit more about "flag usage analysis".

specification/sections/06-tracking.md
nicklasl: As an end user of an event API I would expect the API from OF for tracking an event to be just a fire and forget. But I don't know how deep we want to formulate that here?

specification/sections/06-tracking.md
moredip: Flag usage analysis: tracking which flags are no longer being evaluated, so they can be retired.

But I guess that's probably out of scope for this, because it can probably be implemented entirely within the SDK and/or provider, without needing the app dev to explicitly call anything

moredip: > I believe those use cases fall under the observability initiative [open-telemetry/community#2042](https://github.com/open-telemetry/community/pull/2042), but it would not hurt to check for overlap with tracking.

Our o11y initiative would only help to support controlled rollout if both the provider and the flag consumer are using otel, AND the flag consumer is willing to ship otel traces to the flag provider. I believe that a lot of providers will want an alternative to that - i.e. manually instrumenting the metrics that they want to observe during rollout. 

Also, the metrics which are tracked during a controlled rollout can vary from very technical (and thus likely to be measured via otel) to biz-oriented (and thus much more likely to be tracked via a segment-style tracking API). We should at least call out what's in scope, what's handled elsewhere, and what we're putting on for now

moredip: Oh and also any OF/otel integration is unlikely to help at all for client-side flags, at least for today. AFAIK it's still pretty rare to instrument client-side code with otel (although hopeful that'll change over time)

specification/sections/06-tracking.md
toddbaert: I think it's a bit specific to write it down, I think there's a few other similar products we'd also have to note if we noted this one.

specification/sections/06-tracking.md
toddbaert: > As an end user of an event API I would expect the API from OF for tracking an event to be just a fire and forget. But I don't know how deep we want to formulate that here?

Yes I agree.

I think @federicobond has a good point in that caching and flushing will be important to implementations, but I think that from an OpenFeature _application author_ perspective, this would be transparent. There may be some implications on defining some kind of lifecycle or connecting a hypothetical tracking provider with the existing feature provider lifecycle, to flush data on shutdown, etc.

specification/sections/06-tracking.md
toddbaert: > Flag usage analysis: tracking which flags are no longer being evaluated, so they can be retired.
> 
> But I guess that's probably out of scope for this, because it can probably be implemented entirely within the SDK and/or provider, without needing the app dev to explicitly call anything

Yes, and AFAIK it can be implemented already (though it's not standardized in any way) without the addition of this tracking API. What you're describing would be based on telemetry around impressions... a very important topic, but orthogonal to this initiative.

specification/sections/06-tracking.md
toddbaert: I will clarify this.

specification/sections/06-tracking.md
toddbaert: I will add a point about defining a lifecycle to support existing solutions (caching, flushing, batching, etc)... this also might just meaning hooking into the existing flag provider lifecycle which has similar needs.

@nicklasl @federicobond @liran2000 does that makes sense to you guys?

liran2000: Fine by me.
What about where to implement it - on the provider itself, or as a hook?

federicobond: Agree, from an _application author_ perspective it should be transparent. I was looking at it from a _provider author_ perspective.

nicklasl: I think it is a reasonable goal to have a _decision_ on where the work of things like message batching, flushing and caching takes place.

toddbaert: Great question, but I think it's a technical one for later.

lukas-reining: I think what you said @federicobond makes sense and will be needed but this feels like being part of 
`Specify a flexible API widely compatible with basic vendor functionality`.

`I think it is a reasonable goal to have a decision on where the work of things like message batching, flushing and caching takes place.`

Think so too @nicklasl, but maybe there is a more high level goal we can define for this.
But I am struggling to come up with a more general "subgoal".

specification/sections/06-tracking.md
```diff
@@ -0,0 +1,40 @@
...
+Experimentation is a primary use case for feature flags.
+In practice, this often means flag variants are assigned to users at random or in accordance with a business rule, while the impact of the assigned variant on some business objective is measured.
+Vendors and custom solutions often support a _tracking_ or _goal measuring_ API to facilitate the measurement of these business objectives.
+
+### Goals
+
+- Develop official terminology to support consistent implementation
+- Specify a flexible API widely compatible with basic vendor functionality
+  - Define tracking event payload
+  - Define tracking event identifier
```
Kavindu-Dodan, line 22: Could be an implementation detail, but it could be interesting to cross-link tracking observations with tracing when enabled. And for this we might benefit from linking this identifier with OTel span/trace identifier [1]

[1] - https://opentelemetry.io/docs/specs/otel/trace/api/#retrieving-the-traceid-and-spanid

specification/sections/06-tracking.md
beeme1mr: Yes, I agree that it's implicitly out of scope because `rrweb` is a session replay tool and not an event tracker.

specification/sections/06-tracking.md
nicklasl, line 23: Is it enough the spec the goal as "providing recommendations"?

```suggestion
  - Support client and server paradigms
  - Provide recommendations around: 
    - Async vs sync
    - Flushing mechanisms
    - Event batching
```

toddbaert: I've added `Support underlying SDK technical requirements (batching, caching, and flushing)`. I have intentionally kept this vague since this is a goals document.

toddbaert: Accepted

specification/sections/06-tracking.md
toddbaert: I've clarified this by adding a new goal: `Support A/B testing and experimentation use-cases`

and removing this from non-goals. Non goals are all now negative statements.

beeme1mr (approved):

jonathannorris (approved):

weyert (approved):
Great! Fair goals and non-goals!

nicklasl (approved):

Kavindu-Dodan (approved):

PR #261, opened 2024-05-26
feat: Inline comment support for spec repos

justinabrahms:
## This PR
This should allow someone to mark specifications as done based on inline comments, rather than having to hand-roll regexes.

### How to test
set this in your `.specrc` (example is for go)

```cfg
[spec]
file_extension=go
inline_comment_prefix=//
```

then edit code by adding lines like:

```
import (
	"errors"
	"fmt"
	"sync"

	"github.com/go-logr/logr"
	"github.com/open-feature/go-sdk/openfeature/internal"
	"golang.org/x/exp/maps"
)

//  spec:5.3.5:If the provider emits an event, the value of the client's provider status MUST be updated accordingly.:end

// spec:5.3.4.3:If the provider's on context changed function terminates
// abnormally, and no other invocations have yet to terminate, associated
// PROVIDER_ERROR handlers MUST run.:end

// evaluationAPI wraps OpenFeature evaluation API functionalities
type evaluationAPI struct {
	defaultProvider FeatureProvider
	namedProviders  map[string]FeatureProvider
```

and run `python /path/to/spec/tools/repo_parser/spec_finder.py --code-directory .`

then validate those don't show up in the remaining complaints.

### DIFF
--- a/tools/repo_parser/requirements.txt
+++ b/tools/repo_parser/requirements.txt
@@ -1,0 +1,3 @@
+mypy
+pytest
+pytest-cov
--- a/tools/repo_parser/spec_finder.py
+++ b/tools/repo_parser/spec_finder.py
@@ -6,22 +6,38 @@ import difflib
 import os
 import sys
 import configparser
+from typing import TypedDict, Optional, cast
+
+class Config(TypedDict):
+    file_extension: str
+    multiline_regex: Optional[str]
+    number_subregex: Optional[str]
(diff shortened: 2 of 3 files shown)


lukas-reining (commented):
This looks great!
I will try it out tomorrow, code looks good to me!

beeme1mr (approved):

lukas-reining (approved):

maxveldink (approved):

PR #268, opened 2024-08-26
feat: tracking

toddbaert:
Adds tracking

---

This PR is a first pass at a relatively simple tracking API. The main purpose is to close what is more or less the last gap between the OpenFeature API and that of most SDKs (many SDKs which we currently support have some kind of simple "track" method analogous to what's proposed here). Preceding discussion can be found [here](https://github.com/orgs/open-feature/discussions/376).

Some notable choices:

- I've used the term "occurrence" instead of "event" to minimize terminology overloading
- I've ordered the parameters for the tracking method `occurrenceKey`/`context`/`occurrenceDetails`.
- I've specified the `track` method should never block but instead queue any async work; we agreed there's no reason an application author would ever want to await a I/O associated with a single `track` call; most providers will batch these, I assume
- I've made `value` a "first class" property of the otherwise free-form `OccurrenceDetails` type, so it can be easily mapped into the equivalent field of various providers (something like `targetingKey`)

### DIFF
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -261,3 +261,34 @@ class MyProvider implements Provider {
 Providers may maintain remote connections, timers, threads or other constructs that need to be appropriately disposed of.
 Provider authors may implement a `shutdown` function to perform relevant clean-up actions.
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
+
+### 2.7. Tracking Support
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+Some flag management systems support tracking functionality, which associates feature flag evaluations with subsequent user actions or application state.
+
+See [tracking](./06-tracking.md).
+
+#### Condition 2.7.1
+
+> The `provider` **MAY** define a function for tracking the occurrence of a particular user action or application state, with parameters `occurrence key` (string, required), `evaluation context` (optional) and `occurrence details` (optional) which returns nothing.
+
+```java
+class MyProvider implements Tracking {
+  //...
+
+  /**
+   * Record a tracking occurrence.
+   */
+  void track(String occurrenceKey, OccurrenceDetails details, EvaluationContext context): void;
+  
+  //...
+}
+```
+
+The track function is a void function (function returning nothing).
+The track function performs side effects required to record the `occurrence` in question, which may include network activity or other I/O; this I/O should not block the function call.
+Providers should be careful to complete any communication or flush any relevant uncommitted tracking data before they shut down.
+
+See [shutdown](#25-shutdown).
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -281,7 +281,7 @@ class MyProvider implements Tracking {
   /**
    * Record a tracking occurrence.
    */
-  void track(String occurrenceKey, EvaluationContext context, OccurrenceDetails details): void;
+  void track(String trackingEventName, EvaluationContext context, TrackingEventDetails details): void;
   
   //...
 }
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -272,14 +272,14 @@ See [tracking](./06-tracking.md).
 
 #### Condition 2.7.1
 
-> The `provider` **MAY** define a function for tracking the occurrence of a particular user action or application state, with parameters `occurrence key` (string, required), `evaluation context` (optional) and `occurrence details` (optional) which returns nothing.
+> The `provider` **MAY** define a function for tracking the occurrence of a particular user action or application state, with parameters `tracking event name` (string, required), `evaluation context` (optional) and `tracking event details` (optional) which returns nothing.
 
 ```java
 class MyProvider implements Tracking {
   //...
 
   /**
-   * Record a tracking occurrence.
+   * Record a tracking event.
    */
   void track(String trackingEventName, EvaluationContext context, TrackingEventDetails details): void;
   
@@ -288,7 +288,7 @@ class MyProvider implements Tracking {
 ```
 
 The track function is a void function (function returning nothing).
-The track function performs side effects required to record the `occurrence` in question, which may include network activity or other I/O; this I/O should not block the function call.
+The track function performs side effects required to record the `tracking event` in question, which may include network activity or other I/O; this I/O should not block the function call.
 Providers should be careful to complete any communication or flush any relevant uncommitted tracking data before they shut down.
 
 See [shutdown](#25-shutdown).
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -266,7 +266,7 @@ Alternatively, implementations might leverage language idioms such as auto-dispo
 
 [![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
 
-Some flag management systems support tracking functionality, which associates feature flag evaluations with subsequent user actions or application state.
+Some flag management systems support tracking functionality, which can be used to associate feature flag evaluations with subsequent user actions or application state.
 
 See [tracking](./06-tracking.md).
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -281,7 +281,7 @@ class MyProvider implements Tracking {
   /**
    * Record a tracking occurrence.
    */
-  void track(String occurrenceKey, OccurrenceDetails details, EvaluationContext context): void;
+  void track(String occurrenceKey, EvaluationContext context, OccurrenceDetails details): void;
   
   //...
 }
--- a/specification/sections/06-tracking.md
+++ b/specification/sections/06-tracking.md
@@ -33,20 +33,20 @@ see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
 
 ##### Conditional Requirement 6.1.1.1
 
-> The `client` **MUST** define a function for tracking the occurrence of a particular action or application state, with parameters `occurrence key` (string, required), `evaluation context` (optional) and `occurrence details` (optional) which returns nothing.
+> The `client` **MUST** define a function for tracking the occurrence of a particular action or application state, with parameters `tracking event name` (string, required), `evaluation context` (optional) and `tracking event details` (optional), which returns nothing.
 
 ```java
-// example tracking occurrence recording that a subject reached a page associated with a business goal
+// example tracking event recording that a subject reached a page associated with a business goal
 client.track("visited-promo-page", evaluationContext);
 
-// example tracking occurrence recording that a subject performed an action associated with a business goal, with the occurrence details having a particular numeric value
-client.track("clicked-checkout", evaluationContext, new OccurrenceDetails(99.77));
+// example tracking event recording that a subject performed an action associated with a business goal, with the tracking event details having a particular numeric value
+client.track("clicked-checkout", evaluationContext, new TrackingEventDetails(99.77));
 
-// example tracking occurrence recording that a subject performed an action associated with a business goal, with the occurrence details having a particular numeric value
-client.track("clicked-checkout", evaluationContext, new OccurrenceDetails(99.77).add("currencyCode", "USD"));
+// example tracking event recording that a subject performed an action associated with a business goal, with the tracking event details having a particular numeric value
+client.track("clicked-checkout", evaluationContext, new TrackingEventDetails(99.77).add("currencyCode", "USD"));
 ```
 
-See [evaluation context](../types.md#evaluation-context), [occurrence details](#62-occurrence-details).
+See [evaluation context](../types.md#evaluation-context), [tracking event details](#62-tracking-event-details).
 
 #### Condition 6.1.2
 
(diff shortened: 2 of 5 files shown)


specification/types.md
```diff
@@ -184,3 +188,7 @@ An enumeration of provider events.
 
 A function or method which can be associated with a `provider event`, and runs when that event occurs.
 It declares an `event details` parameter.
+
+### Occurrence Details
+
+See [occurrence details](./sections/06-tracking.md#62-occurrence-details).\
```
beeme1mr, line 194:
```suggestion
See [occurrence details](./sections/06-tracking.md#62-occurrence-details).
```

specification/sections/06-tracking.md
```diff
@@ -10,35 +10,88 @@ toc_max_heading_level: 4
...
+
+see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
+
+##### Conditional Requirement 6.1.1.1
+
+> The `client` **MUST** define a function for tracking the occurrence of a particular action or application state, with parameters `occurrence key` (string, required), `evaluation context` (optional) and `occurrence details` (optional) which returns nothing.
+
+```java
+// example tracking occurrence recording that a subject reached a page associated with a business goal
+client.track("visited-promo-page", evaluationContext);
```
beeme1mr, line 40: How can this be associated with a particular feature flag? Shouldn't flag key be included somewhere?

nicklasl: There is no guarantee that a track invocation happens in a system that has made a feature flag evaluation. The user may have been the target and exposed to a specific value of a feature flag from some other system which impacts the user behaviour. 

The connection between "exposed to feature flag" and "track" can either be done by keeping track of feature flag evaluations, or as suggested in the diagram above, by using a hook to tell the analytics platform that a flag evaluation has happened. The important thing here is that we can provide the evaluation context with the targeting key which is how one would connect the feature flag resolve with the occurrence.

I'm not strongly opposing having a way to append optional metadata (for example flag evaluation results) in the track API, but I do think it introduces a somewhat unnecessary coupling that cannot be guaranteed in all instances.

toddbaert: > There is no guarantee that a track invocation happens in a system that has made a feature flag evaluation. The user may have been the target and exposed to a specific value of a feature flag from some other system which impacts the user behaviour.
> 
> The connection between "exposed to feature flag" and "track" can either be done by keeping track of feature flag evaluations, or as suggested in the diagram above, by using a hook to tell the analytics platform that a flag evaluation has happened. The important thing here is that we can provide the evaluation context with the targeting key which is how one would connect the feature flag resolve with the occurrence.
> 
> I'm not strongly opposing having a way to append optional metadata (for example flag evaluation results) in the track API, but I do think it introduces a somewhat unnecessary coupling that cannot be guaranteed in all instances.

I couldn't have said it better.

Nothing is stopping anyone for adding some flag key in the additional data, but associating user actions with the flag evaluations is primarily the job of the backend flag or analytics platform(s). IMO Flag evaluations and are best thought of as separate transactions that are joined by a user session; And it may be that the flag impacting your business goals isn't even the one some developer though it was. In many ways I think tightly coupling flag evaluations to tracking the way @beeme1mr  is antithetical to the most robust form or experimentation... just looking at the data and querying it for associations, rather than baking some assumptions into the code.

specification/sections/06-tracking.md
```diff
@@ -10,35 +10,88 @@ toc_max_heading_level: 4
...
+
+> The `occurrence details` structure **MUST** define an optional numeric `value`, associating a scalar quality with an `occurrence`.
+
+`Value` is a well-defined field which some providers may map to equivalent numeric values in their API.
+
+See [provider tracking support](./02-providers.md#27-tracking-support).
+
+### Requirement 6.2.2
+
+> The `occurrence details` **MUST** support the inclusion of custom fields, having keys of type `string`, and values of type `boolean | string | number`.
```
kinyoklion, line 97: I think that this details should be equivalent to the structure value available for contexts. Allowing not just a single-level map of raw types, but effectively any JSON representation. If it is just a flag map, then it places what seems to be an unnecessary constraint on less restrictive systems.

kozikow:
It should probably be included in another proposal, but I think it would be also useful to include a concept of "tracking aggregation". At it's core it would be like "list of SQL queries on tracking events".

E.g. see following example - here track event is upvote or downvote. Explore "metrics" by flag, variant:
image
 
Here "aggregation" is "Sentiment Score": AVG(IF(feedback_type="up", 1, IF(feedback_type="down", -1, 0))) and other similar type of metrics.

Then various dashboarding providers can out of the box set up experiment evaluation dashboard. Here Bigquery + Looker Studio is used, but the same flow would be possible with open source Metabase for example.

specification/sections/06-tracking.md
beeme1mr: Thanks for the clarification.

nicklasl:
> It should probably be included in another proposal, but I think it would be also useful to include a concept of "tracking aggregation". At it's core it would be like "list of SQL queries on tracking events".
> ...
> Then various dashboarding providers can out of the box set up experiment evaluation dashboard. Here Bigquery + Looker Studio is used, but the same flow would be possible with open source Metabase for example.

While I appreciate the idea of incorporating "tracking aggregation" as a way to streamline dashboard creation, I believe it's crucial to keep the **specifications** clean and neutral regarding the underlying technology for data storage, combination, and aggregation. 
This approach ensures that our specifications remain versatile and adaptable, allowing various teams/vendors to implement them using the tools and technologies that best fit their needs and existing infrastructure.

For example, there should be nothing stopping me from building a tracking provider that just outputs to a file if that's what my business requires.

However, I am open to have some type of documentation of best practices or examples of use and maybe this is where such proposal belong?

specification/sections/06-tracking.md
```diff
@@ -10,35 +10,88 @@ toc_max_heading_level: 4
...
+##### Conditional Requirement 6.1.1.1
+
+> The `client` **MUST** define a function for tracking the occurrence of a particular action or application state, with parameters `occurrence key` (string, required), `evaluation context` (optional) and `occurrence details` (optional) which returns nothing.
+
+```java
+// example tracking occurrence recording that a subject reached a page associated with a business goal
+client.track("visited-promo-page", evaluationContext);
+
+// example tracking occurrence recording that a subject performed an action associated with a business goal, with the occurrence details having a particular numeric value
+client.track("clicked-checkout", evaluationContext, new OccurrenceDetails(99.77)): void;
```
beeme1mr, line 43: The evaluation context and occurrence details arguments are ordered differently than in the provider section.

specification/sections/02-providers.md
```diff
@@ -256,3 +256,34 @@ class MyProvider implements Provider {
...
+> The `provider` **MAY** define a function for tracking the occurrence of a particular user action or application state, with parameters `occurrence key` (string, required), `evaluation context` (optional) and `occurrence details` (optional) which returns nothing.
+
+```java
+class MyProvider implements Tracking {
+  //...
+
+  /**
+   * Record a tracking occurrence.
+   */
+  void track(String occurrenceKey, OccurrenceDetails details, EvaluationContext context): void;
```
kozikow, line 279: I think the biggest use case for this is joining events with variants to see what "business impact" flag changes have. 

The way I understand it, most people will be just "joining" this with what's proposed in [Feature Flag Impression SemConv](https://docs.google.com/document/d/1sBDD-uifx-Qy0NxtwRcrRAQ48sHQS9dTp6tpArQsoZg/edit)

People just need something to join both entries on. SemConv proposal only suggest logging "context id" without full context. In demo code I am looking to adapt to use this proposal I have been using targetingId. I am thinking - wouldn't it make sense to have track method schema match schema of SemConv? Then people with simple use cases can even use targetingId as contextId.

toddbaert: > I think big use case for occurrences is joining occurrences with variants to see what "business impact" flag changes have.
> 
> The way I understand it, most people will be just "joining" this with what's proposed in [Feature Flag Impression SemConv](https://docs.google.com/document/d/1sBDD-uifx-Qy0NxtwRcrRAQ48sHQS9dTp6tpArQsoZg/edit)

100%. This is how I envision things would work, and how most people will use it, I think.

> People just need something to join both entries on. SemConv proposal only suggest logging "context id" without full context. I have been using targetingId for this use case so far. I am thinking - wouldn't it make sense to have track method schema match schema of SemConv?
> 
> I am not advocating for any particular solution, but for more alignment between those two proposals and enabling use case of joining variant impressions with occurrences.

I struggled a lot with this same question.

I agree that generally most people would probably join on the `targetingKey` but I'm not sure that would be the case 100% of the time. If we include the entire evaluation context in the track API, users have the ability to join on whatever attributes they want which are common to the telemetry data from a flag evaluation and from the `track` call (perhaps an attribute called `email`, for example). It's for this maximum flexibility I proposed the API take the evaluation context instead of the `targetingKey` only.

Additionally, taking the entire `evaluationContext` is more consistent with existing functions, keeping the API a bit more isomorphic.

All that said, I understand why taking the `targetingKey` directly in the track method is appealing. Like I said, I struggled with this question for a while, and I'm open to other perspectives.

cc @beeme1mr @dyladan @nicklasl

specification/sections/06-tracking.md
toddbaert: Thanks, fixed.

specification/sections/02-providers.md
nicklasl: 💯 percent agree with what @toddbaert wrote on the flexibility aspect and I would like to add something:
Passing the whole evaluation context in the track API allows the tracking functionality to stand on its own legs as we (the OpenFeature API) provide enough data to the tracking providers to be enable slicing and dicing the produced tracking data by imaginative dimensions that could be deduced in the evaluation context.

Additionally:
We are talking about joining data between the moment of evaluation (I think of this as the moment of exposure) and the moment of event tracking. The evaluation context may have changed between these two moments in time, it can also be that these two moments are captured in services with different evaluation context. 
**The question is; what context is important to track?**
If I could only choose one context to store, I would keep the context of the tracking event but of course, having both of these stored would for sure be the best. So I'm arguing that from a business perspective the context data from the event is more important than the context data from the evaluation.

Now there may be some nuances in the above reasoning between static- and dynamic context paradigms, especially when it comes to the moment of exposure in the static paradigm. I would argue that for the static paradigm it is not the flag evaluation (when the client bulk fetches the flags from the backend) but rather the flag access that should be counted as the moment of exposure.

specification/sections/06-tracking.md
toddbaert: I'm in agreement with this proposal, with the caveat that we're OK with some additional complexity.

specification/sections/06-tracking.md
```diff
@@ -10,35 +10,92 @@ toc_max_heading_level: 4
...
+
+The track function is a void function (function returning nothing).
+Though it may be associated with network activity or other I/O, it need not be awaited by application authors.
+
+```java
+// example tracking occurrence recording that a subject reached a page associated with a business goal
+client.track("visited-promo-page");
+
+// example tracking occurrence recording that a subject performed an action associated with a business goal, with the occurrence details having a particular numeric value
+client.track("clicked-checkout", new OccurrenceDetails(99.77)): void;
```
roncohen, line 68: Maybe `OccurrenceDetails` was defined somewhere else in detail (or maybe this example is more for illustrative purposese), but it's not clear to me how you'd set both the `value=99.88` and then also a set of other properties for the occurrence - something that all the common providers support.

toddbaert: Good point, I've added this:

```java
new OccurrenceDetails(99.77).add("currencyCode", "USD")
```

roncohen (approved):

beeme1mr (approved):

nicklasl (approved):

toddbaert:
Draft JS implementation available here: https://github.com/open-feature/js-sdk/pull/1020

lukas-reining (approved):

toddbaert:
@moredip suggests we perhaps use something other than "occurrence" (see comment [here](https://github.com/open-feature/js-sdk/pull/1020#discussion_r1777193836) in draft JS implementation).

I went with "occurrence" to reduce collision with our exciting event concepts (for example, we already have `event` and `eventDetails` objects. Still, we can disambiguate these other ways:

Currently I've proposed:

```ts
void track(String occurrenceKey, EvaluationContext context, OccurrenceDetails details): void;
```

but an obvious alternative would be:

```ts
void track(String trackingEventKey, EvaluationContext context, TrackingEventDetails details): void;
```
I'm also not sure if `trackingEventId` or `trackingEventKey` is better; we use `key` for flags themselves, but a `key` to me suggests a string index a specific set (perfect for flags in flags sets) whereas an id might be a "softer" reference, such as a UUID.

kinyoklion (approved):

toddbaert:
Demo of some of these features: https://www.youtube.com/watch?v=Od-7Pjs2dKQ&t=436s

roncohen:
> Does anyone have points in favor of one or the other?

The occurrence terminology threw me off initially. I understand the background for picking `occurrence` over event/trackingEvent, but I feel that in the context of the `track` call, the confusion should be manageable and the advantage of aligning on known terminology is substantial. 

So I’m in favor of going with `trackingEvent`,  `trackEvent` or simply `event` (and `trackingEventKey`, `trackEventKey` or `eventKey` respectively)

> I'm also not sure if trackingEventId or trackingEventKey is better


Agreed that “key” is better here. Some platforms give specific events UUIDs for deduplication purposes etc which is not what we’re talking about here. I think “name” could also be a contender. Segment refers to this as the name (e.g. “event name”): https://segment.com/docs/connections/spec/track/

roncohen:
In the interest of moving this forward, I'd go for:
```
void track(String trackingEventName, EvaluationContext context, TrackingEventDetails details): void;
```

toddbaert:
> In the interest of moving this forward, I'd go for:
> 
> ```ts
> void track(String trackingEventName, EvaluationContext context, TrackingEventDetails details): void;
> ```

I like this and based on slack we have a consensus. Will update.

toddbaert:
I've made updates based on the most recent feedback. As usual with spec PRs, we want a wide consensus; so far I think I've seen that represented in this PR and in various meetings and communications in Slack. 

I will leave this PR open for additional comments, but I plan to merge it early next week. If you have comments or concerns, please voice them here before then! :pray:

toddbaert:
> Nice @toddbaert. Looks like there's still some "occurrence" terminology left in this PR.

Nice catch, @roncohen . I think I got them all this time :sweat_smile: . There's still a few cases of the word "occurrence" but I think they are simply descriptive, not params/types.

roncohen (approved):

PR #269, opened 2024-08-29
feat: logging requirements and logging hook

toddbaert:
As discussed in recent meetings, we've had complaints about this from multiple sources. This PR adds a stipulation that no logging is done by the SDK during flag evaluation. It also defines a very simple `logging hook` concept as an included utility. These should be very simple to write and will provide all the needed logging but give authors more control than built in log statements. It will also be a very nice intro to the hooks concept for users.

Here's the Java implementation: https://github.com/open-feature/java-sdk/pull/1084

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -353,9 +353,15 @@ Configuration code includes code to set the provider, instantiate providers, and
 
 #### Requirement 1.4.11
 
-> In the case of abnormal execution, the client **SHOULD** log an informative error message.
+> Methods, functions, or operations on the client **SHOULD NOT** write log messages.
 
-Implementations may define a standard logging interface that can be supplied as an optional argument to the client creation function, which may wrap standard logging functionality of the implementation language.
+The client methods (particularly the evaluation methods) run in hot code paths.
+Logging (even at error level) can cause a huge volume of log entries.
+For example, in a circumstance in which an application expecting a particular flag to exist is deployed in advance of that flag's being defined in the management system, logs can become inundated with `FLAG_NOT_FOUND` messages and related stack traces.
+Logging in these code paths is highly discouraged.
+Application authors can attach a [logging hook](../appendix-a-included-utilities.md#logging-hook) or author their own custom logging hook(s) to help with debugging or satisfy their particular logging needs.
+
+Logging is encouraged in functions to do with configuration, initialization, shutdown, etc.
 
 #### Requirement 1.4.12
 
--- a/specification.json
+++ b/specification.json
@@ -233,8 +233,8 @@
         {
             "id": "Requirement 1.4.11",
             "machine_id": "requirement_1_4_11",
-            "content": "In the case of abnormal execution, the client SHOULD log an informative error message.",
-            "RFC 2119 keyword": "SHOULD",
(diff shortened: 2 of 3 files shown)


specification/sections/01-flag-evaluation.md
```diff
@@ -381,6 +381,18 @@ This `flag metadata` field is intended as a mechanism for providers to surface a
 
 > Condition: `Flag metadata` **MUST** be immutable.
 
+#### Requirement 1.4.16
+
+> Methods, functions, or operations on the client **SHOULD NOT** write log messages.
```
toddbaert, line 386: This is one of the few times I've thought SHOULD NOT was the best rfc-2119 keyword: https://datatracker.ietf.org/doc/html/rfc2119#section-4

I think there may be exceptions for some languages (possibly debug level logging which is not generally used unless doing some kind of issue investigation or during local development).

beeme1mr (approved):

nicklasl (approved):

specification/sections/01-flag-evaluation.md
```diff
@@ -353,9 +353,15 @@ Configuration code includes code to set the provider, instantiate providers, and
 
 #### Requirement 1.4.11
 
-> In the case of abnormal execution, the client **SHOULD** log an informative error message.
+> Methods, functions, or operations on the client **SHOULD NOT** write log messages.
```
lukas-reining, line 356: Couldn't we go for MAY NOT here?
I think this is fine to enforce.

lukas-reining (approved):
Looks good to me!
Left one question.

specification/sections/01-flag-evaluation.md
toddbaert: My reasoning is that there's some amount of "gray area" here. For example. I left some logging for unhandled errors in event handlers and hooks, which are technically part of the lifecycle, but seem to me to be a lot different than something like a FLAG_NOT_FOUND in an management system (more like a coding error). RFC2113 feels right to me for this, because it allows deviation, but only with thoughtful consideration:

https://datatracker.ietf.org/doc/html/rfc2119#section-4

> This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.
   
I think for logging, this is good. You _can_ log in the hot path, but it's discouraged and should be done very carefully to avoid logspam.

I think if we do MUST NOT, our list of exceptions will be growing forever for valid reasons.

lukas-reining: Okay makes sense to me!

aepfli (approved):

askpt (approved):
LGTM. I have a minor question. With this spec change, will the `LogHook` become part of the SDK and should it be moved to the SDK repo?

toddbaert:
> LGTM. I have a minor question. With this spec change, will the `LogHook` become part of the SDK and should it be moved to the SDK repo?

Yep! I've implemented one in Java now here: https://github.com/open-feature/java-sdk/pull/1084

A `LoggingHook` should have no additional dependencies and is generally valuable, so the idea is they will be included in the SDK just like the in-memory providers.

PR #273, opened 2024-09-30
feat: Add hook-data concept for hooks.

kinyoklion:
## This PR

Add support for the hook-data concept for hooks. Hook-data allows for per-evaluation data to be propagated between hooks. 

This is especially useful for analytics purposes where you may want to measure things that happen between stages, or you want to do something like create a span in one stage and close it in another.

This concept is similar to the `series data` concept for LaunchDarkly hooks. https://github.com/launchdarkly/open-sdk-specs/tree/main/specs/HOOK-hooks#evaluationseriesdata

Unlike `series data` the data in this approach is mutable. This is because the `before` stage already has a return value. We could workaround this by specifying a return structure, but it maybe seems more complex. The data is only passed to a specific hook instance, so mutability is not of great concern.

Some functional languages may still need to use an immutable with return values approach.

I can create an OFEP if we think this merits discussion prior to proposal.

### Related Issues

Related discussion in a PR comment.
https://github.com/open-feature/java-sdk/pull/1049#discussion_r1718895742


specification/sections/04-hooks.md
```diff
@@ -31,7 +31,9 @@ Hooks can be configured to run globally (impacting all flag evaluations), per cl
 
 ### Definitions
 
-**Hook**: Application author/integrator-supplied logic that is called by the OpenFeature framework at a specific stage. **Stage**: An explicit portion of the flag evaluation lifecycle. e.g. `before` being "before the [resolution](../glossary.md#resolving-flag-values) is run. **Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)` is an invocation. **API**: The global API singleton.
+**Hook**: Application author/integrator-supplied logic that is called by the OpenFeature framework at a specific stage.
```
kinyoklion, line 34: These were all on one line, which seemed hard to read. So I put a blank line between them to get them on multiple lines.

specification/sections/04-hooks.md
```diff
@@ -81,75 +83,154 @@ see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
 
 > Condition: The provider `metadata` field in the `hook context` **MUST** be immutable.
 
-### 4.3. Hook creation and parameters
+### 4.3. Hook data
```
kinyoklion, line 86: I prefer not renumbering, but I did in this case as conceptually it makes sense to group all the parameter requirements together. I could move this to retain the existing numbering.

specification/sections/04-hooks.md
```diff
@@ -81,75 +83,154 @@ see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
...
+#### Requirement 4.4.1
+
 > Hooks **MUST** specify at least one stage.
 
-#### Condition 4.3.2
+#### Requirement 4.4.2
+
+> `Hook data` **MUST** must be created before the first `stage` invoked in a hook for a specific evaluation and propagated between each `stage` of the hook.
+
+Example showing data between `before` and `after` stage for two different hooks.
```
kinyoklion, line 123: I am not sure if the diagram is overkill, but a key concept is that data is not shared between hooks and not shared between evaluations.

specification/sections/04-hooks.md
```diff
@@ -31,7 +31,9 @@ Hooks can be configured to run globally (impacting all flag evaluations), per cl
 
 ### Definitions
 
-**Hook**: Application author/integrator-supplied logic that is called by the OpenFeature framework at a specific stage. **Stage**: An explicit portion of the flag evaluation lifecycle. e.g. `before` being "before the [resolution](../glossary.md#resolving-flag-values) is run. **Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)` is an invocation. **API**: The global API singleton.
+**Hook**: Application author/integrator-supplied logic that is called by the OpenFeature framework at a specific stage.
+
+**Stage**: An explicit portion of the flag evaluation lifecycle. e.g. `before` being "before the [resolution](../glossary.md#resolving-flag-values) is run. **Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)` is an invocation. **API**: The global API singleton.
```
beeme1mr, line 36:
```suggestion
**Stage**: An explicit portion of the flag evaluation lifecycle. e.g. `before` being "before the [resolution](../glossary.md#resolving-flag-values) is run.

**Invocation**: A single call to evaluate a flag. `client.getBooleanValue(..)` is an invocation.

**API**: The global API singleton.
```

beeme1mr (commented):
I like this proposal, but I would like to see how it could be implemented in an SDK. Properly scoping hook data may be complicated.

specification/sections/04-hooks.md
toddbaert: I think I prefer moving it to avoid the number thrashing. It's slightly inconvenient that things aren't grouped, but I think it's more important to attempt to maintain stability of the requirement numbers.

specification/sections/04-hooks.md
toddbaert: I think this is diagram-worthy.

specification/sections/04-hooks.md
kinyoklion: I will update it.

specification/sections/04-hooks.md
```diff
@@ -81,75 +87,154 @@ see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
...
 
 > The implementation uses the dynamic-context paradigm.
 
 see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
 
-##### Conditional Requirement 4.3.2.1
+##### Conditional Requirement 4.4.2.1
 
-> The `before` stage **MUST** run before flag resolution occurs. It accepts a `hook context` (required) and `hook hints` (optional) as parameters and returns either an `evaluation context` or nothing.
+> The `before` stage **MUST** run before flag resolution occurs. It accepts a `hook context` (required), `hook hints` (optional), and `hook data` (required) as parameters and returns either an `evaluation context` or nothing.
```
toddbaert, line 183: Should hook data be optional? I think most implementations could tolerate overloads for the stage functions:

```java
before(HookContext context, HookHints hints);
before(HookContext context, HookData data, HookHints hints);
```
This would mean that current hooks wouldn't break, I think.

The other option might be to add the `hook data` to the `hook context` object.

specification/sections/04-hooks.md
```diff
@@ -81,75 +87,154 @@ see: [dynamic-context paradigm](../glossary.md#dynamic-context-paradigm)
...
+### 4.4. Hook creation and parameters
+
+#### Requirement 4.4.1
+
 > Hooks **MUST** specify at least one stage.
 
-#### Condition 4.3.2
+#### Requirement 4.4.2
+
+> `Hook data` **MUST** must be created before the first `stage` invoked in a hook for a specific evaluation and propagated between each `stage` of the hook.
```
toddbaert, line 125: I think it's implied here, but not explicit, that hook data is scoped only to the hook instance in question. I think it might be worth adding a specific point about that, and perhaps even a recommendation or example of how it might be accomplished in some implementations.

toddbaert (commented):
I definitely support this idea. I left a few comments. My biggest question is how we can define this so it's non-breaking for existing hooks.

specification/sections/04-hooks.md
kinyoklion: I can think about the hook context. For the LD hooks the context was fully immutable, but for OpenFeature it isn't exactly the case being as the evaluation context within the hook context can change. So that makes it a potential home. It does make the mutability story a bit more complex.

kinyoklion: Updated to be in the hookContext.

specification/sections/04-hooks.md
kinyoklion: I added some more clarifying language a couple places.

kinyoklion:
> I like this proposal, but I would like to see how it could be implemented in an SDK. Properly scoping hook data may be complicated.

I should be able to find time to update a OF SDK to demonstrate, but it may take me a bit.

It will be a little different now that we added it to hook context, but you can see what we did in the LD SDKs as its own argument:
https://github.com/launchdarkly/js-core/blob/ae2b5bbed9ec32d9640f1c5a168badaad28174ec/packages/shared/sdk-server/src/hooks/HookRunner.ts#L109

In our case we create the data when we call the hook, and the return is the new data. Which means we can just map it to an array.

But, if that was not the case, as it will be in the OF example you instead do this.

1. Create an array the same size as the number of hooks being invoked.
2. As you iterate the hooks pass the data at the same index in the array as the hook is in its array.
3. Maintain the array for the execution of that series of stages.

Some pseudocode: 
```
const hookData = new Array(hooksForInvocation.length);
hookData.fill({})

...
hooksForInvocation.forEach((hook, index) => executeBeforeEvaluation(hookContext.withData(hookData[index])))
```

The hooks are stable for a specific invocation of stages, so you just need an equal sized list of hook data.

beeme1mr (approved):

specification/sections/04-hooks.md
```diff
@@ -225,3 +299,33 @@ see: [Flag evaluation options](./01-flag-evaluation.md#evaluation-options)
...
+    if (value instanceof Span) {
+      Span span = (Span) value;
+      span.end();
+    }
+  }
+```
+
+#### Requirement 4.6.1
+
+> `hook data` **MUST** be a structure supports definition of arbitrary properties, with keys of type `string`, and values of type `boolean | string | number | datetime | structure`.
```
lukas-reining, line 331: I thing it needs to be `supporting`, but I might be reading it wrong.

```suggestion
> `hook data` **MUST** be a structure supporting the definition of arbitrary properties, with keys of type `string`, and values of type `boolean | string | number | datetime | structure`.
```

lukas-reining (approved):
I like the change and as far as I can see, adding the hook data does not break the old hooks when it is implemented.

kinyoklion:
> I like this proposal, but I would like to see how it could be implemented in an SDK. Properly scoping hook data may be complicated.

I think this was resolved via the pseudocode.

guidobrei (approved):
Not that my approval changes anything, but this proposal looks really promising.

toddbaert (approved):
:+1: This has been open for a while. @kinyoklion feel free to merge whenever, and we can do a release.

I'd like to propose additional hardening of the specification soon, and possibly moving this section out of experimental (probably after we've implemented what's outlined in this PR in a few SDKs). When this is merged I'll create a parent issue in the spec and sub issues in each SDK.

PR #278, opened 2024-11-04
chore: use html entity code for asterisk in mermaid diagram

erka:
## This PR

image

The mermaid diagram for `Provider lifecycle` has few arrows with `Unsupported markdown: list` text. It will be fixed by using `#ast`;

https://mermaid.js.org/syntax/flowchart.html#entity-codes-to-escape-characters


lukas-reining (approved):

beeme1mr (approved):
Thank you!

PR #287, opened 2024-12-16
feat: add appendix D (observability)

toddbaert:
- adds small appendix section to spec that defines how particular SDK fields are mapped to the recently merge OTel semantics
- this allows providers and hooks to export OTel data consistently for easy interop
- adds a couple supporting glossary liks


specification/appendix-d-observability.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Evaluation Details
+
+The following describes how fields on the [evaluation details](types.md#evaluation-details) are mapped to feature flag log records:
+
+| Evaluation Details Field | Log Record Attribute                    | Notes                                                |
+| ------------------------ | --------------------------------------- | ---------------------------------------------------- |
+| `flag key`               | `feature_flag.key`                      | See: [flag key](./glossary.md#flag-key)              |
+| `error code`             | `error.type`                            | See: [error code](./types.md#error-code)             |
+| `variant`, `value`       | `feature_flag.variant`                  | See: [variant](./glossary.md#variant)                |
```
beeme1mr, line 26: Should `value` be listed here?

specification/appendix-d-observability.md
```diff
@@ -0,0 +1,50 @@
...
+| `error message`          | `feature_flag.evaluation.error.message` | An error message associated with a failed evaluation |
+| `reason`                 | `feature_flag.evaluation.reason`        | See: [reason](./types.md#reason)                     |
+
+> [!NOTE]  
+> If the provider in question doesn't supply a `variant`, the a string representation of the `value` is used.
```
beeme1mr, line 31: The OTel spec allows for `value` to be "any" type.

specification/appendix-d-observability.md
```diff
@@ -0,0 +1,50 @@
...
+| `flagSetVersion`  | `feature_flag.version`    | A version string (format unspecified) for the [flag set](./glossary.md#flag-set).                                                                                                               |
+
+> [!NOTE]  
+> Keys in flag metadata use the "camelCase" casing convention, while the OpenTelemetry standard uses a namespaced "snake_case" convention.
+
+### Provider Metadata
+
+| Provider Metadata Field | Log Record Attribute         | Notes                                                                                            |
+| ----------------------- | ---------------------------- | ------------------------------------------------------------------------------------------------ |
+| `name`                  | `feature_flag.provider_name` | The name of the provider as defined in the `provider metadata`, available in the `hook context`. |
```
beeme1mr, line 50: I wonder if we should add a note that it's recommended to keep provider names consistent across provider implementations.

specification/appendix-d-observability.md
toddbaert: Ya, it's mentioned in the !NOTE you commended on below.

specification/appendix-d-observability.md
toddbaert: I have removed "string".

specification/appendix-d-observability.md
toddbaert: I've removed this for now until there's more clarity on how Otel is handling event data (attrs vs body)

specification/appendix-d-observability.md
toddbaert: I've removed this whole note for now (see https://github.com/open-feature/spec/pull/287#discussion_r1887371546)

specification/appendix-d-observability.md
toddbaert: I don't think it's worth mentioning. I guess providers can decide for themselves if they want the name to be consistent.

specification/appendix-d-observability.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Evaluation Details
+
+The following describes how fields on the [evaluation details](types.md#evaluation-details) are mapped to feature flag log records:
+
+| Evaluation Details Field | Log Record Attribute                    | Notes                                                |
+| ------------------------ | --------------------------------------- | ---------------------------------------------------- |
+| `flag key`               | `feature_flag.key`                      | See: [flag key](./glossary.md#flag-key)              |
+| `error code`             | `error.type`                            | See: [error code](./types.md#error-code)             |
+| `variant`, `value`       | `feature_flag.variant`                  | See: [variant](./glossary.md#variant)                |
```
toddbaert: There's questions about how exactly to represent `value` (important if provider cannot supply `variant`):

- is it represented in the body?
- can it be of `any` type?

specification/appendix-d-observability.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Evaluation Details
+
+The following describes how fields on the [evaluation details](types.md#evaluation-details) are mapped to feature flag log records:
+
+| Evaluation Details Field | Log Record Attribute                    | Notes                                                |
+| ------------------------ | --------------------------------------- | ---------------------------------------------------- |
+| `flag key`               | `feature_flag.key`                      | See: [flag key](./glossary.md#flag-key)              |
+| `error code`             | `error.type`                            | See: [error code](./types.md#error-code)             |
+| `variant`                | `feature_flag.variant`                  | See: [variant](./glossary.md#variant)                |
```
toddbaert, line 26: There's questions about how exactly to represent `value` (important if provider cannot supply `variant`):

- is it represented in the body?
- can it be of `any` type?

toddbaert:
Moving to draft for now based on some ongoging OTel sem-conv discussion: https://bitbucket.lab.dynatrace.org/projects/DEUS/repos/semantic-dictionary/pull-requests?create&sourceBranch=refs%2Fheads%2Ffeature%2FOA-40566-update-semantic-dictionary-feature-flag-events&targetRepoId=9646

specification/appendix-d-observability.md
lukas-reining: > is it represented in the body?

Do you mean the body of the log record?

The [spec](https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/) says: 

```
value: If and only if feature flag provider does not supply variant or equivalent concept. 
Otherwise, value should be treated as opt-in.
```

And the type is `undefined` there as there is no `any` type in OTEL. 
Recently there was also opened a PR to add the `value_type` to the event: https://github.com/open-telemetry/semantic-conventions/pull/1689

specification/appendix-d-observability.md
lukas-reining: I think so too @toddbaert, it might be interesting for some consumers to see the provider implementations instead of a common name. But I guess @beeme1mr's case is the more common.

lukas-reining (commented):
This looks good to me, we might consider the PR I mentioned regarding the `value_type`, but I think we could also start with the value as `any` type and add the type hint from the PR later as it does not change the any type.

specification/appendix-d-observability.md
```diff
@@ -0,0 +1,51 @@
...
+| `error message`          | `feature_flag.evaluation.error.message` | An error message associated with a failed evaluation |
```
guidobrei, line 27: Should we here also make clear that it is not recommended to not repat `error.type` here, like it's stated in the otel semcon?

specification/appendix-d-observability.md
guidobrei: I'm also with @toddbaert regarding a provider name. But I can imagine your point @beeme1mr of having an identifiying property for post processing. Maybe there should also be a `provider_id` property we could propose in otel?

specification/appendix-d-observability.md
toddbaert: I've added a section about this.

specification/appendix-d-observability.md
toddbaert: Yes, I've update this description a bit: `A human-readable error message associated with a failed evaluation. For programmatic purposes, refer to `error code`. `

kinyoklion (approved):

lukas-reining (approved):

askpt (approved):

toddbaert:
Merged with just 3 approvals since we have a diverse set and this is just an appendix.

PR #289, opened 2025-01-13
feat: Update test harness with metadata assertions #1467

chrfwow:
## This PR
Adds gherkin test to verify that flag evaluations provide metadata

### Related Issues

Part of #1467 ([https://github.com/open-feature/flagd/issues/1467](https://github.com/open-feature/flagd/issues/1467))

### Follow-up Tasks
Implement steps of the gherkin file in the repositories, and add test data according to the issue

### DIFF
--- a/specification/assets/gherkin/metadata.feature
+++ b/specification/assets/gherkin/metadata.feature
@@ -1,0 +1,25 @@
+@Metadata
+Feature: Metadata
+
+  Background:
+    Given a provider is registered
+
+  Scenario: Returns metadata
+    Given a Boolean-flag with key "metadata-flag" and a default value "true"
+    When the flag was evaluated with details
+    Then the resolved metadata value "string" with type "String" should be "1.0.2"
+    And the resolved metadata value "integer" with type "Integer" should be "2"
+    And the resolved metadata value "double" with type "Double" should be "0.1"
+    And the resolved metadata value "boolean" with type "Boolean" should be "true"
+
+  Scenario Outline: Returns no metadata
+    Given a <type>-flag with key "<key>" and a default value "<default_value>"
+    When the flag was evaluated with details
+    Then the resolved metadata is empty
+
+    Examples: Flags
+      | key          | type    | default_value |
+      | boolean-flag | Boolean | true          |
+      | integer-flag | Integer | 23            |
+      | float-flag   | Float   | 2.3           |
+      | string-flag  | String  | value         |
--- a/specification/assets/gherkin/metadata.feature
+++ b/specification/assets/gherkin/metadata.feature
@@ -7,10 +7,12 @@ Feature: Metadata
   Scenario: Returns metadata
     Given a Boolean-flag with key "metadata-flag" and a default value "true"
     When the flag was evaluated with details
-    Then the resolved metadata value "string" with type "String" should be "1.0.2"
-    And the resolved metadata value "integer" with type "Integer" should be "2"
-    And the resolved metadata value "double" with type "Double" should be "0.1"
-    And the resolved metadata value "boolean" with type "Boolean" should be "true"
+    Then the resolved metadata should contain
+      | key     | metadata_type | value |
+      | string  | String        | 1.0.2 |
+      | integer | Integer       | 2     |
+      | double  | Double        | 0.1   |
+      | boolean | Boolean       | true  |
 
   Scenario Outline: Returns no metadata
     Given a <type>-flag with key "<key>" and a default value "<default_value>"
--- a/specification/assets/gherkin/metadata.feature
+++ b/specification/assets/gherkin/metadata.feature
@@ -2,7 +2,7 @@
 Feature: Metadata
 
   Background:
-    Given a provider is registered
+    Given a stable provider
 
   Scenario: Returns metadata
     Given a Boolean-flag with key "metadata-flag" and a default value "true"
--- a/specification/assets/gherkin/metadata.feature
+++ b/specification/assets/gherkin/metadata.feature
@@ -11,17 +11,17 @@ Feature: Metadata
       | key     | metadata_type | value |
       | string  | String        | 1.0.2 |
       | integer | Integer       | 2     |
-      | double  | Double        | 0.1   |
+      | float   | Float         | 0.1   |
       | boolean | Boolean       | true  |
 
   Scenario Outline: Returns no metadata
-    Given a <type>-flag with key "<key>" and a default value "<default_value>"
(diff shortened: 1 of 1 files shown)


specification/assets/gherkin/metadata.feature
```diff
@@ -0,0 +1,25 @@
+@Metadata
+Feature: Metadata
+
+  Background:
+    Given a provider is registered
+
+  Scenario: Returns metadata
+    Given a Boolean-flag with key "metadata-flag" and a default value "true"
+    When the flag was evaluated with details
```
toddbaert, line 9:
```suggestion
    Given a Boolean-flag with key "metadata-flag" is evaluated with default value "false"
```

Let's use the same sentence structure as the [other steps](https://github.com/open-feature/spec/blob/d261f68331b94fd8ed10bc72bc0485cfc72a51a8/specification/assets/gherkin/evaluation.feature#L10), this way the same bindings will "just work" and less code needs to be implemented.

specification/assets/gherkin/metadata.feature
```diff
@@ -0,0 +1,25 @@
...
+    Given a <type>-flag with key "<key>" and a default value "<default_value>"
+    When the flag was evaluated with details
+    Then the resolved metadata is empty
+
+    Examples: Flags
+      | key          | type    | default_value |
+      | boolean-flag | Boolean | true          |
+      | integer-flag | Integer | 23            |
+      | float-flag   | Float   | 2.3           |
+      | string-flag  | String  | value         |
```
toddbaert, line 25: I'm not sure these empty-metadata tests are worth it TBH. cc @aepfli what do you think?

specification/assets/gherkin/metadata.feature
```diff
@@ -0,0 +1,25 @@
...
+  Background:
+    Given a provider is registered
+
+  Scenario: Returns metadata
+    Given a Boolean-flag with key "metadata-flag" and a default value "true"
+    When the flag was evaluated with details
+    Then the resolved metadata value "string" with type "String" should be "1.0.2"
+    And the resolved metadata value "integer" with type "Integer" should be "2"
+    And the resolved metadata value "double" with type "Double" should be "0.1"
+    And the resolved metadata value "boolean" with type "Boolean" should be "true"
```
toddbaert, line 13: We could probably use a table for this, with columns for the metadata key, type, and value.

specification/assets/gherkin/metadata.feature
aepfli: I would specifically separate those two. Sometimes, you need to flag information without an evaluation. Hence, it is only a given state. But the when is the dedicated action. eg when it was modified. With this, we separate data and action. We already did the same thing within the testbed.

toddbaert:
@aepfli do you think it makes more sense to simply add metadata to the existing gherkin suite? We already have error flags there - we could add metadata to some of the existing flags and error flags, and them update the in-memory providers to suppor this.

specification/assets/gherkin/metadata.feature
aepfli: yeah this was more a little exploration showcasing from my side, i still think that a negative case, to ensure we are not polluting the response makes sense

aepfli:
> @aepfli do you think it makes more sense to simply add metadata to the existing gherkin suite? We already have error flags there - we could add metadata to some of the existing flags and error flags, and them update the in-memory providers to support this.

I am not sure if integration into the existing suite is suitable. We could enhance the existing ones, with a description of what the purpose of those flags is. But generally i prefer a flagname which already tells me what i can expect from this flag.

This also brings me to the next topic. As we are currently generating the data in each sdk on our own, it might be worth to add a json here with the testdata, which we load into the inmemory provider, wdt?

aepfli (approved):

toddbaert (approved):

specification/assets/gherkin/metadata.feature
```diff
@@ -0,0 +1,27 @@
...
+    Given a provider is registered
+
+  Scenario: Returns metadata
+    Given a Boolean-flag with key "metadata-flag" and a default value "true"
+    When the flag was evaluated with details
+    Then the resolved metadata should contain
+      | key     | metadata_type | value |
+      | string  | String        | 1.0.2 |
+      | integer | Integer       | 2     |
+      | double  | Double        | 0.1   |
```
guidobrei, line 14: Metadata type should be `Float`

specification/assets/gherkin/metadata.feature
```diff
@@ -0,0 +1,27 @@
+@Metadata
+Feature: Metadata
+
+  Background:
+    Given a provider is registered
```
aepfli, line 5:
```suggestion
    Given a stable provider
```

to be more in sync with our new flagd gherkin files, which have been already an inspiration for this file

toddbaert:
> This also brings me to the next topic. As we are currently generating the data in each sdk on our own, it might be worth to add a json here with the testdata, which we load into the inmemory provider, wdt?

Most in-memory providers can't actually load JSON, so I'm not sure how valuable that would be.

aepfli:
> > This also brings me to the next topic. As we are currently generating the data in each sdk on our own, it might be worth to add a json here with the testdata, which we load into the inmemory provider, wdt?
> 
> Most in-memory providers can't actually load JSON, so I'm not sure how valuable that would be.

i would not load them automagically, but within the tests, we can load the json files, and provide the data as parameter to the inmemory provider. We would centralize the test flag data, and with new gherkin files, which only add new data for experimentation everything should work ootb.

toddbaert:
> > > This also brings me to the next topic. As we are currently generating the data in each sdk on our own, it might be worth to add a json here with the testdata, which we load into the inmemory provider, wdt?
> > 
> > 
> > Most in-memory providers can't actually load JSON, so I'm not sure how valuable that would be.
> 
> i would not load them automagically, but within the tests, we can load the json files, and provide the data as parameter to the inmemory provider. We would centralize the test flag data, and with new gherkin files, which only add new data for experimentation everything should work ootb.
> 
> ```
> data = loadJsonData(jsonfile)
> provider = InMemoryProvider(data) // simple prosa
> ```

Oh, so all implementations would write a simple function `loadJsonData` or similar which would do the conversion? I guess that's not a bad idea since it might be fairly future-proof when new things are added.

aepfli:
> Oh, so all implementations would write a simple function `loadJsonData` or similar which would do the conversion? I guess that's not a bad idea since it might be fairly future-proof when new things are added.

excactly, and this way we ensure, everyone gets the same data afterall

guidobrei (approved):

PR #290, opened 2025-01-14
Add gherkin tests to verify evaluation details in finally hooks

chrfwow:
## This PR
Adds gherkin tests to verify the evaluation details passed to the finally hooks

### Related Issues

Part of [https://github.com/open-feature/.github/issues/65](https://github.com/open-feature/.github/issues/65), [https://github.com/open-feature/php-sdk/issues/140](https://github.com/open-feature/php-sdk/issues/140), [https://github.com/open-feature/python-sdk/issues/403](https://github.com/open-feature/python-sdk/issues/403), [https://github.com/open-feature/dotnet-sdk/issues/328](https://github.com/open-feature/dotnet-sdk/issues/328), [https://github.com/open-feature/java-sdk/issues/1246](https://github.com/open-feature/java-sdk/issues/1246), and [https://github.com/open-feature/js-sdk/issues/1109](https://github.com/open-feature/js-sdk/issues/1109)

### Follow-up Tasks
Implement the test steps in the repos

### DIFF
--- a/specification/assets/gherkin/evaluation.feature
+++ b/specification/assets/gherkin/evaluation.feature
@@ -3,7 +3,7 @@ Feature: Flag evaluation
 # This test suite contains scenarios to test the flag evaluation API.
 
   Background:
-    Given a provider is registered
+    Given a provider is registered with cache disabled
 
   # basic evaluation
   Scenario: Resolves boolean value
@@ -48,6 +48,19 @@ Feature: Flag evaluation
     Then the resolved object details value should be contain fields "showImages", "title", and "imagesPerPage", with values "true", "Check out these pics!" and 100, respectively
     And the variant should be "template", and the reason should be "STATIC"
 
+  Scenario: Passes evaluation details to finally hooks
+    When a hook is added to the client
+    And a boolean flag with key "boolean-flag" is evaluated with details and default value "false"
+    Then non-error hooks should be called
+    And "after, finally after" hook should have evaluation details
+      | flag_type | key           | value        |
+      | string    | flag_key      | boolean-flag |
+      | boolean   | value         | true         |
+      | string    | variant       | on           |
+      | string    | reason        | STATIC       |
+      | string    | error_code    | None         |
+      | string    | error_message | None         |
+
   # context-aware evaluation
   Scenario: Resolves based on context
     When context contains keys "fn", "ln", "age", "customer" with values "Sulisław", "Świętopełk", 29, "false"
@@ -57,11 +70,31 @@ Feature: Flag evaluation
 
   # errors
   Scenario: Flag not found
-    When a non-existent string flag with key "missing-flag" is evaluated with details and a default value "uh-oh"
+    When a hook is added to the client
+    And a non-existent string flag with key "missing-flag" is evaluated with details and a default value "uh-oh"
     Then the default string value should be returned
     And the reason should indicate an error and the error code should indicate a missing flag with "FLAG_NOT_FOUND"
+    And error hooks should be called
+    And "finally after" hook should have evaluation details
+      | type   | key           | value                         |
+      | string | flag_key      | missing-flag                  |
+      | string | value         | uh-oh                         |
+      | string | variant       | None                          |
+      | string | reason        | ERROR                         |
+      | string | error_code    | ErrorCode.FLAG_NOT_FOUND      |
+      | string | error_message | Flag 'missing-flag' not found |
 
   Scenario: Type error
-    When a string flag with key "wrong-flag" is evaluated as an integer, with details and a default value 13
+    When a hook is added to the client
+    And a string flag with key "wrong-flag" is evaluated as an integer, with details and a default value 13
     Then the default integer value should be returned
     And the reason should indicate an error and the error code should indicate a type mismatch with "TYPE_MISMATCH"
+    And error hooks should be called
+    And "finally after" hook should have evaluation details
+      | type    | key           | value                                             |
+      | string  | flag_key      | wrong-flag                                        |
+      | integer | value         | 13                                                |
+      | string  | variant       | None                                              |
+      | string  | reason        | ERROR                                             |
+      | string  | error_code    | ErrorCode.TYPE_MISMATCH                           |
+      | string  | error_message | Expected type <class 'int'> but got <class 'str'> |
--- a/specification/assets/gherkin/evaluation.feature
+++ b/specification/assets/gherkin/evaluation.feature
@@ -48,11 +48,11 @@ Feature: Flag evaluation
     Then the resolved object details value should be contain fields "showImages", "title", and "imagesPerPage", with values "true", "Check out these pics!" and 100, respectively
     And the variant should be "template", and the reason should be "STATIC"
 
-  Scenario: Passes evaluation details to finally hooks
+  Scenario: Passes evaluation details to after and finally hooks
     When a hook is added to the client
     And a boolean flag with key "boolean-flag" is evaluated with details and default value "false"
-    Then non-error hooks should be called
-    And "after, finally after" hook should have evaluation details
+    Then "before" hooks should be called
+    And "after, finally after" hooks should be called with evaluation details
       | flag_type | key           | value        |
       | string    | flag_key      | boolean-flag |
       | boolean   | value         | true         |
@@ -74,8 +74,8 @@ Feature: Flag evaluation
     And a non-existent string flag with key "missing-flag" is evaluated with details and a default value "uh-oh"
     Then the default string value should be returned
     And the reason should indicate an error and the error code should indicate a missing flag with "FLAG_NOT_FOUND"
-    And error hooks should be called
-    And "finally after" hook should have evaluation details
+    Then "before, error" hooks should be called
+    And "finally after" hooks should be called with evaluation details
       | type   | key           | value                         |
       | string | flag_key      | missing-flag                  |
       | string | value         | uh-oh                         |
@@ -89,8 +89,8 @@ Feature: Flag evaluation
     And a string flag with key "wrong-flag" is evaluated as an integer, with details and a default value 13
     Then the default integer value should be returned
     And the reason should indicate an error and the error code should indicate a type mismatch with "TYPE_MISMATCH"
-    And error hooks should be called
-    And "finally after" hook should have evaluation details
+    Then "before, error" hooks should be called
+    And "finally after" hooks should be called with evaluation details
       | type    | key           | value                                             |
       | string  | flag_key      | wrong-flag                                        |
       | integer | value         | 13                                                |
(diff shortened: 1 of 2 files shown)


specification/assets/gherkin/evaluation.feature
```diff
@@ -57,11 +70,31 @@ Feature: Flag evaluation
 
   # errors
   Scenario: Flag not found
-    When a non-existent string flag with key "missing-flag" is evaluated with details and a default value "uh-oh"
+    When a hook is added to the client
```
aepfli, line 73: should this maybe be a "Given"?

```
Given a client hook
```

specification/assets/gherkin/evaluation.feature
```diff
@@ -48,6 +48,19 @@ Feature: Flag evaluation
     Then the resolved object details value should be contain fields "showImages", "title", and "imagesPerPage", with values "true", "Check out these pics!" and 100, respectively
     And the variant should be "template", and the reason should be "STATIC"
 
+  Scenario: Passes evaluation details to after and finally hooks
+    When a hook is added to the client
+    And a boolean flag with key "boolean-flag" is evaluated with details and default value "false"
```
aepfli, line 53: this clashes with your other pr in writing, there are differences: 
```
 Given a Boolean-flag with key "metadata-flag" and a default value "true"
```

hence we should rework this to follow the same structure as the metadata pr

specification/assets/gherkin/evaluation.feature
```diff
@@ -57,11 +70,31 @@ Feature: Flag evaluation
 
   # errors
   Scenario: Flag not found
-    When a non-existent string flag with key "missing-flag" is evaluated with details and a default value "uh-oh"
+    When a hook is added to the client
+    And a non-existent string flag with key "missing-flag" is evaluated with details and a default value "uh-oh"
```
aepfli, line 74: i do not think we need to deviate from the ` Given a Boolean-flag with key "metadata-flag" and a default value "true"` format, if the flag is missing, we know that from the key, but i do not think we need to add the `non-existent` part

specification/assets/gherkin/evaluation.feature
```diff
@@ -57,11 +70,31 @@ Feature: Flag evaluation
...
     Then the default string value should be returned
     And the reason should indicate an error and the error code should indicate a missing flag with "FLAG_NOT_FOUND"
+    Then "before, error" hooks should be called
```
aepfli, line 77: each hook should be an own step, as it makes writing and reusing steps easier

```
Then the before hook should be called
And the error hook should be called
```

specification/assets/gherkin/evaluation.feature
```diff
@@ -48,6 +48,19 @@ Feature: Flag evaluation
     Then the resolved object details value should be contain fields "showImages", "title", and "imagesPerPage", with values "true", "Check out these pics!" and 100, respectively
     And the variant should be "template", and the reason should be "STATIC"
 
+  Scenario: Passes evaluation details to after and finally hooks
```
aepfli, line 51: is this part of the evaluation, or should those new hooks be part of an "hooks.feature" annotated with `@hooks`? i think this is a little bit separated from the default evaluation and more an own thing in the spec

chrfwow:
I extracted the hook tests into their own file` hooks.feature`, and left the `evaluation.feature` file mostly as-is

specification/assets/gherkin/evaluation.feature
```diff
@@ -3,7 +3,7 @@ Feature: Flag evaluation
 # This test suite contains scenarios to test the flag evaluation API.
 
   Background:
-    Given a provider is registered
+    Given a provider is registered with cache disabled
```
aepfli, line 6:
```suggestion
    Given a stable provider
```

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,52 @@
+@hooks
+Feature: Evaluation details through hooks
+
+# This test suite contains scenarios to test the functionality of hooks.
+
+  Background:
+    Given a provider is registered with cache disabled
```
aepfli, line 7:
```suggestion
    Given a stable provider
```

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,52 @@
...
+
+  Scenario: Passes evaluation details to after and finally hooks
+    Given a client with added hook
+    And a boolean-flag with key "boolean-flag" and a default value "false"
+    When the flag was evaluated with details
+    Then "before" hooks should be called
```
aepfli, line 13: is it one hook? or multiple ones, for me it looks more like 
```
Then the "before" hook should be called
```

of to be more in sync with the already existing flagd feature files (https://github.com/open-feature/flagd-testbed/blob/main/gherkin/events.feature#L27)

```
Then the "before" hook should have been executed
```

but i am not sure about this naming

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,52 @@
...
+    And "after, finally after" hooks should be called with evaluation details
```
aepfli, line 14: sorry, maybe i am missing insight but what is `after` and `finally after`

should this maybe be like 
```
And the "after" hook should be called
And the hook should receive evaluation details
```

the reason is, that this is somehow two verification steps, mixed into one. it makes sense if this happens more often, but the first part is actually already covered by the line above.

aepfli (commented):
I still have some suggestions, and most of them are related to my limited knowledge ;) but maybe some are worth investigating

specification/assets/gherkin/hooks.feature
chrfwow: We have `after` hooks that are called after every _successful_ evaluation and we have `finally after` hooks that are called after _every_ evaluation. Since this is a positive test, both of these hooks should be executed with the same evalution details

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,52 @@
...
+    Then the "before" hook should have been executed
+    And the "error" hook should have been executed
+    And the "finally after" hooks should be called with evaluation details
+      | data_type | key           | value                                             |
+      | string    | flag_key      | wrong-flag                                        |
+      | integer   | value         | 13                                                |
+      | string    | variant       | None                                              |
+      | string    | reason        | ERROR                                             |
+      | string    | error_code    | ErrorCode.TYPE_MISMATCH                           |
+      | string    | error_message | Expected type <class 'int'> but got <class 'str'> |
```
beeme1mr, line 52: Do we want to define the error message?

chrfwow: Hmm, the error message might be very implementation specific, so I agree with you. @aepfli @toddbaert what do you think?

specification/assets/gherkin/hooks.feature
aepfli: The general recommendation for gherkin files is, to do less in one step so they are easier reusable. this is a tricky one, as i can see the benefit of both sides. and most likely we will see within the implemenation phase of the gherkin test if this is working as is, or if it would be easier to split it into dedicated steps per hooks. I am fine with it as is currently

specification/assets/gherkin/hooks.feature
aepfli: open for both approaches, i do think that normalized error message could be handy, but we should remove any language specific synatx. eg "Expected type Integer but got String" so we can easier map this in other languages to a desired message. But i like the idea of standardized messages throughout the sdks.

chrfwow: I agree with the standardized error messages. However, I do think this would be a nontrivial change. Should we do this in a follow up issue? Then I would remove the error message from this PR

aepfli (approved):

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,49 @@
...
+    And a boolean-flag with key "boolean-flag" and a default value "false"
+    When the flag was evaluated with details
+    Then the "before" hook should have been executed
+    And the "after, finally" hooks should be called with evaluation details
+      | data_type | key           | value        |
+      | string    | flag_key      | boolean-flag |
+      | boolean   | value         | true         |
+      | string    | variant       | on           |
+      | string    | reason        | STATIC       |
+      | string    | error_code    | None         |
```
beeme1mr, line 20: Should `error_code` be omitted from the table?

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,49 @@
...
+    When the flag was evaluated with details
+    Then the "before" hook should have been executed
+    And the "error" hook should have been executed
+    And the "finally" hooks should be called with evaluation details
+      | data_type | key           | value                         |
+      | string    | flag_key      | missing-flag                  |
+      | string    | value         | uh-oh                         |
+      | string    | variant       | None                          |
+      | string    | reason        | ERROR                         |
+      | string    | error_code    | ErrorCode.FLAG_NOT_FOUND      |
```
beeme1mr, line 35:
```suggestion
      | string    | error_code    | FLAG_NOT_FOUND      |
```

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,49 @@
...
+    When the flag was evaluated with details
+    Then the "before" hook should have been executed
+    And the "error" hook should have been executed
+    And the "finally" hooks should be called with evaluation details
+      | data_type | key           | value                                             |
+      | string    | flag_key      | wrong-flag                                        |
+      | integer   | value         | 13                                                |
+      | string    | variant       | None                                              |
+      | string    | reason        | ERROR                                             |
+      | string    | error_code    | ErrorCode.TYPE_MISMATCH                           |
```
beeme1mr, line 49:
```suggestion
      | string    | error_code    | TYPE_MISMATCH                           |
```

specification/assets/gherkin/hooks.feature
chrfwow: I would leave it in to verify that no error code is given when there is no error

specification/assets/gherkin/hooks.feature
```diff
@@ -0,0 +1,49 @@
...
+    Given a client with added hook
+    And a string-flag with key "missing-flag" and a default value "uh-oh"
+    When the flag was evaluated with details
+    Then the "before" hook should have been executed
+    And the "error" hook should have been executed
+    And the "finally" hooks should be called with evaluation details
+      | data_type | key        | value          |
+      | string    | flag_key   | missing-flag   |
+      | string    | value      | uh-oh          |
+      | string    | variant    | None           |
```
aepfli, line 33: `None` is pretty python specific - i would instead use "null" which is more present in a lot of languages, compared to None and it is IMHO more descriptive

specification/assets/gherkin/hooks.feature
chrfwow: I would leave it in to verify that no error code is given when there is no error

beeme1mr (approved):

PR #304, opened 2025-03-21
fix: Clarify scenario descriptions in context merging feature

### LINKED ISSUE #303: Duplicate tests in contextMerging.feature
Some tests in `contextMerging.feature` are duplicated, and differ in a few test cases to cover implementations that support any combination of hooks and transaction context.
E.g., we have a test
`Scenario: A context entry is added to each level with different keys`
which exists with tags:
```
  @transaction
  Scenario: A context entry is added to each level with different keys
```
```
  @hooks
  Scenario: A context entry is added to each level with different keys
```
```
  @hooks @transaction
  Scenario: A context entry is added to each level with different keys
```
If an implementation supports both hooks and transactions, this can lead to problems as none of the tests will be excluded, and there will be duplicated names.

### Acceptance criteria
Have an unique name for each of the tests, regardless of which tags are active or inactive

askpt:
## This PR

This pull request includes updates to the `specification/assets/gherkin/contextMerging.feature` file to improve the clarity of scenario descriptions by specifying the context (transaction, hook, or both) in which the context entries are added or overwritten.

Changes to scenario descriptions:

* Updated the scenario description to specify that the context entry is added for a transaction.
* Updated the scenario description to specify that the context entry is added for a hook.
* Updated the scenario description to specify that the context entry is added for both a transaction and a hook.
* Updated the scenario outline description to specify that the context entry overwrites values for a transaction.
* Updated the scenario outline description to specify that the context entry overwrites values for a hook.
* Updated the scenario outline description to specify that the context entry overwrites values for both a transaction and a hook.

### Related Issues

Fixes #303

### DIFF
--- a/specification/assets/gherkin/contextMerging.feature
+++ b/specification/assets/gherkin/contextMerging.feature
@@ -34,7 +34,7 @@ Feature: Context merging precedence
       | Before Hooks |
 
   @transaction
-  Scenario: A context entry is added to each level with different keys
+  Scenario: For a transaction, a context entry is added to each level with different keys
     Given A context entry with key "API" and value "API value" is added to the "API" level
     And A context entry with key "Transaction" and value "Transaction value" is added to the "Transaction" level
     And A context entry with key "Client" and value "Client value" is added to the "Client" level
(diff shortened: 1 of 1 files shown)


chrfwow (approved):

toddbaert:
Merging as is since this is just minor test changes.

PR #312, opened 2025-06-05
docs: explicitly mark the Hooks section experimental

sahidvelji:
## This PR

Even though the README says that no explicit status implies experimental status, I think we should explicitly mark the Hooks section experimental to align with the other sections.


beeme1mr (approved):
I suppose it's good to be explicit but we should try to prioritize moving these key features to a higher stability level.

kinyoklion (approved):

lukas-reining (approved):

PR #313, opened 2025-06-05
feat: add standardized Propagation Evaluation to Flag Metadata.

aepfli:
Within a lot of system flag changes are propagated asynchronously to the consumers (not during evaluations). This propagations can be enhanced via spans and traces on their own to see how flag changes are distributed throughout the system. 

But in the end, as an enduser i am curious, which propagation event actually is linked to my evaluation. It is possible to somehow determine this via version, but spans and traces offer a deeper insight. If we standardize this behaviour, we can create out of the box OpenTelemetry Configuration Changed Listeners, to make this feature usable with all our providers.

## This PR

- adds this new feature

### Related Issues

Relates: https://github.com/open-feature/flagd/pull/1595


specification/appendix-d-observability.md
```diff
@@ -51,6 +51,15 @@ The following describes how keys in [flag metadata](types.md#flag-metadata) are
 | ---------------------------- | ----------------------- | ----------------- | -------- | ------------------------------------------------------------------------------------------------ |
 | `feature_flag.provider.name` | `name`                  | `Recommended`     | `string` | The name of the provider as defined in the `provider metadata`, available in the `hook context`. |
 
+### Propagation Metadata
+
+Feature Flags are propagated through different systems with different methods. Often this updates have an asynchronous nature to the evaluation and do not correlate directly (eg. cached values or in-process evaluations). For distributed systems it is important to reflect how changes are populate to all systems, and how those correlate with evaluations. In a simple manner the version could be used to achieve this, but offers additional and more complex solution to correlate the data. Instead we are defining two additional metadata properties `propagationTraceId` and `propagationSpanId` which can be used to link evaluation spans to propagation spans.
```
lukas-reining:
```suggestion
Feature Flags are propagated through different systems with different methods. Often these updates have an asynchronous nature to the evaluation and do not correlate directly to it (eg. cached values or in-process evaluations). For distributed systems it is important to reflect how changes in flag configurations are propagated to all systems, and how those correlate with evaluations. In a simple manner the version could be used to achieve this, but offers additional and more complex solution to correlate the data. Instead we are defining two additional metadata properties `propagationTraceId` and `propagationSpanId` which can be used to link evaluation spans to propagation spans.
```

lukas-reining: > Instead we

I would not use first person here.

specification/appendix-d-observability.md
```diff
@@ -51,6 +51,15 @@ The following describes how keys in [flag metadata](types.md#flag-metadata) are
...
+
+| Provider Metadata Field | Requirement level | Type     | Notes                                  |
```
lukas-reining: Do you mean this provider metadata: https://openfeature.dev/specification/sections/providers#requirement-211
Or the one for the evaluations? https://openfeature.dev/specification/sections/providers#requirement-2210

Or am I on the completely wrong track?

specification/appendix-d-observability.md
lukas-reining: > `propagationTraceId` and `propagationSpanId`

I have some things that are not fully clear to me from reading this:

How are these defined? 
Do we typically fill these with the OTEL values? Which span do we use then? Trace would probably be the root one?
How are we setting them?

lukas-reining: > In a simple manner the version could be used to achieve this, but offers additional and more complex solution to correlate the data. 

Do you mean the version that we defined on the OTEL semconv?
What do you mean by "offers additional and more complex solution"?

lukas-reining (commented):
I like the idea @aepfli!
From reading this I am not 100% sure how it would would exactly, left some questions.
The example answers some of the questions but from the spec I do not fully understand how it is meant to work.

specification/appendix-d-observability.md
lukas-reining: From reading the example I would prefer using something like `configuration changed event attribute` or something like this, ore make clear before that these fields are meant for the event details of the `configuration changed event` of the provider.

specification/appendix-d-observability.md
aepfli: In appendix D, we specify that flag metadata can contain a `version` https://openfeature.dev/specification/appendix-d#flag-metadata - theoretically,y we could use the version field too, to link the propagating span/trace, to the evaluation span/trace.

aepfli: The propagationTraceId and propagationSpanId are flexible. eg. in flagd we would create a trace and span for each GRPC event and add this information to the payload. eg. GRPC streams don't offer a way to propagate headers for events only, they have to be part of the payload. Other connection methods might autopropagate them if possible. But with the persistence in the metadata, we can link the propagating span to the evaluating span (which might be two totally different occasions/traces).

I am not the best one, when writing specs, this is my first attempt, and I am happy to explain my thoughts, and maybe this will help to solve the confusion

specification/appendix-d-observability.md
aepfli: there are two parts to it.

1. I want to track the propagation, and see when the propagation is done on the consumer.
2. I want to link evaluations to the propagation event, to have a whole view at the evaluation, which trace/span actually brought the data into provider

specification/appendix-d-observability.md
lukas-reining: > In appendix D, we specify that flag metadata can contain a version

Yeah, I mean, that I would love this to be a bit more clear that this version is used. 

> theoretically,y we could use the version field too, to link the propagating span/trace, to the evaluation span/trace.

Okay, but what what do you mean with: `but offers additional and more complex solution to correlate the data.`?

lukas-reining: > But with the persistence in the metadata, we can link the propagating span to the evaluating span (which might be two totally different occasions/traces).

I get that concept. 

What I am not sure about is, how do we define all these things exactly. 
E.g. for the `propagating span` I can imagine multiple definitions, one of them being `the span of the incoming http request". 
This span id e.g. will be different between all services, the traceparent id in otel might be the same between all of the services. This part is not shown in your example. 

Maybe it is good enough to add 1 or 2 good examples for a good value for the ids.

aepfli: we could use this version to correlate, but this proposal adds another attribute, which contains the traceid and spanid of the propagating cause.

eg. in flagd in-process, we have two non-connected steps for evaluations.

1. Flag configurations are distributed asynchronously. When Flagd detects a change in the flag source, it sends out a new configuration, which will have a version attribute in the metadata. 
2. When I do evaluations, I can create a new span, and if I want to correlate this with the span of the propagation, I need to check for the version attribute, and I don't have a direct link. Otel offers to link traces, to represent a correlation. With the information of the trace and span, we can create this link. And create a holistic image from propagation span to evaluation span, and how it is linked.

chrfwow (approved):

dyladan (commented):
It looks like this PR is attempting to recreate span links. Any reason you chose to go this route rather than using that mechanism? https://opentelemetry.io/docs/concepts/signals/traces/#span-links

aepfli:
```mermaid
sequenceDiagram
    participant FS as Flag Source
    participant P as Provider
    participant SDK as OpenFeature SDK
    participant App as Application
    participant OTel as OTel Backend

    Note over FS,P: Flag change detected

    alt Flag Source provides trace context
        FS->>OTel: Start propagation span<br/>[traceId: T1, spanId: S1]
        FS->>+P: gRPC stream message<br/>(payload includes T1, S1)
        FS->>OTel: End propagation span
        P->>P: Pass through T1, S1<br/>into flag metadata
    else Provider creates trace context
        FS->>+P: gRPC stream message<br/>(no trace context)
        P->>OTel: Start propagation span<br/>[traceId: T1, spanId: S1]
        P->>P: Write into flag metadata:<br/>propagationTraceId = T1<br/>propagationSpanId = S1
        P->>OTel: End propagation span
    end

    P->>-SDK: ConfigurationChanged event<br/>(flag metadata with T1, S1)

    Note over SDK,App: Later — independent evaluation

    App->>+SDK: evaluateFlag("my-flag", context)

    SDK->>SDK: Read flag metadata<br/>propagationTraceId = T1<br/>propagationSpanId = S1

    SDK->>OTel: Start evaluation span<br/>+ SpanLink → [T1, S1]

    SDK->>-App: return flag value

    OTel->>OTel: Link: propagation span ↔ evaluation span
```

Issue #270, opened 2024-09-05
Requirement 2.5.2 might not be relevant anymore


fabriziodemaria:
Link to requirement [here](https://github.com/open-feature/spec/blob/80ceaa0465ac17ad88856550f284a93e36a7f382/specification/sections/02-providers.md?plain=1#L221-L228). 
It reads:

> After a provider's shutdown function has terminated successfully, the provider's state MUST revert to its uninitialized state.
If a provider requires initialization, once it's shut down, it must transition to its initial NOT_READY state. Some providers may allow reinitialization from this state. Providers not requiring initialization are assumed to be ready at all times.

According to [recent spec changes](https://github.com/open-feature/spec/pull/241), the `Provider Status` is now meant to be maintained by the SDK, so it's not a Provider's responsibility to transition to `NOT_READY` state.

lukas-reining:
> If a provider requires initialization, once it's shut down, it must transition to its initial NOT_READY state. Some providers may allow reinitialization from this state. Providers not requiring initialization are assumed to be ready at all times.

Mh, I think the content here is right but has to be adjusted to "the clients status accessor" instead of "providers status".
But it is a good point, I think we should adjust this and move it to section one.
Does that make sense?

toddbaert:
I will address this before we merge our hardening PR. Thanks @fabriziodemaria

Issue #164, opened 2022-11-27
Hook evaluation specification clarity


tcarrio:
This is in reference to the hooks evaluation order in [**Requirement 4.4.2**](https://docs.openfeature.dev/docs/specification/sections/hooks#requirement-442). This reads:

> Hooks **MUST** be evaluated in the following order:
>
> - before: API, Client, Invocation, Provider
> - after: Provider, Invocation, Client, API
> - error (if applicable): Provider, Invocation, Client, API
> - finally: Provider, Invocation, Client, API

Currently this is implemented differently between the Python and JavaScript SDKs, but neither is technically wrong based on the loose specification.

The JS SDK generates the `before` list of hooks, then reverses it for the `after`, `error`, and `finally`. This appeases the overall ordering of hook executions as listed above. Hence, the `before` is `[...API, ...Client, ...Invocation, ...Provider]`. The `after` is `[...API, ...Client, ...Invocation, ...Provider].reverse()`. So the `after` hooks will have all of the `Provider` hooks execute before the `Invocation` hooks, those before the `Client` hooks, and those before the `API` hooks.

The Python SDK generates the `before` list of hooks and the `after/error/finally` list of hooks by spreading them in their own declarations. So, `before` is `API + Client + Invocation + Provider` and the `after` is `Provider + Invocation + Client + API`. So the `after` hooks satisfy the same as mentioned above for the JS SDK.

The issue here is that the overall hook execution order would vary if any of these hooks have more than one entry. For example:

```
API = [A, B]
Client = [C, D]
Invocation = [E, F]
Provider = [G, H]

# `before` list
before_hooks = API + Client + Invocation + Provider

# spreading directly (like Python)
list1 = Provider + Invocation + Client + API
# results in [G, H, E, F, C, D, A, B]

# reversing the before list (like JavaScript)
list2 = before_hooks[:]
list2.reverse()
# results in [H, G, F, E, D, C, B, A]
```

This detail should be clear from the specification so there is not any discrepancy between implementations of the API.

toddbaert:
I think this was a conscious choice, not an oversight. Every specified behavior means tests and maintenance so there's a cost associated with specifying in such detail. It would take some time to dig up the conversation, but we felt specifying ordering of hooks between stages was a bad idea.

[...]

PR #314, opened 2025-06-09
feat: stabilize - evaluation, providers, harden - hooks, events, context

toddbaert:
This PR: 

* stabilizes evaluation API
* stabilizes providers
* hardens hooks, events and context
* hardens all "static context" (client-side) points

This is a pretty substantial step.
`stable` means that we have no more leeway at all in SDKs - **any breaking changes to the parts of the SDK implementing stable APIs means a new major version of the SDK is required, without exception**.

I've also opted to "harden" the `static context paradigm` since we have numerous client implementations in the wild, in particular React and Angular, but also Android. There could still be breaking changes here but the bar is higher. I think this is how we're acting already with regards to these APIs, in all honesty.

I did not promote the [context propagation](https://openfeature.dev/specification/sections/evaluation-context#33-context-propagation) sub-section, or the new tracking section - these are _relatively_ new, though I'm open to opinions on hardening these further.

Potential blockers:

- https://github.com/open-feature/go-sdk/issues/389
- https://github.com/open-feature/spec/issues/270

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -6,7 +6,7 @@ toc_max_heading_level: 4
 
 # 1. Flag Evaluation API
 
-[![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
+[![stable](https://img.shields.io/static/v1?label=Status&message=stable&color=green)](https://github.com/open-feature/spec/tree/main/specification#stable)
 
 ## Overview
 
@@ -203,7 +203,7 @@ See [evaluation context](./03-evaluation-context.md) for details.
 
 #### Condition 1.3.2
 
-[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+[![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
 
 > The implementation uses the static-context paradigm.
 
@@ -286,7 +286,7 @@ FlagEvaluationDetails<MyStruct> myStructDetails = client.getObjectDetails<MyStru
 
 #### Condition 1.4.2
 
-[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+[![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
 
 > The implementation uses the static-context paradigm.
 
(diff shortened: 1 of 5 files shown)


kinyoklion:
> One other thing to point out is that [hook data](https://openfeature.dev/specification/sections/hooks/#46-hook-data) isn't widely implemented yet, so we could mark that subsection experimental... however I don't see much risk with it.

Yeah, it is very new so that seems like a good idea.

specification/sections/04-hooks.md
```diff
@@ -6,7 +6,7 @@ toc_max_heading_level: 4
 
 # 4. Hooks
 
-[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+[![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
```
beeme1mr, line 9: There is an open question on the order hooks that should execute. We should either address the concern in the spec or close the issue with our reasoning before changing the status of this section.

https://github.com/open-feature/spec/issues/164

specification/sections/01-flag-evaluation.md
```diff
@@ -6,7 +6,7 @@ toc_max_heading_level: 4
 
 # 1. Flag Evaluation API
 
-[![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
+[![stable](https://img.shields.io/static/v1?label=Status&message=stable&color=green)](https://github.com/open-feature/spec/tree/main/specification#stable)
```
guidobrei, line 9: What does it mean, for the Flag Evaluation API section to be `stable` if subsections are still in `experimental` or `hardening`. If every subsection has its own lifecycle, shouldn't we then set the status only on subsection?

jonathannorris (approved):

specification/sections/01-flag-evaluation.md
toddbaert: Basically, the sub-sections override the overall document level. This is implied [here](https://github.com/open-feature/spec/tree/main/specification#document-statuses). This is specifically for new and well-scoped features within existing documents. For instance, the `provider` section is quite old and stable, but there's a new point or two about `tracking` which are experimental - we would allow changes to the tracking subsection but not the overall section.

specification/sections/04-hooks.md
lukas-reining: I replied there, I would vote for adding the order in this PR to have consistent behavior between languages.

lukas-reining (commented):
I would vote for addressing #164, and then merge.

gruebel (approved):

specification/sections/04-hooks.md
toddbaert: I've opened https://github.com/open-feature/spec/pull/315

chrfwow (approved):

toddbaert:
> Another small blocker [here](https://github.com/open-feature/spec/issues/270).

I've opened https://github.com/open-feature/spec/pull/317 to address this.

PR #315, opened 2025-06-10
fix: clarify hook ordering

toddbaert:
This clarifies some ambiguity in hook execution order, mentioned in https://github.com/open-feature/spec/issues/164:

> The issue here is that the overall hook execution order would vary if any of these hooks have more than one entry. For example:
> 
> ```
> API = [A, B]
> Client = [C, D]
> Invocation = [E, F]
> Provider = [G, H]
> 
> # `before` list
> before_hooks = API + Client + Invocation + Provider
> 
> # spreading directly (like Python)
> list1 = Provider + Invocation + Client + API
> # results in [G, H, E, F, C, D, A, B]
> 
> # reversing the before list (like JavaScript)
> list2 = before_hooks[:]
> list2.reverse()
> # results in [H, G, F, E, D, C, B, A]
> ```

This change makes clear that `[H, G, F, E, D, C, B, A]` is the correct answer here, which is the more intuitive choice, in my opinion.

This is also one of the few (only?) normative spec points that isn't a single sentence, but instead has point-form clauses, so I've also attempted to fix that.

I've work-shopped this with a few people and they interpreted it correctly, though admittedly, the term "stack-wise" is doing a lot of work here. I'm hopeful that our audience is familiar enough with that sort of term that the intent is clear.

I believe there's a few SDKs where this is incorrectly implemented, I'll audit those and create issues if this is merged. I don't expect that such an implementation change will significantly impact anyone.

### DIFF
--- a/specification/sections/04-hooks.md
+++ b/specification/sections/04-hooks.md
@@ -252,6 +252,19 @@ Before flag resolution (the `before` stage), hooks run in the order `API` -> `Cl
 After flag evaluation (the `after`, `error`, or `finally` stages), hooks run in the order `Provider` -> `Invocation` -> `Client` -> `API`, and within those, in reverse of the order in which they were added.
 This achieves intuitive "stack-like" or "LIFO" behavior for side effects and transformations.
 
+```
+Given hooks A - H, each implementing the both the `before` and `after` stages, added at the following levels and order:
+
+API: [A, B]
+Client: [C, D]
+Invocation: [E, F]
+Provider: [G, H]
+
+The expected order of execution is:
+
+A.before -> B.before -> C.before -> D.before -> E.before -> F.before -> G.before -> H.before -> flagResolution -> H.after -> G.after -> F.after -> E.after -> D.after -> C.after -> B.after -> A.after
+```
+
 #### Requirement 4.4.3
 
 > If a `finally` hook abnormally terminates, evaluation **MUST** proceed, including the execution of any remaining `finally` hooks.


beeme1mr (approved):

lukas-reining:
If I see it correctly, Java is doing FIFO: https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/HookSupport.java#L22

gruebel:
> > Looks good, thanks!
> > Do you know of the top of your head if any SDKs do it differently right now @toddbaert?
> 
> I think Python is not compliant to this at the moment. I'm continuing to investigate other SDKs.

I checked and Python does it in the proposed order. It was fixed a while back via open-feature/python-sdk#71 by @tcarrio 😄

gruebel (approved):

toddbaert:
> I think it may be worth something like the explicit example from the PR body in the spec text.
> 
> I especially think it should be that clear that if you have one list of hooks in "before" order, that iterating that list forward is "before" and "after" is equivalent to iterating it in reverse.
> 
> This also allows you to implement this in a simple way without generating intermediate objects or garbage to have a reversed form. Just start at 0 and increment, or start at the last index and decrement. And it also facilitates implementing hook data in an efficient way with an array equal to the size of the number of hooks.

Ya I think that's a good idea. I can add something like that.

toddbaert:
> > I think it may be worth something like the explicit example from the PR body in the spec text.
> > I especially think it should be that clear that if you have one list of hooks in "before" order, that iterating that list forward is "before" and "after" is equivalent to iterating it in reverse.
> > This also allows you to implement this in a simple way without generating intermediate objects or garbage to have a reversed form. Just start at 0 and increment, or start at the last index and decrement. And it also facilitates implementing hook data in an efficient way with an array equal to the size of the number of hooks.
> 
> Ya I think that's a good idea. I can add something like that to the non-normative section.

Done.

chrfwow (approved):

askpt (approved):

toddbaert:
> Thank you. Suggestion from my side would be to also think about a gherkin test suite. this seems to be like a really great fit

Totally support this, but I think it's a different effort.

PR #317, opened 2025-06-19
fix: correct shutdown points after status removal

toddbaert:
This is a minor point of clarification with no material changes.

When we removed `status` from the providers, we didn't migrate this point to the evaluation API, as @fabriziodemaria pointed out [here](https://github.com/open-feature/spec/issues/270).

Fixes: https://github.com/open-feature/spec/issues/270

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -513,3 +513,9 @@ see: [error codes](../types.md#error-code), [flag value resolution](./02-provide
 The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
 
 see: [error codes](../types.md#error-code)
+
+#### Requirement 1.7.9
+
+> The client's `provider status` accessor **MUST** indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.
+
+Regardless of the success of the provider's `shutdown` function, the `provider status` should convey the provider is no longer ready to use once the shutdown function terminates.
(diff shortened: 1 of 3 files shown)


specification/sections/02-providers.md
```diff
@@ -225,9 +225,10 @@ class MyProvider implements Provider, AutoDisposable {
 
 #### Requirement 2.5.2
 
-> After a provider's shutdown function has terminated successfully, the provider's state **MUST** revert to its uninitialized state.
+> After a provider's shutdown function has terminated, the provider **SHOULD** revert to its uninitialized state.
```
toddbaert, line 228: I've re-worded this to simply suggest that it's ideal (but not strictly required) that providers can be re-initialized after shutdown.

It's already possible some providers don't support this, and I think it's quite fine as long as they throw when re-initialized.

specification/sections/01-flag-evaluation.md
```diff
@@ -513,3 +513,9 @@ see: [error codes](../types.md#error-code), [flag value resolution](./02-provide
 The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
 
 see: [error codes](../types.md#error-code)
+
+#### Requirement 1.7.9
+
+> The client's `provider status` accessor **MUST** indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.
```
toddbaert, line 519: This is basically the old requirement, but relocated to the API, since providers themselves don't have this field.

This point is very similar to points `1.7.3` to `1.7.5` above.

chrfwow (approved):

kylejuliandev (approved):

thomaspoignant (approved):

aepfli (approved):

kinyoklion (approved):

PR #323, opened 2025-07-10
feat: add point 1.6.2, 2.5.3 about shutdown details

toddbaert:
This is a change that I suspect will have little practical impact for most users, but helps disambiguate some behavior that recently caused [issues](https://github.com/open-feature/go-sdk/issues/397) and [confusion](https://github.com/open-feature/go-sdk/pull/400/files#r2191214490).

This is only one means of addressing this concern, so please don't hesitate to put forward alternative proposals.

"Idempotent" may not be exactly the right word here, since it's only idempotent within the scope of one execution of the life-cycle, so I'm open to copy changes here as well.

# EDIT

:warning: I've substantially changed this PR based on feedback, and dismissed all approvals:

As @erka and others have pointed out, practically, `shutdown` has been used for more than just cleaning up providers, it's also used frequently to "reset" the API for testing purposes. I think this is a valid use-case and I don't see any reason why it can't be added to what (at least my understanding of) the original intent of the function was... so I've changed 1.6.2 to say that the `shutdown` now also resets the state of the API fully (removes hooks, providers, event handlers, etc) from the API.

As @chrfwow noted (and I was concerned about as well) guaranteeing that `shutdown` is not called while a provider is still starting up will be tricky to implement. Instead I've changed 1.6.1 to say we will unconditionally run shutdown on all providers, and also added a recommendation that providers handle this gracefully in the provider spec.

@sahidvelji I've also added a pre-amble about shutdown in general as you requested.

@lukas-reining @erka @beeme1mr @sahidvelji @chrfwow  please re-review.

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -407,7 +407,7 @@ It's recommended that application-authors call this function on application shut
 
 > The API **MUST** define a function to propagate a shutdown request to all providers.
 
-The global API object defines a `shutdown` function, which will call the respective `shutdown` function on the active providers.
+The global API object defines a `shutdown` function, which will call the respective `shutdown` function on all providers.
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
 This shutdown function unconditionally calls the shutdown function on all registered providers, regardless of their state. 
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -401,10 +401,19 @@ See [hooks](./04-hooks.md) for details.
 
 #### Requirement 1.6.1
 
-> The API **MUST** define a mechanism to propagate a shutdown request to active providers.
+> The API **MUST** define a function to propagate a shutdown request to active providers.
 
-The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the registered providers.
+The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the active providers.
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
+
+see: [`shutdown`](./02-providers.md#25-shutdown)
+
+#### Requirement 1.6.2
+
+> The API's `shutdown` function **MUST NOT** call shutdown on providers which are in state `NOT_READY`.
+
+With respect to the lifecycle of a given active provider, the global API object's `shutdown` function should be idempotent; multiple calls to `shutdown` should result in only a single execution of the provider's `shutdown` function.
+Implementations should take care to await or cancel the initialization of providers if `shutdown` is called while providers are still being initialized, and have yet to transition to `READY` or some other state.
 
 see: [`shutdown`](./02-providers.md#25-shutdown)
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -401,7 +401,7 @@ See [hooks](./04-hooks.md) for details.
 
 The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
 It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, providers, and setting a "No-op provider"; this is useful for testing purposes.
-It's recommended that application-authors to call this function on application shutdown, and after the completion of test suites which make use of the SDK.
+It's recommended that application-authors call this function on application shutdown, and after the completion of test suites which make use of the SDK.
 
 #### Requirement 1.6.1
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -399,23 +399,26 @@ See [hooks](./04-hooks.md) for details.
 
 [![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
 
+The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
+It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, and providers; this is useful for testing purposes.
+It's recommended that application-authors to call this function on application shutdown, and after the completion of test suites which make use of the SDK.
+
 #### Requirement 1.6.1
 
-> The API **MUST** define a function to propagate a shutdown request to active providers.
+> The API **MUST** define a function to propagate a shutdown request to all providers.
 
-The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the active providers.
+The global API object defines a `shutdown` function, which will call the respective `shutdown` function on the active providers.
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
+This shutdown function unconditionally calls the shutdown function on all registered providers, regardless of their state. 
 
 see: [`shutdown`](./02-providers.md#25-shutdown)
 
 #### Requirement 1.6.2
 
-> The API's `shutdown` function **MUST NOT** call shutdown on providers which are in state `NOT_READY`.
+> The API's `shutdown` function **MUST** reset all state of the API, removing all hooks, event handlers, and providers.
 
-With respect to the lifecycle of a given active provider, the global API object's `shutdown` function should be idempotent; multiple calls to `shutdown` should result in only a single execution of the provider's `shutdown` function.
-Implementations should take care to await or cancel the initialization of providers if `shutdown` is called while providers are still being initialized, and have yet to transition to `READY` or some other state.
-
-see: [`shutdown`](./02-providers.md#25-shutdown)
+After shutting down all providers, the `shutdown` function resets the state of the API.
+This is especially useful for testing purposes to restore the API to a known state.
 
 ### 1.7. Provider Lifecycle Management
 
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -400,7 +400,7 @@ See [hooks](./04-hooks.md) for details.
 [![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
 
 The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
-It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, and providers; this is useful for testing purposes.
+It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, providers, and setting a "No-op provider"; this is useful for testing purposes.
 It's recommended that application-authors to call this function on application shutdown, and after the completion of test suites which make use of the SDK.
 
 #### Requirement 1.6.1
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -225,11 +225,20 @@ class MyProvider implements Provider, AutoDisposable {
 
 #### Requirement 2.5.2
 
-> After a provider's shutdown function has terminated, the provider **SHOULD** revert to its uninitialized state.
+> After a provider's `shutdown` function has terminated, the provider **SHOULD** revert to its uninitialized state.
 
 If a provider requires initialization, once it's shut down, it must transition to its uninitialized state.
 Some providers may allow reinitialization from this state.
 Providers not requiring initialization are assumed to be ready at all times.
+Providers in the process of initializing abort initialization if shutdown is called while they are still starting up.
+
+see: [initialization](#24-initialization)
+
+#### Requirement 2.5.3
+
+> A Provider's `shutdown` function **SHOULD** be idempotent.
+
+If a provider's `shutdown` function has been called, subsequent calls (without an intervening call to `initialize`) should have no effect.
 
 see: [initialization](#24-initialization)
 
--- a/specification.json
+++ b/specification.json
@@ -283,15 +283,15 @@
         {
             "id": "Requirement 1.6.1",
             "machine_id": "requirement_1_6_1",
-            "content": "The API MUST define a function to propagate a shutdown request to active providers.",
+            "content": "The API MUST define a function to propagate a shutdown request to all providers.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.6.2",
             "machine_id": "requirement_1_6_2",
-            "content": "The API's `shutdown` function MUST NOT call shutdown on providers which are in state `NOT_READY`.",
-            "RFC 2119 keyword": "MUST NOT",
+            "content": "The API's `shutdown` function MUST reset all state of the API, removing all hooks, event handlers, and providers.",
+            "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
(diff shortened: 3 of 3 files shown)


specification/sections/01-flag-evaluation.md
```diff
@@ -401,13 +401,22 @@ See [hooks](./04-hooks.md) for details.
...
 
-The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the registered providers.
+The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the active providers.
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
 
 see: [`shutdown`](./02-providers.md#25-shutdown)
 
+#### Requirement 1.6.2
+
+> The API's `shutdown` function **MUST NOT** call shutdown on providers which are in state `NOT_READY`.
```
toddbaert: There's other ways of handling this, but I think for provider authors and users, this offers a very simple mental modal: "my provider will never be shut down while it's NOT_READY, including when it's initializing".

erka:
> Shutdown is meant to be called on application exit

This aspect is not covered in the specification. In practice, some developers call the global API shutdown after each their test, expecting a clean state before next test run.

specification/sections/01-flag-evaluation.md
```diff
@@ -401,13 +401,22 @@ See [hooks](./04-hooks.md) for details.
...
+
+With respect to the lifecycle of a given active provider, the global API object's `shutdown` function should be idempotent; multiple calls to `shutdown` should result in only a single execution of the provider's `shutdown` function.
+Implementations should take care to await or cancel the initialization of providers if `shutdown` is called while providers are still being initialized, and have yet to transition to `READY` or some other state.
```
lukas-reining: To me, this conflicts with the normative section. 
If shutdown is called during initialization, from the normative part only, I would expect the provider never to be shut down.

Especially, as an initializing provider is in `NOT_READY` state.
Or do I read that wrong?

lukas-reining: But I really struggle to come up with something that solves what I just said, but still does not result in providers that never shut down because during shut down they were initializing.

specification/sections/01-flag-evaluation.md
```diff
@@ -401,13 +401,22 @@ See [hooks](./04-hooks.md) for details.
...
+The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the active providers.
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
 
 see: [`shutdown`](./02-providers.md#25-shutdown)
 
+#### Requirement 1.6.2
+
+> The API's `shutdown` function **MUST NOT** call shutdown on providers which are in state `NOT_READY`.
+
+With respect to the lifecycle of a given active provider, the global API object's `shutdown` function should be idempotent; multiple calls to `shutdown` should result in only a single execution of the provider's `shutdown` function.
```
lukas-reining: When the shutdown function is called async or the program is concurrent in any way, this would also mean, that the provider may still be in `READY` state. 
Maybe we can add something like this: 


```suggestion
With respect to the lifecycle of a given active provider, the global API object's `shutdown` function should be idempotent; multiple calls to `shutdown` should result in only a single execution of the provider's `shutdown` function, even if the provider has not yet transitioned into `NOT_READY` state.
```

lukas-reining (commented):
I like the change. 
Left 2 questions :)

lukas-reining:
>> Should we also add a blurb about what exactly the SDK should do when the API's Shutdown function is invoked? It looks like some SDKs clear the entire internal state of providers (effectively unregistering all providers). Is that what is expected, or do we just want implementations to transition the provider status to NOT_READY? See https://github.com/open-feature/go-sdk/pull/400#discussion_r2187518510
>
> We could make this another point if everyone agrees, but I'm a bit skeptical of the value...

Mh, I think clearing the providers after the SDK has been shutdown could be a good thing to not have undefined behavior. @toddbaert 
For tests or special ways of using the SDK like in some serverless envs it could be good to know what to expect from the SDK.

toddbaert:
> > Shutdown is meant to be called on application exit
> 
> This aspect is not covered in the specification. In practice, some developers call the global API shutdown after each their test, expecting a clean state before next test run.

@erka Hmmm... should we detatch all event listeners and hooks as well? This starts to look to me more like a cleanup testing utility and less like a way to manage the lifecycle of the providers, which was the original intent at least on my end (`shutdown` was introduced along with the `initialization/shutdown` of providers).

specification/sections/01-flag-evaluation.md
toddbaert: I was imagining using locks or other such mechanisms to make sure an `init` completed (or failed) before `shutdown` was called.

toddbaert: I should add... I think this will add some implementation complexity, and we also don't have any recommendation that providers have a timeout (I think we could consider a `SHOULD` point for this, for providers which take time to start up).

specification/sections/01-flag-evaluation.md
```diff
@@ -401,13 +401,22 @@ See [hooks](./04-hooks.md) for details.
...
 Alternatively, implementations might leverage language idioms such as auto-disposable interfaces or some means of cancellation signal propagation to allow for graceful shutdown.
 
 see: [`shutdown`](./02-providers.md#25-shutdown)
 
+#### Requirement 1.6.2
+
+> The API's `shutdown` function **MUST NOT** call shutdown on providers which are in state `NOT_READY`.
+
+With respect to the lifecycle of a given active provider, the global API object's `shutdown` function should be idempotent; multiple calls to `shutdown` should result in only a single execution of the provider's `shutdown` function, even if the provider has not yet transitioned into `NOT_READY` state.
+Implementations should take care to await or cancel the initialization of providers if `shutdown` is called while providers are still being initialized, and have yet to transition to `READY` or some other state.
```
chrfwow: It seems to me as if implementing this without a race condition somewhere and without the possibility to call `shutdown` twice on a provider will be very tricky. I think it would be easier to require providers to be able to handle multiple calls to `shutdown`.

toddbaert: I have been worried about this as well. I was looking into some implementations, and I can see a few where it wouldn't be so bad, and other where it would be difficult... I know you have a knack for concurrency so I think I'll make a change if you also have this concern.

~I think maybe we could simplify it to say that we call shutdown on all providers in state NOT_READY, but don't make other guarantees (so it's possible to get re-entrant calls to shutdown, but once shutdown is complete there should be no more).~

Actually, I think we have another problem if we only call shutdown on `NOT_READY` providers but don't await all startu-ps... if a provider is starting up, but hasn't yet started, it's in `NOT_READY` (and so never receives a shutdown) but then after shutdown it might become `READY`, leaving us in a somewhat troublesome state.

I think if we can't guarantee that the `shutdown` won't be called re-entrantly, we have to run it for all providers.

specification/sections/01-flag-evaluation.md
toddbaert: @lukas-reining I've substantially changed the PR as mentioned in the description. Please re-review.

specification/sections/01-flag-evaluation.md
toddbaert: @chrfwow please see my edit to the description, and re-review.

specification/sections/01-flag-evaluation.md
```diff
@@ -399,15 +399,27 @@ See [hooks](./04-hooks.md) for details.
 
 [![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
 
+The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
+It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, and providers; this is useful for testing purposes.
```
beeme1mr: I believe we have separate APIs to do this in JS, but I think it makes sense for it to be included as part of the shutdown process.

specification/sections/01-flag-evaluation.md
bduffany: As someone who is not familiar with the openfeature spec, but has worked with the Go SDK a little bit, and has experience with concurrency as well as API design, this is how I wish things could behave:

- Global API `shutdown`: this function should call `shutdown` on all registered providers, regardless of their state. It should synchronously block until shutdown is complete. It should clean up all state associated with the global singleton. This includes any references to the provider objects, global hooks, event listeners, etc. It should be idempotent - so if no providers / hooks etc are registered, it just does nothing.
  - This is adding more responsibilities to `shutdown` of the global API singleton, but the benefit is that it provides a mechanism for cleaning up *all* global state associated with the global API object, which provides maximum flexibility for applications, and in particular, improves testability.
 
- Provider `shutdown`: this function should terminate the provider and clean up any resources associated with the provider. It should not care about the state that the provider is in. It should synchronously block until all resources are freed. (If callers prefer async shutdown, they are free to start their own threads etc. to do so, but providers themselves should shut down synchronously.) Provider shutdown should **not** have to be idempotent. Idempotency sometimes requires increased complexity (for example, in golang, channels cannot be closed twice). So requiring every provider to be idempotent shifts more burden and complexity onto providers, for no benefit (since the global API object is already keeping track of whether it has called shutdown or not).

Maybe the behavior I'm describing could be instead be called something like "destroy" or "cleanup" but I think it would add confusion to have two similarly named methods (both shutdown and cleanup).

Side note: I believe the current issues at least with the Go SDK mostly stem from the global singleton design. If the API object were not global, and users could instead create their own instances and manage the lifecycle themselves, this problem wouldn't exist. IMO it would be strictly better to allow multiple instances rather than having a singleton. Applications would be free to create a global singleton if they wanted, but that should be up to the application, and not enforced by the spec.

specification/sections/01-flag-evaluation.md
beeme1mr: Should we explicitly mention that the noop provider should be the default?

toddbaert: > I believe we have separate APIs to do this in JS, but I think it makes sense for it to be included as part of the shutdown process.

Ya I agree.

> Should we explicitly mention that the noop provider should be the default?

The noop provider is an implementation detail not mentioned in any normative part of the spec, AFAIK. We only mention that things should "no-op", but we can mention something like that and I think implementations will get the point.

specification/sections/01-flag-evaluation.md
toddbaert: @bduffany I _think_ what you've proposed is basically reflected in the PR after my recent update just now, with the exception of the recommendation (RFC 2119 "SHOULD")  that _provider shutdown_ be idempotent. I'm open to dropping this recommendation but I don't think it hurts, and again it is only a recommendation.

> Side note: I believe the current issues at least with the Go SDK mostly stem from the global singleton design. If the API object were not global, and users could instead create their own instances and manage the lifecycle themselves, this problem wouldn't exist.

Yes. This is probably true, and is true of a few parts of the API. The global singleton presents some challenges, but similar to the OTel global singleton, also is basically the only way to allow for some powerful 3rd party integrations; it was a design trade-off early in the project that we can't go back on now without a very substantial redesign that would compromise features.

specification/sections/01-flag-evaluation.md
toddbaert: We mention the no-op provider in [this non-normative section](https://openfeature.dev/specification/sections/flag-evaluation#overview) so I'm OK with mentioning it in a non-normative section here as well. Will add.

specification/sections/01-flag-evaluation.md
bduffany: > I'm open to dropping this recommendation but I don't think it hurts, and again it is only a recommendation.

I think it might hurt in the sense that it increases the burden on every provider (despite being a recommendation, providers probably will strive to implement all of the recommendations). Also it increases the overall size of the spec, without adding any benefit, since the global API singleton is already guaranteeing idempotence if I'm understanding correctly.

specification/sections/02-providers.md
```diff
@@ -225,11 +225,20 @@ class MyProvider implements Provider, AutoDisposable {
...
 If a provider requires initialization, once it's shut down, it must transition to its uninitialized state.
 Some providers may allow reinitialization from this state.
 Providers not requiring initialization are assumed to be ready at all times.
+Providers in the process of initializing abort initialization if shutdown is called while they are still starting up.
+
+see: [initialization](#24-initialization)
+
+#### Requirement 2.5.3
+
+> A Provider's `shutdown` function **SHOULD** be idempotent.
```
lukas-reining, line 239: I know that a MUST for idempotency would leave open much room for discussions about idempotency of implementations but I am not sure if it is still necessary.

Without it, the API could never assume it to be safe to call showdown multiple times.
But if it can not do this, the whole idempotency seems to not solve the problem sufficiently, because the SDK would have to track calls to shutdown again because from the status of a provider you can never see if it is READY or transitioning to NOT_READY 🤔

Does that make sense @toddbaert?

lukas-reining (commented):
This looks good to me, I am only still not sure about the provider idempotency problem.

specification/sections/02-providers.md
toddbaert: @lukas-reining  I'm not sure I would say it's "unsafe", I think it only means that the provider might be left in a weird state if the provider implementation doesn't handle multiple shutdowns, which is why "SHOULD" seems acceptable to me. We will, with this proposal as currently written, be removing the provider from the registry after shutting it down, so I think as far as the API is concerned, we are in a known state.

specification/sections/01-flag-evaluation.md
toddbaert: >  since the global API singleton is already guaranteeing idempotence if I'm understanding correctly

Well, since my changes earlier, we aren't guaranteeing this. The more implementations I looked at, the more I became convinced this would be hard :tm: to absolutely guarantee, as @chrfwow agreed [here](https://github.com/open-feature/spec/pull/323#discussion_r2199642376), without a commensurate benefit. I think as written, because the provider is removed after shutdown, it's extremely unlikely, but not guaranteed that a provider might never have it's shutdown called twice.

erka (approved):

beeme1mr (approved):

specification/sections/02-providers.md
lukas-reining: Okay, seems good to me as you describe it :)

specification/sections/01-flag-evaluation.md
```diff
@@ -399,15 +399,27 @@ See [hooks](./04-hooks.md) for details.
 
 [![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
 
+The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
+It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, providers, and setting a "No-op provider"; this is useful for testing purposes.
+It's recommended that application-authors to call this function on application shutdown, and after the completion of test suites which make use of the SDK.
```
sahidvelji:
```suggestion
It's recommended that application-authors call this function on application shutdown, and after the completion of test suites which make use of the SDK.
```

specification/sections/01-flag-evaluation.md
```diff
@@ -399,15 +399,27 @@ See [hooks](./04-hooks.md) for details.
...
+
 #### Requirement 1.6.1
 
-> The API **MUST** define a mechanism to propagate a shutdown request to active providers.
+> The API **MUST** define a function to propagate a shutdown request to all providers.
 
-The global API object might expose a `shutdown` function, which will call the respective `shutdown` function on the registered providers.
+The global API object defines a `shutdown` function, which will call the respective `shutdown` function on the active providers.
```
chrfwow: Only on the active providers or on all providers?

toddbaert: Good catch, this should match the normative section. Corrected.

toddbaert:
```suggestion
The global API object defines a `shutdown` function, which will call the respective `shutdown` function on all providers.
```

toddbaert: Fixed.

chrfwow (approved):

PR #360, opened 2026-01-15
fix!: Fix numbering for immutable flag metadata requirement

keelerm84:


beeme1mr (approved):

beeme1mr:
Thanks @keelerm84! I'm going to merge this because it's a clear typo and doesn't change the intent defined by the spec.

PR #367, opened 2026-02-25
fix: require SDK to update status before emitting events

dd-oleksii:
## This PR

Update the wording to require SDKs to update status before emitting events.

This is required, so that handlers see a consistent status. Otherwise, a READY handler may try evaluating flags only to discover that provider is still "officially" in NOT_READY or ERROR state and does not respond to evaluations.

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -1090,7 +1090,7 @@
         {
             "id": "Requirement 5.3.5",
             "machine_id": "requirement_5_3_5",
-            "content": "If the provider emits an event, the value of the client's `provider status` MUST be updated accordingly.",
+            "content": "If the provider emits an event, the value of the client's `provider status` MUST be updated to the status associated with that event **before** the SDK invokes any event handlers for that event, so that handlers observe a consistent status.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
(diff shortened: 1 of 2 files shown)


lukas-reining (approved):
That is a good and necessary concretization!
Thanks @dd-oleksii!

beeme1mr (approved):

gruebel (approved):

gruebel:
Do we handle this as a bugfix in the SDKs or a breaking change, because it kind of changes the behavior, but in a good way 😅

erka (approved):

lukas-reining:
> Do we handle this as a bugfix in the SDKs or a breaking change, because it kind of changes the behavior, but in a good way 😅

I get why we would mark this as breaking change. To me though, we can mark this as a bug fix @gruebel. 
The new behavior is the "logical" one and would make sense to assume to me. 
The current allowed race condition seems like a real bug even though it is allowed by the spec rn.

askpt (approved):

toddbaert (approved):
Totally agree.

I will audit SDKs tomorrow, unless you have already @dd-oleksii ...

toddbaert:
> Do we handle this as a bugfix in the SDKs or a breaking change, because it kind of changes the behavior, but in a good way 😅

In general I would consider such changes non breaking - we are dealing with nuances of ordering. We can call it out in the changelog, but it's not something I would expect to create issues for most users.

toddbaert:
Seems to me that only go is in violation of this at the moment: https://github.com/open-feature/go-sdk/issues/493. I think Kotlin already has an open PR.

PR #368, opened 2026-02-26
feat: add isolated instance

### LINKED ISSUE #359: Considering alternatives to the OpenFeature global singleton
The OpenFeature API features a global singleton. Though this has advantages (ease-of-use, available everywhere without pollute the call stack, facilitation of transparent 3rd party integrations) it also causes issues. A couple examples:

- can break tests, particularly in languages like Java, where tests mutating the singleton can trample on each-other
- micro-frontends, which combine multiple codebases which each might maniplate the singleton
- server-side programs comprised of multiple submodules, each setting their own provider
- DI frameworks and IoC containers, which inherently don't work well with singletons managed outside their scope

We'd like to explore means of "opting out" or working around the singleton.

# OpenFeature Singleton Brainstorming Meeting

Find below resources from our Dec 18th meeting on this topic (LFX account required for some assets):

- Transcript + Recording: https://zoom.us/rec/share/OQUwHFumc1DXRumVxIpD1BKq_epVS9ajj00jJ-7Vt4oLQCLMrW3P7waEvVb_Nw7I.CxKvHOr4meTgZjEd
- AI Meeting Summary: https://zoom-lfx.platform.linuxfoundation.org/meeting/96839205440-1766073600000/summaries?password=db7e61a2-5aa4-4e19-ad42-c34799ce1397

## Questions/comments

- In javascript, why did we bind the singleton to window \- truely global \- vs. binding to the module  
  - Todd says because we intentionally wanted to make the API object a true globalton  
- [Next.js](http://Next.js) \- issues here because server-side rendered code is half-dynamic-context, half-static-context paradigm  
- If you can have multiple API objects floating around, what's the value in distinguishing between client and API  
  - We could still keep separate interfaces that are backed by the same object instances, even if we did this

### Option A

Expose ability to create non-singleton instances of the API object

- Benefits  
  - Simple solution  
  - Potential to allow us to remove "domains" concept  
- Drawbacks  
  - Could be some gotchas  
    - Bit of a fear of unknown-unknowns.  
      - Some deep research on where we assume a globalton would be needed  
    - ? bugs around shutdown lifecycle   
  - Confusing footgun \- new user might stumble into newing up when they should use the singleton  
    - Could we mitigate this by not exposing the ctor but instead exposing OpenFeature.advanced\_use\_only\_new\_instance() or whatever  
      - Matt proposes OpenFeature.isolated()  
      - Jonathan proposes hiding it in a different class  
  - We would need to ensure that a specific provider instance cannot be bound to multiple API instances  
    - Event lifecycle etc.  
  - If we replaced domains with this then it could make end-user code more complex/boilerplate-y

### Option B

Add methods to the singleton that would give you an isolated scope \- sort of like "domains" in the client?

- Drawbacks  
  - Adds a lot of surface area to the api

### Option C

*JS-specific solution*: bind the singleton at module scope, rather than on window

### Option D

Extended API so that you can specify a provider when you create a client   
This is what .NET impl does already for DI

- Benefits  
  - Seems to be working well in .NET so far (dynamic context only)  
- Drawbacks  
  - Need to expose more on API surface area in order to support context management in "user land"  
    - On client, and possibly also on provider (:scream:)  
  - Opening up the option means we have to support it everywhere, which means the value of the API class sort of evaporates

# Actions

### PoC some of the options and present back findings

- @MattIPv4  offers to PoC Option A  
- @jonathannorris  offers to PoC Option A where we have a separate class or factory function or whatever, to make the footgun less discoverable  
- @lukas-reining  wants to PoC ensuring that a provider instance can only be bound to a single API instance (if we went down the Option A path)  
- @cupofcat offers to prototype options in Go, starting mid Jan

toddbaert:
Adds spec for isolated instances.

Resolves: https://github.com/open-feature/spec/issues/359

Relevant work: 
- https://github.com/open-feature/js-sdk/pull/1325
- https://github.com/open-feature/js-sdk/pull/1342

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -574,34 +574,10 @@ import { createIsolatedOpenFeatureAPI } from '@openfeature/web-sdk/isolated';
 
 #### Requirement 1.8.4
 
-> A `provider` instance **MUST NOT** be registered with more than one `API` instance simultaneously.
+> A `provider` instance **SHOULD NOT** be registered with more than one `API` instance simultaneously.
 
-Because the `API` instance manages the [lifecycle](./02-providers.md) of its associated providers (including initialization, shutdown, and event handling), binding a `provider` to more than one `API` instance could result in undefined behavior. A `provider` instance **MAY** be registered with multiple `domains` within a single `API` instance.
+Because the `API` instance manages the [lifecycle](./02-providers.md) of its associated providers (including initialization, shutdown, and event handling), binding a `provider` to more than one `API` instance could result in undefined behavior.
+A `provider` instance can be registered with multiple `domains` within a single `API` instance.
+When a provider is no longer associated with an `API` instance, it can be registered to another. 
 
 See [setting a provider](#setting-a-provider), [domain](../glossary.md#domain) for details.
-
-#### Requirement 1.8.5
-
-> The `provider mutator` **MUST** indicate an error if the specified `provider` instance is already bound to a different `API` instance.
-
-Implementations indicate an error in a manner idiomatic to the language in use (returning an error, throwing an exception, etc).
-
-```java
-// example: registering a provider already bound to another instance results in an error
-OpenFeatureAPI isolatedA = createIsolatedOpenFeatureAPI();
-OpenFeatureAPI isolatedB = createIsolatedOpenFeatureAPI();
-
-Provider myProvider = new MyProvider();
-isolatedA.setProvider(myProvider);
-isolatedB.setProvider(myProvider); // error: provider is already bound to a different API instance
-```
-
-#### Requirement 1.8.6
-
-> A `provider` **MUST** be unbound from an `API` instance when it is no longer registered with that instance.
-
-A `provider` is no longer registered when it has been replaced (by setting a new `provider` for the same `domain`), when `clearProviders` has been called, or when the `API`'s `shutdown` function has been called.
-Providers which remain registered under other `domains` within the same `API` instance are not unbound.
-Once unbound, a `provider` instance may be registered with any `API` instance.
-
-See [shutdown](#16-shutdown), [setting a provider](#setting-a-provider) for details.
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -530,3 +530,78 @@ see: [error codes](../types.md#error-code)
 > The client's `provider status` accessor **MUST** indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.
 
 Regardless of the success of the provider's `shutdown` function, the `provider status` should convey the provider is no longer ready to use once the shutdown function terminates.
+
+### 1.8. Isolated API Instances
+
+[![experimental](https://img.shields.io/static/v1?label=Status&message=experimental&color=orange)](https://github.com/open-feature/spec/tree/main/specification#experimental)
+
+While the `API` [functions as a global singleton](#requirement-111) in the default case, certain use cases require independent instances of the `API` with fully isolated state. Examples include micro-frontend architectures (where separately developed applications coexist in a single runtime), dependency injection frameworks and IoC containers (which manage object lifecycles outside the singleton's scope), testing scenarios (where parallel tests mutating shared state can interfere with each other), and server-side applications composed of multiple submodules each requiring distinct providers.
+
+#### Requirement 1.8.1
+
+> The `API` **MUST** expose a factory function which creates and returns a new, independent instance of the `API`.
+
+Each instance returned by this factory function maintains its own state, including providers, `evaluation context`, hooks, event handlers, and transaction context propagators.
+Instances created by the factory function do not share state with the "default" global singleton or with each other.
+
+```java
+// example factory function
+OpenFeatureAPI isolated = createIsolatedOpenFeatureAPI();
+isolated.setProvider(new MyProvider());
+Client client = isolated.getClient();
+```
+
+See [application integrator](../glossary.md#application-integrator), [isolated API instance](../glossary.md#isolated-api-instance) for details.
+
+#### Requirement 1.8.2
+
+> Instances returned by the factory function **MUST** conform to the same `API` contract as the global singleton, including flag evaluation, provider management, context, hooks, events, and shutdown functionality.
+
+An isolated `API` instance is functionally equivalent to the global singleton, but with independent state.
+Application code which uses a `client` obtained from an isolated instance behaves identically to code using a `client` from the global singleton.
+
+#### Requirement 1.8.3
+
+> The factory function for creating isolated instances **SHOULD** be housed in a distinct module, import path, package, or namespace from the global singleton `API`.
+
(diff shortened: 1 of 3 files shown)


MattIPv4 (approved):

specification/sections/01-flag-evaluation.md
```diff
@@ -530,3 +530,78 @@ see: [error codes](../types.md#error-code)
...
+#### Requirement 1.8.2
+
+> Instances returned by the factory function **MUST** conform to the same `API` contract as the global singleton, including flag evaluation, provider management, context, hooks, events, and shutdown functionality.
+
+An isolated `API` instance is functionally equivalent to the global singleton, but with independent state.
+Application code which uses a `client` obtained from an isolated instance behaves identically to code using a `client` from the global singleton.
+
+#### Requirement 1.8.3
+
+> The factory function for creating isolated instances **SHOULD** be housed in a distinct module, import path, package, or namespace from the global singleton `API`.
```
toddbaert, line 565: I went with **SHOULD** here, as opposed to **MUST** because I can imagine there may be some implementations where this is difficult or impractical, so it's just a strong recommendation.

specification/sections/01-flag-evaluation.md
```diff
@@ -530,3 +530,78 @@ see: [error codes](../types.md#error-code)
...
+The distinct import path serves as an explicit signal that the consumer is opting into advanced, non-default behavior.
+
+```typescript
+// example: the factory function is accessed from a distinct module
+import { createIsolatedOpenFeatureAPI } from '@openfeature/web-sdk/isolated';
+```
+
+#### Requirement 1.8.4
+
+> A `provider` instance **MUST NOT** be registered with more than one `API` instance simultaneously.
```
chrfwow: Since different instances of the API must not share state, we cannot enforce this requirement, except if we make a (possibly breaking) change to the providers

specification/sections/01-flag-evaluation.md
lukas-reining: This makes sense!

specification/sections/01-flag-evaluation.md
```diff
@@ -530,3 +530,78 @@ see: [error codes](../types.md#error-code)
...
+
+```java
+// example: registering a provider already bound to another instance results in an error
+OpenFeatureAPI isolatedA = createIsolatedOpenFeatureAPI();
+OpenFeatureAPI isolatedB = createIsolatedOpenFeatureAPI();
+
+Provider myProvider = new MyProvider();
+isolatedA.setProvider(myProvider);
+isolatedB.setProvider(myProvider); // error: provider is already bound to a different API instance
+```
```
lukas-reining: My draft for this for JS can be found here: https://github.com/open-feature/js-sdk/pull/1342

lukas-reining (approved):
This looks good to me!

specification/sections/01-flag-evaluation.md
lukas-reining: This is my draft for JS @chrfwow that does not need a change to the provider: https://github.com/open-feature/js-sdk/pull/1342

lukas-reining: I would still like to check how to do it for JS and Go.

erka: `Object.freeze(provider)` or `Object.seal(provider)`?

MattIPv4: Both of those would prevent a provider from being assigned to multiple domains within the same API instance. I think Lukas' solution is the way to go, having the API instance attach some hidden property to a provider to track which API instance it is being used with.

erka: If I freeze the provider before `OpenFeature.setProvider`.  What will happen?

lukas-reining: Do you mean today? Freezing a provider does nothing if I am not overseeing anything. 
The client and api do not make any mutations on the provider. 
So I also think that `Object.freeze/seal` would not be suitable for locking the provider to an API instance. 

My proposed solution for JS would simply break :D 
But why would an app author seal a provider object they did not write and never use directly?
I would expect problems after that.

erka: There could be many reasons. Tests, in-memory or environment providers, etc...

lukas-reining: I understand this but I am not convinced that we should assume this to be valid. 
E.g. I do not see why you would need to freeze the `in-memory or environment provider` instances.

But let's discuss this in the https://github.com/open-feature/js-sdk/pull/1342.

kinyoklion: It does feel like this requirement, without a provider change, may result in somewhat unexpected solutions. Which may be difficult for a consumer to reason about, add unexpected state, or be unexpected for contributors.

Additional global maps tracking instances, things like swift associated objects, defining additional properties on providers ( I had to double check, as a provider author, I hadn't froze my server-side js provider.). 

Not to interpret that I think we shouldn't have this requirement, it is a must. But I do wonder, in some languages, if just telling people not to do it is sufficient.

erka: My main point is that the spec should focus on what all SDKs can reasonably implement, not on tricks behind the scenes. Rust... it will not allow this without a mutable reference. 

Provider status with replacement has similar complexity, and we all know where things stand with it.

toddbaert: We can reduce this to a `SHOULD` and implement it only where reasonable, and otherwise just document it.

This entire feature is more important in some implementations (looking at you Javascript) than others.

askpt: I believe in some implementations/languages, you can have an `internal` method at provider level that "locks" the provider from being bonded to another `API` during the registration. This might be a breaking in some languages (for dotnet I believe it is not, because it will load the new Abstract Provider from the SDK).

By saying this, from `dotnet` standpoint, I don't see an issue on having this requirement.

lukas-reining: > Not to interpret that I think we shouldn't have this requirement, it is a must. But I do wonder, in some languages, if just telling people not to do it is sufficient.

I think the point `for some languages` is the important one. 

> We can reduce this to a SHOULD and implement it only where reasonable, and otherwise just document it.

`SHOULD` is a good thing for this.
To me that sounds sensible.

toddbaert: I will reduce this to a `SHOULD NOT`. It's just a recommendation, and if maintainers decide it can't be done in a reasonable way, I'm OK side-stepping it.

I won't dwell on the particulars of the JS solution here.

toddbaert: @erka @MattIPv4 @lukas-reining @chrfwow @askpt 

I relaxed 1.8.4, and removed 1.8.5 and 1.8.6: https://github.com/open-feature/spec/pull/368/changes/d48f002524c4ae282a5f0dbba2cd2f79b7990e04

MattIPv4: LGTM

erka (approved):

askpt (approved):

kinyoklion (approved):

PR #375, opened 2026-04-08
fix: add evaluation context and transaction context propagators to shutdown reset requirement (1.6.2)

toddbaert:
## Summary

Requirement 1.6.2 says shutdown MUST reset all state, removing "hooks, event handlers, and providers" but doesn't mention evaluation context or transaction context propagators. The non-normative preamble already says "resetting the API object to its default state" which implies these, but the normative text doesn't list them.

This adds "evaluation context" and "transaction context propagators" to the list in both the normative requirement and the non-normative preamble.

Some SDKs already reset these on shutdown (Go, .NET), others don't (Java, JS). This makes the expectation clear.

### DIFF
--- a/specification/sections/01-flag-evaluation.md
+++ b/specification/sections/01-flag-evaluation.md
@@ -399,7 +399,7 @@ See [hooks](./04-hooks.md) for details.
 [![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
 
 The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
-It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, providers, and setting a "No-op provider"; this is useful for testing purposes.
+It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, evaluation context, transaction context propagators, providers, and setting a "No-op provider"; this is useful for testing purposes.
 It's recommended that application-authors call this function on application shutdown, and after the completion of test suites which make use of the SDK.
 
 #### Requirement 1.6.1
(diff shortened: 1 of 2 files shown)


specification/sections/01-flag-evaluation.md
```diff
@@ -399,7 +399,7 @@ See [hooks](./04-hooks.md) for details.
 [![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
 
 The API's `shutdown` function defines a means of graceful shutdown, calling the `shutdown` function on all providers, allowing them to flush telemetry, clean up connections, and release any relevant resources.
-It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, providers, and setting a "No-op provider"; this is useful for testing purposes.
+It also provides a means of resetting the API object to its default state, removing all hooks, event handlers, evaluation context, providers, and setting a "No-op provider"; this is useful for testing purposes.
```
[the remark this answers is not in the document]
toddbaert: Good point!

specification/sections/01-flag-evaluation.md
```diff
@@ -414,7 +414,7 @@ see: [`shutdown`](./02-providers.md#25-shutdown)
 
 #### Requirement 1.6.2
 
-> The API's `shutdown` function **MUST** reset all state of the API, removing all hooks, event handlers, and providers.
+> The API's `shutdown` function **MUST** reset all state of the API, removing all hooks, event handlers, evaluation context, transaction context propagators, and providers.
```
beeme1mr, line 417: How would you shut down or reset the transaction context propagator?

toddbaert: Just remove any attached translation propagators, similarly to removing attached hooks.

I think they are basically equivalent in terms of how they are state that has side effects, so IMO makes sense to clear them as well.

beeme1mr: Update: never mind, I understand. We would clear the registered transaction context propagator, similar to unregistering hooks.

beeme1mr (approved):

askpt (approved):
LGTM! I wonder if the SDKs are already doing this by default.

kinyoklion (approved):

PR #380, opened 2026-04-23
fix: provider state race

### LINKED ISSUE #365: Provider lifecycle race condition in multi-threaded SDKs
I've been digging provider lifecycle in Go and Swift SDKs, and I believe there's a couple of things in OF spec that make it hard to implement it correctly in the presence of true multi-threading.

The main one is that there's no clear ownership of the provider state and provider events — it's split between the provider and the API, and both provider and the API are allowed to emit events. This is an issue because neither the provider nor the API are equipped to ensure the consistent ordering of events

The API is required to emit "provider ready" event and update the state after initialization is complete. However, once the provider returns from the init function, there's a brief window before the API updates the state and emits the event. This is not an issue in single-threaded languages like js or python. However, with multi-threading, the provider is allowed to do work in the background and emit an error event within that window — and now the order of provider ready and error events is undefined.

Current state across multi-threaded SDKs:
- Go: race condition
- Swift: 🐛 implements eventing but does not update state on provider events
- Java: race condition
- Kotlin: race condition
- Dotnet: race condition
- Rust: eventing not implemented

This would be trivial to fix if only one of API or provider were allowed to emit events (or update the state), so they could use locks/queues to order the events/updates properly.

If both are allowed to emit, the only way to prevent a race is for the API to guard eventing with a lock and to hold this lock throughout the provider initialization. This way, no provider events would come through until API processes its own ready event first.

The main issue with this approach is that it locks up the API/provider and serializes all calls to init/shutdown/on context change. This more or less completely destroys concurrency and prevents the provider from reacting to context change / shutdown if initialization is taking a while. This also easily leads to deadlocks if provider tries to emit events from/during lifecycle handlers.

Both race condition and total locking sound bad, so I would really suggest moving state management responsibility back to providers — they are the only party who knows their own state at all times, and they are equipped to resolve any races.

I see that the spec is trying to lift the burden from provider authors but imo SDK trying to manage providers' state is not helping. Anecdotally, I can add that as a provider author, the eventing and state management has been the most confusing part of the OF spec/SDKs, and I've spent quite a bit of time trying to make SDK set the state I want.

--- 

PoCs by @toddbaert here:

Go: https://github.com/open-feature/go-sdk/pull/487
Java: https://github.com/open-feature/java-sdk/pull/1892
JS: https://github.com/open-feature/js-sdk/pull/1362

toddbaert:
* moves "provider status" back into the provider interface
* clarifies some ordering and atomicity
* adds "migration" appendix describing the proposed migration strategy

For background, see: https://github.com/open-feature/spec/issues/365

--- 

PoCs for the migration strategy (not the actual spec changes) by @toddbaert here:

Go: https://github.com/open-feature/go-sdk/pull/487
Java: https://github.com/open-feature/java-sdk/pull/1892
JS: https://github.com/open-feature/js-sdk/pull/1362

---

Fixes: https://github.com/open-feature/spec/issues/365

If/when this is merged, I will create issues for every implementation (thought I think Kotlin/Swift might already be corrected).

:warning: IMO this is superseded by https://github.com/open-feature/spec/pull/385

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -310,7 +310,7 @@
                 {
                     "id": "Conditional Requirement 1.7.2.1",
                     "machine_id": "conditional_requirement_1_7_2_1",
-                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `RECONCILING`.",
+                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the `provider status` accessor must support possible value `RECONCILING`.",
                     "RFC 2119 keyword": null,
                     "children": []
                 }
@@ -319,21 +319,21 @@
         {
             "id": "Requirement 1.7.3",
             "machine_id": "requirement_1_7_3",
-            "content": "The client's `provider status` accessor MUST indicate `READY` if the `initialize` function of the associated provider terminates normally.",
+            "content": "The `provider status` MUST indicate `READY` if the `initialize` function of the associated provider terminates normally.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.7.4",
             "machine_id": "requirement_1_7_4",
-            "content": "The client's `provider status` accessor MUST indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.",
+            "content": "The `provider status` MUST indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.7.5",
             "machine_id": "requirement_1_7_5",
-            "content": "The client's `provider status` accessor MUST indicate `FATAL` if the `initialize` function of the associated provider terminates abnormally and indicates `error code` `PROVIDER_FATAL`.",
+            "content": "The `provider status` MUST indicate `FATAL` if the `initialize` function of the associated provider terminates abnormally and indicates `error code` `PROVIDER_FATAL`.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -361,7 +361,7 @@
         {
             "id": "Requirement 1.7.9",
             "machine_id": "requirement_1_7_9",
-            "content": "The client's `provider status` accessor MUST indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.",
+            "content": "The `provider status` MUST indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -562,6 +562,49 @@
             "machine_id": "condition_2_7_1",
             "content": "The `provider` MAY define a function for tracking the occurrence of a particular user action or application state, with parameters `tracking event name` (string, required), `evaluation context` (optional) and `tracking event details` (optional) which returns nothing.",
             "RFC 2119 keyword": "MAY",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.8.1",
+            "machine_id": "requirement_2_8_1",
+            "content": "The provider MUST define a `status` accessor which indicates the provider's current readiness, with possible values `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Condition 2.8.2",
+            "machine_id": "condition_2_8_2",
+            "content": "The implementation uses the static-context paradigm.",
+            "RFC 2119 keyword": null,
+            "children": [
+                {
+                    "id": "Conditional Requirement 2.8.2.1",
+                    "machine_id": "conditional_requirement_2_8_2_1",
+                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`, the provider's `status` accessor MUST support possible value `RECONCILING`.",
+                    "RFC 2119 keyword": "MUST",
+                    "children": []
+                }
+            ]
+        },
+        {
+            "id": "Requirement 2.8.3",
+            "machine_id": "requirement_2_8_3",
+            "content": "The provider's `status` MUST be `NOT_READY` before `initialize` is called and after `shutdown` terminates.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.8.4",
+            "machine_id": "requirement_2_8_4",
+            "content": "The provider's `status` accessor MUST be safe for concurrent access.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.8.5",
+            "machine_id": "requirement_2_8_5",
+            "content": "Status changes and any associated event emissions MUST be atomic from the perspective of external observers.",
+            "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
@@ -984,7 +1027,7 @@
         {
             "id": "Requirement 5.1.1",
             "machine_id": "requirement_5_1_1",
-            "content": "The `provider` MAY define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with a `provider event details` payload.",
+            "content": "The `provider` MAY define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED`, `PROVIDER_STALE`, `PROVIDER_RECONCILING`, and `PROVIDER_CONTEXT_CHANGED`, with a `provider event details` payload.",
             "RFC 2119 keyword": "MAY",
             "children": []
         },
@@ -1118,7 +1161,7 @@
         {
             "id": "Requirement 5.3.5",
             "machine_id": "requirement_5_3_5",
-            "content": "If the provider emits an event, the value of the client's `provider status` MUST be updated to the status associated with that event **before** the SDK invokes any event handlers for that event, so that handlers observe a consistent status.",
+            "content": "When a provider emits an event, the value of the `provider status` MUST reflect the status associated with that event **before** any event handlers for that event are invoked, so that handlers observe a consistent status.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -310,8 +310,8 @@
                 {
                     "id": "Conditional Requirement 1.7.2.1",
                     "machine_id": "conditional_requirement_1_7_2_1",
-                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the `provider status` accessor must support possible value `RECONCILING`.",
-                    "RFC 2119 keyword": null,
+                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`, the `provider status` accessor MUST support possible value `RECONCILING`.",
+                    "RFC 2119 keyword": "MUST",
                     "children": []
                 }
             ]
--- a/specification/appendix-e-migrations.md
+++ b/specification/appendix-e-migrations.md
@@ -1,34 +1,60 @@
 ---
 id: appendix-e
 title: "Appendix E: Migrations"
-description: SDK migration guidance for breaking spec changes
+description: Migration guidance for breaking spec changes
 sidebar_position: 6
 ---
 
 # Appendix E: Migrations
 
-This appendix provides non-normative guidance for SDK authors on migrating to new or changed specification requirements without breaking existing providers or application consumers.
+This appendix provides non-normative guidance for provider authors and SDK authors on migrating to new or changed specification requirements.
 
 ## Provider Status Ownership
 
 ### Background
 
-Prior to this change, provider status (e.g. `NOT_READY`, `READY`, `ERROR`) was managed by the SDK on behalf of the provider. The SDK would set status and emit events after lifecycle methods (`initialize`, `shutdown`, `on context change`) returned. This created a race condition in multi-threaded SDKs: the provider could change its own state (e.g. emit an error event from a background thread) in the window between the lifecycle method returning and the SDK writing its post-lifecycle status and emitting the corresponding event. The result was incorrect event ordering and inconsistent status.
+Prior to `v0.9.0`, provider status (e.g. `NOT_READY`, `READY`, `ERROR`) was managed by the SDK on behalf of the provider.
+The SDK would set status and emit events after lifecycle methods (`initialize`, `shutdown`, `on context change`) returned.
+This created a race condition in multi-threaded SDKs: the provider could change its own state (e.g. emit an error event from a background thread) in the window between the lifecycle method returning and the SDK writing its post-lifecycle status and emitting the corresponding event.
+The result was incorrect event ordering and inconsistent status.
 
 The spec now requires providers to own their status and emit events atomically with status transitions (see [provider status](./sections/02-providers.md#28-provider-status)).
 
-### Migration: the `StateManagingProvider` interface
+### For provider authors
 
-To avoid breaking existing providers, SDKs should introduce an opt-in interface (or equivalent mechanism) that new providers implement to signal they manage their own status. Existing providers that do not implement this interface continue to have their status managed by the SDK, preserving backwards compatibility. This legacy behavior is deprecated and should be removed in the next major version.
+Providers are now responsible for maintaining their own `status` and emitting events atomically with status transitions.
 
-The interface should expose:
+#### What to implement
+
+- A `status` accessor returning the provider's current readiness: `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL` (plus `RECONCILING` in the static-context paradigm)
+- `status` must be `NOT_READY` before `initialize` is called and after `shutdown` terminates
+- `status` must be safe for concurrent access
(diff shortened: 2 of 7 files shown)


specification.json
```diff
@@ -310,7 +310,7 @@
                 {
                     "id": "Conditional Requirement 1.7.2.1",
                     "machine_id": "conditional_requirement_1_7_2_1",
-                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `RECONCILING`.",
+                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the `provider status` accessor must support possible value `RECONCILING`.",
```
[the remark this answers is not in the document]
toddbaert: Fixed, and fixed parser to catch the missing keyword.

specification/sections/01-flag-evaluation.md
```diff
@@ -477,27 +477,30 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
 
 ##### Conditional Requirement 1.7.2.1
 
-> In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `RECONCILING`.
+> In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the `provider status` accessor must support possible value `RECONCILING`.
```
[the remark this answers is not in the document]
toddbaert: https://github.com/open-feature/spec/pull/380#discussion_r3130740810

specification/sections/01-flag-evaluation.md
```diff
@@ -463,9 +463,9 @@ stateDiagram-v2
 
 > The `client` **MUST** define a `provider status` accessor which indicates the readiness of the associated provider, with possible values `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`.
 
-The SDK at all times maintains an up-to-date state corresponding to the success/failure of the last lifecycle method (`initialize`, `shutdown`, `on context change`) or emitted event.
+The client's `provider status` accessor delegates to the associated provider's `status` accessor, which the provider keeps in sync with the success/failure of the last lifecycle method (`initialize`, `shutdown`, `on context change`) or emitted event.
```
dd-oleksii: minor: given that providers must emit events on lifecycle methods, this paragraph can be simplified to:
```suggestion
The client's `provider status` accessor delegates to the associated provider's `status` accessor, which the provider keeps in sync with the last emitted event.
```

specification/sections/02-providers.md
```diff
@@ -305,3 +305,60 @@ The track function performs side effects required to record the `tracking event`
...
+> The provider's `status` accessor **MUST** be safe for concurrent access.
+
+In languages supporting multi-threaded execution, the provider must ensure that concurrent reads of the `status` accessor do not observe torn or inconsistent values.
+
+#### Requirement 2.8.5
+
+> Status changes and any associated event emissions **MUST** be atomic from the perspective of external observers.
+
+When a provider transitions between statuses and emits an event associated with that transition, external observers (such as SDK event handlers) must observe a consistent view: the updated `status` value and the emitted event are visible together.
+This prevents ordering anomalies where, for example, a `PROVIDER_READY` handler runs while `status` still indicates `NOT_READY` or `ERROR`, or where the provider transitions out of a status before the associated event is dispatched.
```
dd-oleksii, line 362: "Dispatch" may be a big ambiguous here:
> or where the provider transitions out of a status before the associated event is dispatched

Is the intent here that event handlers see the exact status that triggered the event? If so, this may be problematic as it requires holding the status lock while all handlers run, preventing the provider from changing its own status.

I think what we're after here is establishing an observable "happens-before" relationship:
- Status change must happen before the corresponding event handlers run.
- Event handlers must only be run after all event handlers for the previous event has finished.

So an event handler _may_ observe next status changes, the important bit is that it _must_ observe the status change that triggered the event.

specification/sections/05-events.md
```diff
@@ -205,45 +204,45 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
 
-The implementation must run any `PROVIDER_CONTEXT_CHANGED` handlers associated with the provider after the provider has reconciled its state and returned from the `on context changed` function.
-The `PROVIDER_CONTEXT_CHANGED` is not emitted from the provider itself; the SDK implementation must run the `PROVIDER_CONTEXT_CHANGED` handlers if the `on context changed` function terminates normally.
-It's possible that the `on context changed` function is invoked simultaneously or in quick succession; in this case the SDK will only run the `PROVIDER_CONTEXT_CHANGED` handlers after all reentrant invocations have terminated, and the last to terminate was successful (terminated normally).
+`PROVIDER_CONTEXT_CHANGED` handlers associated with the provider must run after the provider has reconciled its state and returned from the `on context changed` function.
+It's possible that the `on context changed` function is invoked simultaneously or in quick succession; in this case `PROVIDER_CONTEXT_CHANGED` handlers only run after all reentrant invocations have terminated, and the last to terminate was successful (terminated normally).
```
dd-oleksii: Nitpicking on words: the current phrasing says event handler must run after `on context changed` function _returns_, which doesn't make sense if event is emitted from that function.

The handling for reetrancy also seems overly-prescriptive and infeasible (the provider generally cannot emit events after all invocations terminated).

I think we can simplify this to:
- Provider must emit `PROVIDER_CONTEXT_CHANGED` after it successfully reconciled the context.
- It's possible that the `on context changed` function is invoked simultaneously or in quick succession, so the provider must be prepared to handle that.

specification/appendix-e-migrations.md
```diff
@@ -0,0 +1,104 @@
...
+- Updating status when the provider emits events at runtime (the provider already set its own status atomically with the event)
+- (Static-context paradigm only) Setting `RECONCILING` status, emitting `PROVIDER_RECONCILING`, setting `READY`/`ERROR` status, or emitting `PROVIDER_CONTEXT_CHANGED`/`PROVIDER_ERROR` during `on context change` handling
+
+#### What the SDK still does for all providers
+
+Regardless of whether the provider manages its own state, the SDK continues to:
+
+- Call `initialize()`, `shutdown()`, and `on context change` lifecycle methods on the provider
+- Forward provider-emitted events to registered domain and API-level event handlers
+- Run late-attached handlers immediately if the provider is already in the associated state
```
dd-oleksii, line 98: major: I believe SDKs can't do this safely as this requires that subscription and emission of the current status is atomic (there should be no new events emitted between reading current status and running the handler) — and you can't guarantee that when another thread is emitting events  (unless SDK implements event buffering which is cumbersome).

specification/appendix-e-migrations.md
```diff
@@ -0,0 +1,104 @@
...
+- Enforce short-circuit behavior for `NOT_READY` and `FATAL` statuses during flag evaluation
```
dd-oleksii, line 99: major: multi-threaded SDKs can't enforce this because of TOCTOU

specification/appendix-e-migrations.md
```diff
@@ -0,0 +1,104 @@
...
+### For SDK authors
+
+SDKs must detect whether a registered provider manages its own state and branch behavior accordingly.
+
+#### Detecting state-managing providers
+
+At registration time, check whether the provider implements the `StateManagingProvider` interface (or equivalent).
+Store this as a flag on the internal provider wrapper for use during lifecycle calls and event handling.
+
+#### SDK wrapper behavior
```
dd-oleksii, line 54: minor/aside: I'd argue that the better design is implementing an adapter that takes a non-`StateManagingProvider` and implements the `StateManagingProvider` interface. This way, all special-casing/mapping is concentrated in one place, and it is about the only thing that needs to be deleted in the next major release.

specification/sections/01-flag-evaluation.md
```diff
@@ -477,27 +477,30 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
 Once the provider has initialized, the `provider status` should indicate the provider is ready to be used to evaluate flags.
+The provider is responsible for its own status transition; the client's `provider status` accessor reflects it.
 
 #### Requirement 1.7.4
 
-> The client's `provider status` accessor **MUST** indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.
+> The `provider status` **MUST** indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.
 
 If the provider has failed to initialize, the `provider status` should indicate the provider is in an error state.
+The provider is responsible for its own status transition; the client's `provider status` accessor reflects it.
```
nicklasl: not sure this addition gives much added value?

nicklasl: And that goes for the other places where this sentence sits.

specification/appendix-e-migrations.md
toddbaert: IIUC this is essentially what the PoCs already do; FeatureProviderStateManager in Java and providerReference in Go are effectively that adapter/wrapper. They wrap legacy providers and handle state management on their behalf, so downstream SDK code doesn't need to special-case. At the next major version, the wrapper is the only thing that gets deleted.

Or am I missing something?

dd-oleksii: I was referring to using adapter pattern to adapt legacy interface to the new one. Strictly speaking, none of the PoCs do that.

[Kotlin PR](https://github.com/open-feature/kotlin-sdk/pull/241/changes#diff-918389ee8af3dea131804ade5c9b044ffc8bd2b01bd8ac806098c9045a97b907) is an example.:`LegacyFeatureProviderAdapter` wraps a non-state-managing provider and implement state-managing interface (manage status, emits events, etc.). It nicely contains/abstracts the knowledge of non-managing providers, and the rest of openfeature sdk handles all providers equally as if they can manage their state.

This also reduces the cyclomatic complexity overall: the adapter knows the wrapped provider doesn't manage status, so it doesn't need any ifs. The SDK can treat all providers as state-managing, so it doesn't need any ifs either. The only `if` is in `setProvider` to determine whether wrapping is needed.

In other PoCs, the provider wrapper now has a double-duty of whatever-it-was-doing-before + adapting legacy providers, so it's riddled with `if`s. And it still doesn't contain the adapting completely and this part doesn't hold:

> At the next major version, the wrapper is the only thing that gets deleted.

In JS, `OpenFeatureCommonAPI` reaches inside with `wrappedProvider.delegateManagesState` to determine whether to emit events or not. In Java, `ProviderRepository` reaches in with `newManager.delegateManagesState`. Go is doing the checks in both event executor and openfeature API. So we'd need to modify these components as well.

It's not a huge deal though, especially for a temporary migration. I find the proper adapter slightly cleaner but it is more work than throwing a bunch of if's in

specification/sections/02-providers.md
toddbaert: I will remove this line, I think.

I have some bigger concerns though.

> this may be problematic as it requires holding the status lock while all handlers run

I think we would also need some sort of lock to :

> Event handlers must only be run after all event handlers for the previous event has finished.

As far as I can tell, none of the SDKs currently do this. I'm a bit hesitant to add this though, since it means a misbehaving or blocked handler could block subsequent event emissions.

specification/sections/05-events.md
toddbaert: I think I've addressed this.

specification/appendix-e-migrations.md
toddbaert: Ah, OK, you were just talking about encapsulating this logic more elegantly then? If that's the case I agree. The PoCs were just there to prove the flow was possible, not necessarily in the cleanest way.

specification/sections/01-flag-evaluation.md
toddbaert: Yep.

specification/appendix-e-migrations.md
toddbaert: I agree this is a real concern. With provider-owned state, I don't see how the SDK can safely guarantee atomic late-attached handler execution or short-circuit evaluation without adding significant complexity (e.g. shared locks between the SDK and provider, or subscription-awareness in the provider).

That said, this is existing behavior that's important for usability; users need to be able to attach `PROVIDER_READY` handlers after initialization and still have them fire. Removing it would be a significant regression in developer experience.

Your suggestion about an event queue is interesting; if the SDK serializes event dispatch through a queue, late-attached handler checks could go through the same queue, which might also address concerns about event handler interleaving. Maybe we need this even though it will be heavy?

specification/sections/02-providers.md
toddbaert: Possible relates to: https://github.com/open-feature/spec/pull/380#discussion_r3255337860

dd-oleksii: > I think we would also need some sort of lock to :
>> Event handlers must only be run after all event handlers for the previous event has finished.
>
> As far as I can tell, none of the SDKs currently do this. I'm a bit hesitant to add this though, since it means a misbehaving or blocked handler could block subsequent event emissions.

Given that all event handlers are synchronous and most SDKs use either a queue or direct event firing, I think this is actually the default behavior we have today?

Perhaps relaxing this a bit we can have: for each event handler, it should be fired in the order of events happening and each invocation should wait for the previous invocation to finish (this corresponds to a per-subscriber queue). If we relax it more, that would allow the publisher to either publish events out-of-order or concurrently, and subscriber would have no way to establish the real order — not very great.

fwiw, I think we can drop this from the spec — SDKs should be able to figure out what's appropriate for each language

toddbaert:
I kept running into TOCTOU issues with the provider state, and no matter what abstractions I used to solve it, synthetic events were a problem. Rather than adding state back to the provider, [this proposal](https://github.com/open-feature/spec/pull/385) keeps the state in the SDK but removes all the synthetic event emission, which I believe was the real source of the issue. The SDK state is fully derived from the provider's event emission. The only remaining "synthetic" behavior is running a handler immediately when it's registered and the provider is already in the matching state (e.g. attaching a READY handler when the provider is already READY). This is less problematic since it doesn't involve an actual state transition; it's just replaying the current status to a late subscriber.

Marking this one as draft for now. cc @nicklasl @dd-oleksii

specification/appendix-e-migrations.md
toddbaert: I've taken a different approach: https://github.com/open-feature/spec/pull/380#issuecomment-4479544463

specification/appendix-e-migrations.md
toddbaert: See https://github.com/open-feature/spec/pull/380#issuecomment-4479544463

specification/sections/02-providers.md
toddbaert: Moot in the alternative approach: https://github.com/open-feature/spec/pull/380#issuecomment-4479544463

specification/appendix-e-migrations.md
dd-oleksii: Swift SDK is actually doing this events/subscription serialization with a lock in `ProviderStatusTracker`. Actual event handlers are run on a separate (per-subscriber) dispatch queue to prevent event handlers from locking the provider.

dd-oleksii: Thinking about this more, I believe we're confusing "event observing" and "state observing" and this adds a lot of accidental complexity.

Many languages have separate tools for a regular channel and a channel that remembers the last value:
- Swift: [`PassthroughSubject`](https://developer.apple.com/documentation/combine/passthroughsubject) vs [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject)
- Kotlin: [`SharedFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/) vs [`StateFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-state-flow/)
- Rust: [`broadcast`](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) vs [`watch`](https://docs.rs/tokio/latest/tokio/sync/watch/index.html) channels

The issue is that we want a weird mix of the two: we want it to remember the last state... except "not ready" status because it doesn't have a corresponding event to re-emit)... and don't remember "configuration changed" event (because it doesn't change status)... and "context changed" should be remapped to "ready" (but only if it's a replay and not an online observation).

This leads to a couple of issues. The obvious one is that the behavior is convoluted and is complicated to implement. In MultiProvider, we get a mismatch between emitted events and actual provider state (the spec requires MultiProvider to re-emit all events from the underlying providers, but underlying provider going into fatal state does not mean that the whole MultiProvider does).

And the last one is that waiting for `PROVIDER_READY` is already an unreliable way to wait for "ready" status. If provider is currently in reconciling status, most SDKs don't emit `PROVIDER_READY` after reconciliation is done. (This can easily happen if user does `api.setProvider(...); api.setContext(...); api.addEventHandler(PROVIDER_READY, ...);`)

I think a much cleaner API is to add "status observers" that remember the last status and are guaranteed to fire when the provider switches to the corresponding status (i.e. a successful reconciliation should invoke ready handler). And keep event observers without this last-value memory

We can even keep this backward-compatible for SDKs that have subscription topics (JS, Go) by saying that `PROVIDER_READY`/`PROVIDER_STALE`/`PROVIDER_ERROR`/`PROVIDER_RECONCILING` _are_ status observers. This would make `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_CONTEXT_CHANGED` the only true events (which is kinda already true today).

An alternative take is that this issue only exists for languages that have these native tools for broadcast and last-state observers (swift, kotlin), so SDKs decided to expose native observers instead of named event handlers as the spec describes. So now these SDKs are fighting with synchronizing these subjects/flows. As they are already deviating from the spec, perhaps they can deviate further to implement status subscription in addition to events (or alternatively align with the spec and implement named event handlers instead of subject/flow — though this is less idiomatic and will be more cumbersome for users)

cc @nicklasl @fabriziodemaria @typotter for your thoughts as this is a pain point in kotlin and swift

toddbaert: I haven't had time to fully consider this, but I think it's close to what I've proposed in my other pr: https://github.com/open-feature/spec/pull/385

It doesn't mention the "BehaviorSubject" construct exactly, but describes something basically similar. I agree that we could use those abstractions though.

Issue #366, opened 2026-02-25
NOT_READY event?


dd-oleksii:
There are a couple of cases when client's underlying provider may change its status to `NOT_READY`:
- SDK is shutdown
- domain provider is swapped for another provider, which is still initializing

However, there's no event to signal this.

Issue #265, opened 2024-06-13
Consider the ability to opt-out from automatic functionality such as automatic provider ready events.


kinyoklion:
For providers which support changing their provider status at runtime the automatic events introduce additional complexity.

The provider needs to defer its handling of the status until after the automatic status provided by the OpenFeature API. Otherwise it would emit duplicate provider ready or error events. The handling can be somewhat complex for multi-threaded providers and SDKs.

It would be nice if the provider had meta-data, or other fields, which could allow for opting out of the automatic behavior. This may then be extended to other automatic functionality in the future.

Related discussion: https://github.com/open-feature/dotnet-sdk/pull/276#issuecomment-2163766166

toddbaert:
I feel conflicted about this.

On the one hand, I see the value and the simplicity that can be gained here for underlying SDKs that emit events. On the other hand we're adding more complexity to the SDK, and cognitive load to provider authors.

One other question is I have is if this setting would also apply to if the `init` throws. Would we have some other indicator for that to prevent the same issue for duplicate error events?

kinyoklion:
> cognitive load to provider authors

I am not sure if this is true or not. I recently updated the LD node, python, java, and dotnet providers to handle this behavior and handling the mixed requirements of runtime events, initialization events, and the blocking behavior is by far the largest cognitive load in any of these provider implementations (aside from the node one, where we happen to have an underlying mechanism that matches well), and also very easy to get incorrect. 

This optionality may or may not be the correct way to reduce the complexity. Is optimizing for the simple case the correct approach, or is it likely that this increases complexity more often than it decreases it?

Ultimately here the event channel is shared. The provider can produce events and the OF SDK can also produce events. There isn't a delineation of the types of events provided by both. So you have to encode into the provider that specific events are emitted automatically at specific points, and that those events also need to be emitted by the provider at different points.

toddbaert:
I think I'm nearly convinced. What's your take on this?

> One other question is I have is if this setting would also apply to if the init throws. Would we have some other indicator for that to prevent the same issue for duplicate error events?

Do you think we should have 2 settings? 1 for preventing emission of error and 1 for emission of ready? Or can we handle both with one?

Do you think this should be a field on `ProviderMetadata`? Or should we have a dedicated field on the provider interface itself for it?

cc @lukas-reining since I think some of his recent work with a config-cat provider runs into this exact issue.

lukas-reining:
Having this data on the metadata feels pretty complicated to me. 

[...]

PR #385, opened 2026-05-18
fix: provider state ownership via events

### LINKED ISSUE #365: Provider lifecycle race condition in multi-threaded SDKs

toddbaert:
In short: providers emit events to signal all status transitions; the SDK derives provider status from these events rather than inferring from lifecycle method return values.

This resolves a race condition wherein after the provider's init() completes, but before the SDK has emitted its synthetic READY event, the provider emits some other event, and ordering is not preserved, or messages are lost.

### What changes for provider authors:
- Providers emit their own lifecycle events (PROVIDER_READY, PROVIDER_ERROR, etc.) instead of the SDK emitting them on the provider's behalf
- Trivial providers (no lifecycle methods, no event mechanism) are unaffected; the SDK treats them as ready immediately

### What changes for SDK authors:
- The SDK no longer fires synthetic events after lifecycle methods return
- Provider status is whatever the most recent event implies
- Short-circuit behavior for NOT_READY and FATAL is no longer required (it created its own race)

A migration appendix is included for SDKs supporting legacy providers.

Resolves #365

---

This is an alternative to #380 which took a different approach (adding getState() to providers). That one works but is ~10x larger and introduces concurrent-safety requirements on the provider that don't exist if state lives only in the SDK. Same problem, simpler solution IMO. Happy to discuss tradeoffs.

### DIFF
--- a/specification.json
+++ b/specification.json
@@ -553,7 +553,7 @@
         {
             "id": "Requirement 2.8.1",
             "machine_id": "requirement_2_8_1",
-            "content": "The provider MUST emit an event to signal each status transition, including transitions resulting from lifecycle methods (`initialize`, `shutdown`, `on context change`) and spontaneous transitions.",
+            "content": "The provider MUST emit an event to signal each status transition, including transitions resulting from lifecycle methods (`initialize`, `on context changed`) and spontaneous transitions.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -610,7 +610,7 @@
         {
             "id": "Condition 2.8.5",
             "machine_id": "condition_2_8_5",
-            "content": "The provider does not define an `initialize` function or an event emission mechanism.",
+            "content": "The provider does not define an `initialize` function.",
             "RFC 2119 keyword": null,
             "children": [
                 {
--- a/specification.json
+++ b/specification.json
@@ -582,7 +582,7 @@
         {
             "id": "Requirement 2.8.1",
             "machine_id": "requirement_2_8_1",
-            "content": "The provider MUST emit an event to signal each status transition, including transitions resulting from lifecycle methods (`initialize`, `shutdown`, `on context change`) and spontaneous transitions.",
+            "content": "The provider MUST emit an event to signal each status transition, including transitions resulting from lifecycle methods (`initialize`, `on context changed`) and spontaneous transitions.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -310,8 +310,8 @@
                 {
                     "id": "Conditional Requirement 1.7.2.1",
                     "machine_id": "conditional_requirement_1_7_2_1",
-                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, or `ERROR`, the  `provider status` accessor must support possible value `RECONCILING`.",
-                    "RFC 2119 keyword": null,
+                    "content": "In addition to `NOT_READY`, `READY`, `STALE`, `ERROR`, or `FATAL`, the `provider status` accessor MUST support possible value `RECONCILING`.",
+                    "RFC 2119 keyword": "MUST",
                     "children": []
                 }
             ]
@@ -319,21 +319,21 @@
         {
             "id": "Requirement 1.7.3",
             "machine_id": "requirement_1_7_3",
-            "content": "The client's `provider status` accessor MUST indicate `READY` if the `initialize` function of the associated provider terminates normally.",
+            "content": "The client's `provider status` accessor MUST indicate `READY` after the provider emits `PROVIDER_READY` following initialization.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.7.4",
             "machine_id": "requirement_1_7_4",
-            "content": "The client's `provider status` accessor MUST indicate `ERROR` if the `initialize` function of the associated provider terminates abnormally.",
+            "content": "The client's `provider status` accessor MUST indicate `ERROR` after the provider emits `PROVIDER_ERROR` following initialization.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 1.7.5",
             "machine_id": "requirement_1_7_5",
-            "content": "The client's `provider status` accessor MUST indicate `FATAL` if the `initialize` function of the associated provider terminates abnormally and indicates `error code` `PROVIDER_FATAL`.",
+            "content": "The client's `provider status` accessor MUST indicate `FATAL` after the provider emits `PROVIDER_ERROR` with error code `PROVIDER_FATAL` following initialization.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -563,6 +563,49 @@
             "content": "The `provider` MAY define a function for tracking the occurrence of a particular user action or application state, with parameters `tracking event name` (string, required), `evaluation context` (optional) and `tracking event details` (optional) which returns nothing.",
             "RFC 2119 keyword": "MAY",
             "children": []
+        },
+        {
+            "id": "Requirement 2.8.1",
+            "machine_id": "requirement_2_8_1",
+            "content": "The provider MUST emit an event to signal each status transition, including transitions resulting from lifecycle methods (`initialize`, `shutdown`, `on context change`) and spontaneous transitions.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.8.2",
+            "machine_id": "requirement_2_8_2",
+            "content": "The provider MUST emit `PROVIDER_READY` if its `initialize` function terminates normally.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.8.3",
+            "machine_id": "requirement_2_8_3",
+            "content": "The provider MUST emit `PROVIDER_ERROR` if its `initialize` function terminates abnormally.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Requirement 2.8.4",
+            "machine_id": "requirement_2_8_4",
+            "content": "The provider MUST emit `PROVIDER_CONTEXT_CHANGED` if its `on context changed` function terminates normally, and `PROVIDER_ERROR` if it terminates abnormally.",
+            "RFC 2119 keyword": "MUST",
+            "children": []
+        },
+        {
+            "id": "Condition 2.8.5",
+            "machine_id": "condition_2_8_5",
+            "content": "The provider does not define an `initialize` function or an event emission mechanism.",
+            "RFC 2119 keyword": null,
+            "children": [
+                {
+                    "id": "Conditional Requirement 2.8.5.1",
+                    "machine_id": "conditional_requirement_2_8_5_1",
+                    "content": "The SDK MUST treat such providers as `READY` from registration and MUST run `PROVIDER_READY` handlers on their behalf.",
+                    "RFC 2119 keyword": "MUST",
+                    "children": []
+                }
+            ]
         },
         {
             "id": "Requirement 3.1.1",
@@ -984,8 +1027,8 @@
         {
             "id": "Requirement 5.1.1",
             "machine_id": "requirement_5_1_1",
-            "content": "The `provider` MAY define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with a `provider event details` payload.",
-            "RFC 2119 keyword": "MAY",
+            "content": "The `feature provider` interface MUST define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED`, `PROVIDER_STALE`, `PROVIDER_RECONCILING`, and `PROVIDER_CONTEXT_CHANGED`, with a `provider event details` payload.",
+            "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
@@ -1068,14 +1111,14 @@
         {
             "id": "Requirement 5.3.1",
             "machine_id": "requirement_5_3_1",
-            "content": "If the provider's `initialize` function terminates normally, `PROVIDER_READY` handlers MUST run.",
+            "content": "When the provider emits `PROVIDER_READY`, associated `PROVIDER_READY` handlers MUST run.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
         {
             "id": "Requirement 5.3.2",
             "machine_id": "requirement_5_3_2",
-            "content": "If the provider's `initialize` function terminates abnormally, `PROVIDER_ERROR` handlers MUST run.",
+            "content": "When the provider emits `PROVIDER_ERROR`, associated `PROVIDER_ERROR` handlers MUST run.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
@@ -1118,7 +1161,7 @@
         {
             "id": "Requirement 5.3.5",
             "machine_id": "requirement_5_3_5",
-            "content": "If the provider emits an event, the value of the client's `provider status` MUST be updated to the status associated with that event **before** the SDK invokes any event handlers for that event, so that handlers observe a consistent status.",
+            "content": "When a provider emits an event, the SDK MUST update the `provider status` to the status associated with that event **before** invoking any event handlers for that event, so that handlers observe a consistent status.",
             "RFC 2119 keyword": "MUST",
             "children": []
         },
--- a/specification.json
+++ b/specification.json
@@ -340,27 +340,13 @@
         {
             "id": "Requirement 1.7.6",
             "machine_id": "requirement_1_7_6",
-            "content": "The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 1.7.7",
-            "machine_id": "requirement_1_7_7",
-            "content": "The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `FATAL`.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 1.7.8",
-            "machine_id": "requirement_1_7_8",
             "content": "Implementations SHOULD propagate the `error code` returned from any provider lifecycle methods.",
             "RFC 2119 keyword": "SHOULD",
             "children": []
         },
         {
-            "id": "Requirement 1.7.9",
-            "machine_id": "requirement_1_7_9",
+            "id": "Requirement 1.7.7",
+            "machine_id": "requirement_1_7_7",
             "content": "The client's `provider status` accessor MUST indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.",
             "RFC 2119 keyword": "MUST",
             "children": []
--- a/specification.json
+++ b/specification.json
@@ -355,27 +355,13 @@
         {
             "id": "Requirement 1.7.6",
             "machine_id": "requirement_1_7_6",
-            "content": "The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `NOT_READY`.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 1.7.7",
-            "machine_id": "requirement_1_7_7",
-            "content": "The client MUST default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `FATAL`.",
-            "RFC 2119 keyword": "MUST",
-            "children": []
-        },
-        {
-            "id": "Requirement 1.7.8",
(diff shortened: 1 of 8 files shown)


erka (approved):
Nice work. Migrating providers will be challenging, but it’s for the best.

toddbaert:
> Nice work. Migrating providers will be challenging, but it’s for the best.

Agree

specification/sections/02-providers.md
```diff
@@ -305,3 +307,52 @@ The track function performs side effects required to record the `tracking event`
...
+
+### 2.8. Provider status
+
+[![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
+
+The SDK derives provider status from events emitted by the provider.
+Providers signal all state transitions by emitting the appropriate event; the SDK updates its internal status accordingly and runs associated handlers.
+
+Providers that do not define lifecycle methods or an event emission mechanism cannot emit events by design; see Condition 2.8.5.
+Requirements 2.8.1-2.8.4 apply only to providers that define lifecycle methods and an event emission mechanism.
```
lukas-reining: Can a provider without lifecycle methods but an event emission mechanism emit events? 
I feel like in this case this works.  But is does not work if the provider has lifecycle methods but no event emission mechanism. Right? 
I feel like `2.8.1` indirectly requires an eventing mechanism if there is lifecycle methods.

specification/sections/02-providers.md
```diff
@@ -305,3 +307,52 @@ The track function performs side effects required to record the `tracking event`
...
+
+#### Requirement 2.8.4
+
+> The provider **MUST** emit `PROVIDER_CONTEXT_CHANGED` if its `on context changed` function terminates normally, and `PROVIDER_ERROR` if it terminates abnormally.
+
+see: [provider context reconciliation](#26-provider-context-reconciliation)
+
+#### Condition 2.8.5
+
+> The provider does not define an `initialize` function or an event emission mechanism.
```
lukas-reining: As above I am confused by the `or`. 
If a provider does not have an initialize function we run `PROVIDER_READY`. 
If I get `2.8.1` right, it could basically be:

```suggestion
> The provider does not define an `initialize` function.
```

specification/sections/02-providers.md
```diff
@@ -305,3 +307,52 @@ The track function performs side effects required to record the `tracking event`
...
+
+#### Requirement 2.8.2
+
+> The provider **MUST** emit `PROVIDER_READY` if its `initialize` function terminates normally.
+
+#### Requirement 2.8.3
+
+> The provider **MUST** emit `PROVIDER_ERROR` if its `initialize` function terminates abnormally.
+
+If the error is irrecoverable, the error code must indicate `PROVIDER_FATAL`.
```
lukas-reining, line 364: in `Requirement 5.3.5` of the current spec we say:

```
Some providers may emit events spontaneously, based on changes in their internal state (connections, caches, etc). The SDK must update its internal representation of the provider's state accordingly:
```

Should we add a requirement here that says that providers `may` emit a `PROVIDER_ERROR` at any time? I know that we also not say that they can't but I think it would make it clearer. 
This would make the non normative from the beginning more misleading as the provider may error without initialization:
`Providers that do not define lifecycle methods or an event emission mechanism cannot emit events by design`

specification/appendix-e-migrations.md
```diff
@@ -0,0 +1,44 @@
...
+
+### For SDK authors
+
+SDKs must detect whether a provider emits its own lifecycle events, via an opt-in marker (e.g. an interface, boolean property, or type-level tag).
+
+- **Marker present:** the SDK does not emit synthetic lifecycle events. It derives status solely from the provider's event stream.
+- **Marker absent (legacy):** the SDK emits synthetic lifecycle events after lifecycle methods return, as before. This path is deprecated.
+
+The legacy path should be deprecated in the release that introduces the marker, with removal targeted for the next major version.
+SDK authors should update any first-party providers and provider base classes to emit their own lifecycle events.
```
lukas-reining, line 64: This makes much sense to me!

lukas-reining (approved):
This looks good to me! Makes sense to go forward with this. 
I left some small questions/remarks.

specification/sections/05-events.md
```diff
@@ -205,45 +206,45 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
 
 ##### Conditional Requirement 5.3.4.2
 
 > If the provider's `on context changed` function terminates normally, and no other invocations have yet to terminate, associated `PROVIDER_CONTEXT_CHANGED` handlers **MUST** run.
 
-The implementation must run any `PROVIDER_CONTEXT_CHANGED` handlers associated with the provider after the provider has reconciled its state and returned from the `on context changed` function.
-The `PROVIDER_CONTEXT_CHANGED` is not emitted from the provider itself; the SDK implementation must run the `PROVIDER_CONTEXT_CHANGED` handlers if the `on context changed` function terminates normally.
-It's possible that the `on context changed` function is invoked simultaneously or in quick succession; in this case the SDK will only run the `PROVIDER_CONTEXT_CHANGED` handlers after all reentrant invocations have terminated, and the last to terminate was successful (terminated normally).
+The provider must emit `PROVIDER_CONTEXT_CHANGED` after it has successfully reconciled the context.
+The `on context changed` function may be invoked simultaneously or in quick succession; the provider must be prepared to handle this case.
```
nicklasl: We are loosening the instruction a bit here around when the `PROVIDER_CONTEXT_CHANGED` is meant to run, is that deliberate? 
Could/should we have that somewhere else?

specification/sections/01-flag-evaluation.md
```diff
@@ -477,55 +477,37 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
-see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
-
-#### Requirement 1.7.7
-
-> The client **MUST** default, run error hooks, and indicate an error if flag resolution is attempted while the provider is in `FATAL`.
-
-The client defaults and returns the `PROVIDER_FATAL` `error code` if evaluation is attempted after the provider has transitioned to an irrecoverable error state.
-The SDK avoids calling the provider's resolver functions entirely ("short-circuits") if the provider is in this state.
-
-see: [error codes](../types.md#error-code), [flag value resolution](./02-providers.md#22-flag-value-resolution)
```
nicklasl, line 531: Do we have the client defaults case on evaluation during NOT_READY/FATAL defined somewhere else?
I guess we'll need to drop the "short-circuit" parts here but just want to make sure we keep the definitions on the defaults.

We could consider moving this into 1.4 alongside 1.4.10, without the short circuit of course.

specification.json
```diff
@@ -310,57 +310,43 @@
...
-            "machine_id": "requirement_1_7_8",
             "content": "Implementations SHOULD propagate the `error code` returned from any provider lifecycle methods.",
             "RFC 2119 keyword": "SHOULD",
             "children": []
         },
         {
-            "id": "Requirement 1.7.9",
-            "machine_id": "requirement_1_7_9",
+            "id": "Requirement 1.7.7",
+            "machine_id": "requirement_1_7_7",
```
nicklasl: Nit (and very much a question to learn): I'm not very familiar with how these work, but assuming that code/tests are being generated from them. Is this re-numbering a good idea?

specification/sections/02-providers.md
nicklasl: Also a bit confused.
With 5.1.1 going from `MAY` to `MUST`, does that mean providers without an event mechanism are not conformant to this version of the spec? 
With this condition it seems we (SDK) intentionally support them. I guess that is just for the time being? Or expected to long term?

specification/sections/05-events.md
```diff
@@ -205,45 +206,45 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
-It's possible that the `on context changed` function is invoked simultaneously or in quick succession; in this case the SDK will only run the `PROVIDER_ERROR` handlers after all reentrant invocations have terminated, and the last to terminate was unsuccessful (terminated abnormally).
+The provider must emit `PROVIDER_ERROR` if reconciling the context fails.
+The `on context changed` function may be invoked simultaneously or in quick succession; the provider must be prepared to handle that.
 
 see: [provider event types](../types.md#provider-events), [provider events](#51-provider-events), [provider context reconciliation](02-providers.md#26-provider-context-reconciliation)
 
 #### Requirement 5.3.5
 
-> If the provider emits an event, the value of the client's `provider status` **MUST** be updated to the status associated with that event **before** the SDK invokes any event handlers for that event, so that handlers observe a consistent status.
+> When a provider emits an event, the SDK **MUST** update the `provider status` to the status associated with that event **before** invoking any event handlers for that event, so that handlers observe a consistent status.
```
nicklasl, line 234: The PR description states:
> Trivial providers (no lifecycle methods, no event mechanism) are unaffected; the SDK treats them as ready immediately.

This requirement contradicts that.

nicklasl (approved):

jonathannorris (approved):

kinyoklion:
I've not been following along with the gherkin effort, but the re-numbering in this PR would affect some of the gherkin tagging. https://github.com/open-feature/spec/blob/203c25f93495cee3a7191bf906dbdf61939e5f68/specification/assets/gherkin/evaluation_v2.feature#L195

kinyoklion (approved):
Approving the idea. There are some inline comments from others that I agree need some clarification.

specification.json
toddbaert: In some cases yes, and I generally like to avoid too much thrashing, but I think it's better than completely incoherent numbering.

specification/sections/01-flag-evaluation.md
toddbaert: I think this inherently becomes a provider requirement; it's basically an extension of 2.2.7. We need to get rid of the short-circuit in the SDK because it re-introduces the race (the SDK has to check state to short-circuit, and that state can lag the provider's actual state). The cleanest fix is pushing the responsibility down to the provider, which knows its own state without lag.

I'd lean toward keeping this in section 2 rather than 1.4. The provider is the only party that can answer "am I ready?" without a race, so it makes sense to put the obligation there. We could either add a non-normative note to 2.2.7 or add a new requirement for it. WDYT?

specification/sections/05-events.md
toddbaert: I don't understand how this line in particular contradicts the sentence you're quoting. Maybe you meant to make this comment on 5.1.1? If so, let me clarify: the thing 5.1.1 is describing is the abstraction that an implementation (the SDK) must define; it doesn't require every provider to make use of that abstraction.

The things providers themselves must do are generally described in section 2.

Totally willing to work to make this clearer. Please let me know if I've misunderstood.

specification/sections/02-providers.md
toddbaert: Good point; agreed. Simplified per your suggestion.

specification/sections/02-providers.md
toddbaert: Good observation. I think this is fine implicitly; if a provider has lifecycle methods it needs a way to signal completion. Let me know if you think we need to make it explicit.

specification/sections/02-providers.md
toddbaert: Added a non-normative note in 2.8.5.1's explanation clarifying that spontaneous error emission is still possible and handled per 5.3.5. Let me know if that addresses it.

specification/sections/05-events.md
toddbaert: I don't think anything was loosened normatively; 5.3.4.2's MUST still encodes both the timing ("terminates normally") and the coalescing ("no other invocations have yet to terminate"). That said, I dropped some of the explanatory restatement when reframing the mechanism around the provider emitting the event, and I think you were right that this made the requirement harder to follow at a glance. I've restored the explanatory clarification; let me know if that helps, or if I'm missing something.

toddbaert:
Good catch; I'll update the gherkin tags as a follow-up. Agreed it's better to take the renumbering hit once than to live with permanently incoherent numbering.

specification/sections/02-providers.md
toddbaert: 5.1.1 describes the abstraction that an implementation (the SDK) must define; it doesn't require every provider to make use of that abstraction. The things providers themselves must do are generally described in section 2. Trivial providers (no lifecycle) are intentionally supported long-term via Condition 2.8.5; the SDK handles them by emitting PROVIDER_READY on their behalf at registration.

specification/sections/02-providers.md
toddbaert: Related: https://github.com/open-feature/spec/pull/385#discussion_r3501751413

specification/sections/05-events.md
```diff
@@ -31,10 +31,11 @@ see: [domain](../glossary.md#domain)
...
 #### Requirement 5.1.1
 
-> The `provider` **MAY** define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED` and `PROVIDER_STALE`, with a `provider event details` payload.
+> The `feature provider` interface **MUST** define a mechanism for signaling the occurrence of one of a set of events, including `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED`, `PROVIDER_STALE`, `PROVIDER_RECONCILING`, and `PROVIDER_CONTEXT_CHANGED`, with a `provider event details` payload.
 
-Providers cannot emit `PROVIDER_CONTEXT_CHANGED` or `PROVIDER_RECONCILING` events.
-These are emitted only by the SDK during context reconciliation.
+Providers must emit events to signal all state transitions, including those resulting from lifecycle methods (initialize, reconciliation).
+The SDK derives provider status from these events.
+Providers without lifecycle methods or an event emission mechanism cannot emit events by design; see [Condition 2.8.5](./02-providers.md#condition-285).
```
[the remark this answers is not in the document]
toddbaert: Clarified with latest changeset.

specification/appendix-e-migrations.md
```diff
@@ -0,0 +1,44 @@
...
+# Appendix E: Migrations
+
+This appendix provides non-normative guidance for provider authors and SDK authors on migrating to new or changed specification requirements.
+
+## Provider Lifecycle Event Emission
+
+### Background
+
+Prior to `v0.9.0`, the SDK emitted synthetic lifecycle events (`PROVIDER_READY`, `PROVIDER_ERROR`) on behalf of providers after lifecycle methods (`initialize`, `shutdown`, `on context change`) returned.
+This created a race condition in multi-threaded SDKs: the provider could emit events from background threads concurrently with SDK-emitted synthetic events, resulting in incorrect event ordering and inconsistent status.
```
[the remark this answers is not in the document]
toddbaert: Fixed.

toddbaert:
Follow-up: the gherkin tags for the affected scenarios (formerly `@spec-1.7.6` and `@spec-1.7.7`) have been retagged to `@spec-2.2.7` in this PR (see `evaluation_v2.feature`). The old 1.7.6/1.7.7 have been removed; the behavior they described (client returns default on NOT_READY/FATAL) is now covered by the provider's abnormal error indication (2.2.7) combined with 1.4.10.

toddbaert:
@kinyoklion following up on the migration-burden concerns you raised in #365 and #265 (the "cognitive load / most intuitive way is wrong / duplicate events" issues) — I've expanded Appendix E to cover concrete migration patterns for the cases you mentioned: blocking init, event-driven init (the LD-style pattern), and context reconciliation coalescing. I also added anti-patterns and made the legacy-path behavior explicit for SDK authors.

See the [updated appendix](https://github.com/open-feature/spec/pull/385/files#diff-caa8091762365e26382a8d43f705aa48e4a1076f). LMK if there are patterns you hit during the LD provider migrations that aren't covered.

specification/sections/05-events.md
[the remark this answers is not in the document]
toddbaert: Try again now?

toddbaert:
Merging this; we can make adjustments if needed as we implement.

specification/sections/05-events.md
```diff
@@ -23,18 +23,19 @@ graph
...
```
dd-oleksii, line 38: Is this true? My understanding is that lifecycle methods and event emission are currently decoupled and a provider could emit events even if it does not implement lifecycle methods

specification/appendix-e-migrations.md
```diff
@@ -0,0 +1,76 @@
...
+Providers with no `initialize` function are handled per [Condition 2.8.5](./sections/02-providers.md#condition-285) and require no changes.
+
+#### Common migration patterns
+
+**Blocking initialization.**
+If `initialize` performs setup synchronously and returns when ready, emit `PROVIDER_READY` immediately before returning (or `PROVIDER_ERROR` before throwing).
+
+**Event-driven initialization.**
+If setup completes asynchronously (e.g. the underlying vendor SDK signals readiness via its own event), forward that signal as `PROVIDER_READY`.
+`initialize` may return before the event is emitted; the SDK treats the return as a synchronization signal only.
```
dd-oleksii, line 45: This recommendation looks like it may break "set provider and wait" methods? What does "synchronization signal" mean?

specification/sections/01-flag-evaluation.md
```diff
@@ -490,59 +490,34 @@ see: [static-context paradigm](../glossary.md#static-context-paradigm)
...
-The SDK ensures that if the provider's lifecycle methods terminate with an `error code`, that error code is included in any associated error events and returned/thrown errors/exceptions.
-
-see: [error codes](../types.md#error-code)
-
-#### Requirement 1.7.9
-
 > The client's `provider status` accessor **MUST** indicate `NOT_READY` once the `shutdown` function of the associated provider terminates.
 
 Regardless of the success of the provider's `shutdown` function, the `provider status` should convey the provider is no longer ready to use once the shutdown function terminates.
+Unlike other status transitions, this shutdown is inferred by the SDK; because the SDK initiates the shutdown call, no event from the provider is required.
```
dd-oleksii, line 520: Not directly related to the change here but not having a "not ready" event creates an asymmetry where users can't be notified when a provider becomes not ready. If application author subscribes too all events, they'd still not have the up-to-date status of the provider. https://github.com/open-feature/spec/issues/366

specification/sections/05-events.md
toddbaert: I think this sentence is redundant and ambiguous:

> Providers without lifecycle methods or an event emission mechanism cannot emit events by design.

A provider without lifecycle methods can still emit spontaneous events. A provider without an event-emission mechanism obviously cannot.

The useful constraint is the other direction: under the new model, a provider implementing initialize must also support events so it can emit PROVIDER_READY or PROVIDER_ERROR. Where practical, SDKs should couple those abstractions. The legacy migration path is the temporary exception.

specification/appendix-e-migrations.md
toddbaert: Good catch. As written, this is under-specified. `initialize` returning should only complete the method invocation; it must not settle setProviderAndWait or change provider status by itself.

`setProviderAndWait` should wait until initialize has terminated and the corresponding `PROVIDER_READY` or `PROVIDER_ERROR` event has been processed by the SDK. It does not need to wait for application handlers. If initialize returns first, the wait remains pending.

I’ll clarify this in 1.1.2.4 and Appendix E.

It also reinforces what I mentioned [here](https://github.com/open-feature/spec/pull/385#discussion_r3572677128).

toddbaert: See: https://github.com/open-feature/spec/pull/408

specification/sections/01-flag-evaluation.md
toddbaert: See: https://github.com/open-feature/spec/pull/408

Issue #258, opened 2024-05-09
tooling: Publish repo spec compliance docker container to github packages


justinabrahms:
See #254 . We should get that auto-publishing for easy `docker run ghcr.io/...` usage.

beeme1mr:
@justinabrahms I wonder if we should create a GitHub action to run the spec checker from this repo. We could create a matrix of SDK repos that are checked out and parsed. The output could be aggregated into a report. We could even automatically open issues in repos where gaps are detected.

PR #389, opened 2026-06-05
feat: Create a GitHub action to validate test coverage/compliance with the spec for SDK repos

JamieSinn:
## This PR
- Adds a github action and docs for use in downstream SDK repos to make sure that the SDKs have adequate test coverage and appropriately flag where and how each component of the tests are validating the spec.

### Related Issues
#258 
This is stacked on #388 and #388 should be merged first.

### Follow-up Tasks

- Implement this in all the SDK repos that are >= 1.0 as a required test, and for all non 1.0 make this a non-required action.
- Update existing 1.0+ descriptions where they've fallen out of date
- Identify/create tickets for any gaps in test coverage for the spec in 1.0+ repos

### DIFF
--- a/tools/repo_parser/action.yml
+++ b/tools/repo_parser/action.yml
@@ -1,0 +1,83 @@
+name: OpenFeature Spec Compliance Check
+description: Checks an OpenFeature SDK repo for spec compliance using the repo_parser tool.
+
+inputs:
+  image-version:
+    description: Tag of the ghcr.io/open-feature/spec/repo_parser image to use.
+    required: false
+    default: latest
+
+outputs:
+  report:
+    description: Full JSON compliance report ({ missing, extra, different-text, good }).
+    value: ${{ steps.check.outputs.report }}
+  compliant:
+    description: "'true' if fully compliant with the spec, 'false' otherwise."
+    value: ${{ steps.check.outputs.compliant }}
+  missing:
+    description: JSON array of spec requirement IDs not covered by the repo (e.g. ["1.1.1","1.2.3"]).
+    value: ${{ steps.check.outputs.missing }}
+  extra:
+    description: JSON array of test reference IDs not found in the spec.
+    value: ${{ steps.check.outputs.extra }}
+  different-text:
+    description: JSON array of spec requirement IDs whose text does not match the spec.
+    value: ${{ steps.check.outputs.different_text }}
+  missing-count:
+    description: Number of spec requirements not covered by the repo.
+    value: ${{ steps.check.outputs.missing_count }}
+  extra-count:
+    description: Number of test references not found in the spec.
+    value: ${{ steps.check.outputs.extra_count }}
+  different-text-count:
+    description: Number of spec requirements with mismatched text.
+    value: ${{ steps.check.outputs.different_text_count }}
+
+runs:
+  using: composite
+  steps:
+    - name: Check spec compliance
+      id: check
+      shell: bash
+      run: |
+        ARGS="--code-directory /appdir --json-report --refresh-spec --diff-output"
+
+        set +e
+        docker run \
+          --mount type=bind,src=${{ github.workspace }},target=/appdir \
+          -w /appdir \
+          ghcr.io/open-feature/spec/repo_parser:${{ inputs.image-version }} \
+          $ARGS
+        EXIT_CODE=$?
+        set -e
+
+        REPORT_FILE=$(ls "${{ github.workspace }}"/*-report.json 2>/dev/null | head -1)
+        if [[ -n "$REPORT_FILE" ]]; then
+          REPORT=$(cat "$REPORT_FILE")
+          rm "$REPORT_FILE"
+
+          DELIMITER=$(openssl rand -hex 8)
+          echo "report<<$DELIMITER" >> "$GITHUB_OUTPUT"
+          echo "$REPORT"           >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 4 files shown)


tools/repo_parser/action.yml
```diff
@@ -0,0 +1,83 @@
...
+          $ARGS
+        EXIT_CODE=$?
+        set -e
+
+        REPORT_FILE=$(ls "${{ github.workspace }}"/*-report.json 2>/dev/null | head -1)
+        if [[ -n "$REPORT_FILE" ]]; then
+          REPORT=$(cat "$REPORT_FILE")
+          rm "$REPORT_FILE"
+
+          DELIMITER=$(openssl rand -hex 8)
```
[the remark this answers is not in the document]
JamieSinn: This runs on Ubuntu; which has openssl.

.github/workflows/publish-repo-parser.yaml
```diff
@@ -0,0 +1,47 @@
+name: Publish repo_parser image to GHCR
```
jonathannorris, line 1: I think we are going to need a more descriptive name than `repo_parser`

JamieSinn: Suggestions? That's literally what it's doing/the script name so I mostly defaulted to being visually descriptive.

jonathannorris (approved):

tools/repo_parser/examples/spec-compliance.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  spec-compliance:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      # Binding to the main branch is not secure. Replace this with the specific commit SHA of the repo_parser image you want to use for a more secure setup.
+      - uses: open-feature/spec/tools/repo_parser@main
```
kinyoklion, line 21: Potentially a tangent, or future work, but it feels to me like we should have a release-please workflow for publishing these. Then RP could update this when it made its release PR.

kinyoklion: Though generally I think the comment is overly cautious, because the action is in the repo. But I would be more concerned about using latest from GHCR.

tools/repo_parser/action.yml
```diff
@@ -0,0 +1,83 @@
+name: OpenFeature Spec Compliance Check
+description: Checks an OpenFeature SDK repo for spec compliance using the repo_parser tool.
+
+inputs:
+  image-version:
+    description: Tag of the ghcr.io/open-feature/spec/repo_parser image to use.
+    required: false
+    default: latest
```
kinyoklion, line 8: Maybe something like:
```yaml
# renovate: datasource=docker depName=ghcr.io/open-feature/spec/repo_parser
default: "latest@sha256:9efeb6197492e839130dd27411399af237beb4d2662b849a95eb444ac15557aa"
```

Then a customer renovate updater?
```json
{
  "customManagers": [
    {
      "customType": "regex",
      "versioningTemplate": "docker",
      "managerFilePatterns": ["/(^|/)action\\.ya?ml$/"],
      "matchStrings": [
        "# renovate: datasource=(?<datasource>\\S+) depName=(?<depName>\\S+)\\s+default:\\s*[\"']?(?<currentValue>[^@\"'\\s]+)@(?<currentDigest>sha256:[a-f0-9]+)[\"']?"
      ]
    }
  ]
}
```

tools/repo_parser/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        echo "compliant=$([[ $EXIT_CODE -eq 0 ]] && echo 'true' || echo 'false')" >> "$GITHUB_OUTPUT"
+        exit $EXIT_CODE
+
+    - name: Log compliance summary
+      if: always()
+      shell: bash
+      run: |
+        echo "Compliant:       ${{ steps.check.outputs.compliant }}"
+        echo "Missing:         ${{ steps.check.outputs.missing }}"
+        echo "Extra:           ${{ steps.check.outputs.extra }}"
```
kinyoklion, line 82: This shouldn't be interpolated like this and instead it should use the environment variable. Because interpolations are ran before the shell execution and you can technically use injection attacks against them, so it is better to avoid it.

tools/repo_parser/action.yml
erka: Renovate only creates a PR if there is a security vulnerability with the current setup.

kinyoklion: Yeah, my main thought here is just not to use `latest`.

tools/repo_parser/action.yml
JamieSinn: Conceptually yes; however this is a pretty print of the output from the previous step. Not sure where you see an attack vector here.

tools/repo_parser/examples/spec-compliance.yml
JamieSinn: Attestation is a thing still here - and I'm less concerned about using the latest for GHCR than I am with the repo itself; as there's a lot of things that can be weird with it. 

Concept wise - this is intended to be sourced from other repos; which should be updating their spec reference based on this status check - not based on the latest revision on main as that may not be a released version of the spec, so the concept applies here to use the specific hash/release tag.

RP would be a nice way to update these downstream - I'll toy around with that later.

PR #393, opened 2026-06-18
feat: supply the bound domain to provider initialization

jonathannorris:
Amends provider initialization so the `domain` a provider is bound to is supplied to its `initialize` function, letting providers scope domain-specific behavior (such as partitioning a persistent cache) to that domain.

## Motivation

Provider-level persistent caches (e.g. the OFREP web provider's local persistence) currently have no way to know which `domain` they're bound to, so two providers on the same storage partition can collide on cache entries even though OpenFeature's domain model is supposed to keep them isolated. See [slack thread](https://cloud-native.slack.com/archives/C06E4DE6S07/p1780526010480869), [open-feature/js-sdk-contrib#1566](https://github.com/open-feature/js-sdk-contrib/pull/1566), and [ADR 0009](https://github.com/open-feature/protocol/blob/main/service/adrs/0009-local-storage-for-static-context-providers.md).

A follow-up OFREP ADR will add the domain to the cache key.

## Changes

Two parts, both kept language-agnostic so implementations can use overloads, optional arguments, etc.:

**Supply the bound domain to init** (folded into the existing init requirements, per earlier review feedback):

- `2.4.1` (provider): the `initialize` function accepts the global `evaluation context` **and the bound `domain`** (if any).
- `1.1.2.2` (API): the `provider mutator` **MUST** supply the bound `domain` (if any) when invoking `initialize`.

**Let a provider declare it's domain-scoped, enforced by the API** (new requirements):

- `2.4.3` (provider): a `provider` **MAY** declare itself `domain-scoped` when it holds per-`domain` state such as a persistent cache.
- `1.1.8` (condition) / `1.1.8.1` (API): when a provider is `domain-scoped`, the `provider mutator` **MUST NOT** bind it to more than one `domain`, and rejects any attempt to do so.

## Resolved

The earlier open question (a provider bound to multiple `domains` is initialized once and receives only its first binding's `domain`) is now addressed for stateful providers: declaring `domain-scoped` restricts the instance to a single `domain`, so the `domain` it keys on is unambiguous. Stateless providers are unaffected and can still back multiple `domains`.

### DIFF
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -221,6 +221,13 @@ Most providers are stateless with respect to their `domain` and can safely back 
 Providers that persist or cache `domain`-specific data need a stable, unambiguous `domain` to key that state on.
 By declaring itself `domain-scoped`, such a provider signals that the `API` must bind it to at most one `domain` (see [Requirement 1.1.8](./01-flag-evaluation.md#condition-118)), guaranteeing the `domain` supplied to `initialize` is the only one the instance will ever serve.
 
+#### Requirement 2.4.4
+
+> A `provider` that declares itself `domain-scoped` **MUST** accept the bound `domain` during initialization.
+
+A `domain-scoped` declaration is only meaningful if the provider consumes the `domain` it is given to scope its state.
+This is a contract on the provider; implementations may not be able to detect or reject a violation automatically, so it is not guaranteed to surface as a runtime error.
+
 ### 2.5. Shutdown
 
 [![hardening](https://img.shields.io/static/v1?label=Status&message=hardening&color=yellow)](https://github.com/open-feature/spec/tree/main/specification#hardening)
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -180,16 +180,18 @@ The `domain` the provider is registered under is also supplied, allowing the pro
 A `provider` instance is initialized only once, even when bound to multiple `domains`; in that case the `domain` supplied is the one under which it was first registered.
 The default provider, which is not bound to a domain, is initialized without one.
 
-```typescript
+```java
 // MyProvider implementation of the initialize function defined in Provider
 class MyProvider implements Provider {
   //...
 
   // the global context and the bound domain are passed to the initialization function
-  async initialize(initialContext: EvaluationContext, domain?: string): Promise<void> {
+  void initialize(EvaluationContext initialContext, String domain) {
     this.domain = domain;
-    // A hypothetical initialization function: make an initial call doing some bulk initial evaluation, start a worker to do periodic updates
-    this.flagCache = await this.restClient.bulkEvaluate(initialContext);
+    /*
+      A hypothetical initialization function: make an initial call doing some bulk initial evaluation, start a worker to do periodic updates
+    */
+    this.flagCache = this.restClient.bulkEvaluate(initialContext);
     this.startPolling();
   }
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -178,6 +178,7 @@ The initialization function is an ideal place for such logic.
 
 The `domain` the provider is registered under is also supplied, allowing the provider to scope domain-specific behavior, such as partitioning a persistent cache, so that multiple providers sharing the same storage do not collide.
 A `provider` instance is initialized only once, even when bound to multiple `domains`; in that case the `domain` supplied is the one under which it was first registered.
+A provider that maintains `domain`-specific state can instead declare itself `domain-scoped` (see [Requirement 2.4.3](#requirement-243)), in which case it is restricted to a single `domain` and this ambiguity does not arise.
 The default provider, which is not bound to a domain, is initialized without one.
 
 ```java
@@ -211,6 +212,14 @@ If a provider is unable to start up correctly, it should indicate abnormal execu
 If the error is irrecoverable (perhaps due to bad credentials or invalid configuration) the `PROVIDER_FATAL` error code should be used.
 
 see: [error codes](../types.md#error-code)
+
+#### Requirement 2.4.3
+
+> The `provider` **MAY** declare that it is `domain-scoped`, indicating that it maintains state specific to a single `domain`, such as a persistent cache, that cannot be shared across `domains`.
+
+Most providers are stateless with respect to their `domain` and can safely back multiple `domains` from a single instance.
+Providers that persist or cache `domain`-specific data need a stable, unambiguous `domain` to key that state on.
+By declaring itself `domain-scoped`, such a provider signals that the `API` must bind it to at most one `domain` (see [Requirement 1.1.8](./01-flag-evaluation.md#condition-118)), guaranteeing the `domain` supplied to `initialize` is the only one the instance will ever serve.
 
 ### 2.5. Shutdown
 
--- a/specification/sections/02-providers.md
+++ b/specification/sections/02-providers.md
@@ -170,7 +170,7 @@ class MyProvider implements Provider {
 
 #### Requirement 2.4.1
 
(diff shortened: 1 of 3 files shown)


specification/sections/02-providers.md
```diff
@@ -207,6 +207,26 @@ If the error is irrecoverable (perhaps due to bad credentials or invalid configu
 
 see: [error codes](../types.md#error-code)
 
+#### Requirement 2.4.3
+
+> The `provider` **MAY** define an initialization function which additionally accepts the `domain` it is registered under, if any.
```
jonathannorris: @toddbaert do you think we need to make this required?

MattIPv4:
Thank you :) this looks good to me

lukas-reining (approved):
Looks good to me!

specification/sections/02-providers.md
```diff
@@ -170,23 +170,26 @@ class MyProvider implements Provider {
...
-> The `provider` **MAY** define an initialization function which accepts the global `evaluation context` as an argument and performs initialization logic relevant to the provider.
+> The `provider` **MAY** define an initialization function which accepts the global `evaluation context` and the bound `domain` (if any) as arguments and performs initialization logic relevant to the provider.
 
 Many feature flag frameworks or SDKs require some initialization before they can be used.
 They might require the completion of an HTTP request, establishing persistent connections, or starting timers or worker threads.
 The initialization function is an ideal place for such logic.
 
-```java
+The `domain` the provider is registered under is also supplied, allowing the provider to scope domain-specific behavior, such as partitioning a persistent cache, so that multiple providers sharing the same storage do not collide.
+A `provider` instance is initialized only once, even when bound to multiple `domains`; in that case the `domain` supplied is the one under which it was first registered.
```
lukas-reining, line 180: I am a bit worried about this being confusing to new provider authors but I think it is worth it.

jonathannorris: We could potentially have an optional domain updated function on the provider if we think its necessary.

specification/sections/02-providers.md
```diff
@@ -170,19 +170,24 @@ class MyProvider implements Provider {
 
 #### Requirement 2.4.1
 
-> The `provider` **MAY** define an initialization function which accepts the global `evaluation context` as an argument and performs initialization logic relevant to the provider.
+> The `provider` **MAY** define an initialization function which accepts the global `evaluation context` and the bound `domain` (if any) as arguments and performs initialization logic relevant to the provider.
```
JamieSinn: is it optional to supply either, or if you supply one you must supply the other? it's not clear from the wording

specification/sections/02-providers.md
```diff
@@ -170,19 +170,24 @@ class MyProvider implements Provider {
...
 
 Many feature flag frameworks or SDKs require some initialization before they can be used.
 They might require the completion of an HTTP request, establishing persistent connections, or starting timers or worker threads.
 The initialization function is an ideal place for such logic.
 
+The `domain` the provider is registered under is also supplied, allowing the provider to scope domain-specific behavior, such as partitioning a persistent cache, so that multiple providers sharing the same storage do not collide.
+A `provider` instance is initialized only once, even when bound to multiple `domains`; in that case the `domain` supplied is the one under which it was first registered.
```
aepfli, line 180: How does this change, if by any chance somebody reassigns the first domain to a different provider? What is the expectation, should this change the behaviour? Theoretically, there could be a new provider with a different backend and still the same cache keys.

specification/sections/02-providers.md
toddbaert: I think this is mostly documentation. Most providers can probably ignore it - but if you need to maintain some cached state, it's important use to use this. I think that's the simple doc line we can add - this is a param for provider-maintained namepacing.

specification/sections/02-providers.md
toddbaert: > How does this change, if by any chance somebody reassigns the first domain to a different provider? What is the expectation, should this change the behaviour? Theoretically, there could be a new provider with a different backend and still the same cache keys.

Ya, that's a good point. 

Similarly, you can, hypothetically, use the same provider instance in 2 domains. That also brings up a similar question (which domain assignment "wins"). I don't have great answers to these at the moment.

That said, I since domains are the fundamental mechanism of isolation, I think we need to supply this information to providers so they can properly scope their caches, etc.

An obvious alternative is to provide a domain setter, but if that's separated from the init, I think it brings up a bunch of possible races we don't want to contend with.

jonathannorris:
Pushed an update adding a `domain-scoped` provider declaration, based on the #openfeature-js thread.

The point that a provider isn't 1:1 with a domain is fair, so rather than relying on providers to use the bound domain correctly, this makes it an enforced contract:

- `2.4.3`: a provider **MAY** declare itself `domain-scoped` when it holds per-domain state (like a persistent cache).
- `1.1.8` / `1.1.8.1`: when declared, the API **MUST NOT** bind that instance to more than one domain.

Stateless providers are unaffected and can still back multiple domains; only opt-in providers are constrained. This also resolves the open question about a shared instance only seeing its first domain.

OFREP cache key construction lives in the follow-up ADR: open-feature/protocol#80.

specification/sections/02-providers.md
jonathannorris: @toddbaert @JamieSinn I think we just keep 2.4.1 language-agnostic and word it so the `domain` is clearly optional/additive, leaving each SDK to introduce it backwards-compatibly in its own idiom (optional arg, overload, separate interface, whatever fits).

The `MAY` already allows this; I'd just tweak the wording so it's obvious the `domain` is optional to consume and existing providers aren't forced into a breaking change. Worth noting this will still be a breaking change in Go (its `Init(EvaluationContext) error` is a required interface method with no overloads) and likely a few other languages, where a separate optional interface with a fallback would be needed instead.

specification/sections/02-providers.md
jonathannorris: @aepfli @toddbaert I think both of these are covered now:

- **Reassigning a domain to a different provider with a different backend:** the OFREP cache key in the follow-up ADR (open-feature/protocol#80) includes the OFREP URL and auth, so a different backend produces a different key even on the same `domain`. The new provider won't read the old one's cached entries.
- **Same instance bound to two domains ("which wins"):** with the `domain-scoped` declaration (2.4.3) and 1.1.8.1, a provider that maintains per-domain state can only be bound to a single `domain`, and the API rejects a second binding. So there's no ambiguity for the providers that actually care. Stateless providers can still share across domains since they don't key anything on it.

jonathannorris:
Added `2.4.4`: a `domain-scoped` provider **MUST** accept the bound `domain` at initialization. The `domain-scoped` declaration is only meaningful if the provider actually consumes the `domain` to scope its state, so this makes that an explicit contract rather than leaving it implied.

I kept it as a provider-side `MUST` rather than an API-enforced check, since whether the API can detect a violation varies and we can't rely on surfacing it as an error consistently.

specification/sections/02-providers.md
jonathannorris: @toddbaert agreed, I think we keep this as a `MAY` rather than required; it's a documentation/namespacing param most providers can safely ignore. And the single-domain binding for `domain-scoped` providers (1.1.8.1) removes the need for a separate domain-setter I floated earlier: the `domain` is fixed at init for the providers that actually care, so we avoid the races a separate setter would introduce.

MattIPv4 (approved):

specification/sections/02-providers.md
toddbaert: Will re-review today, especially considering the other race condition changes.

JamieSinn (approved):

beeme1mr (approved):

toddbaert (approved):
I made one very small change to fix some wording that was a bit redundant, nothing substantive.

Issue #417, opened 2026-07-27
[Tracking] A cross-language conformance suite for OpenFeature providers


aepfli:
## The problem

OpenFeature's central promise is vendor neutrality: swapping providers must not change application
behaviour. Nothing currently verifies that.

Every provider tests differently. Some have thorough suites, some have almost none, and no two
agree on what "implements the provider contract" means. The consequences are real and already
visible:

- Providers disagree on whether a type mismatch returns the code default or throws.
- Providers disagree on whether an unreachable backend yields `ERROR` promptly, hangs, or throws
  out of provider registration.
- Providers disagree on which lifecycle events fire, and in what order, around a disconnect.
- Some providers silently coerce between numeric types instead of reporting `TYPE_MISMATCH`.

Every one of these is invisible until an application swaps providers in production. The spec
describes the correct behaviour in prose; there is no executable definition of it.

## The proposal

A **language-agnostic conformance suite for providers** — a set of Gherkin feature files, a
standardised backend control API, and a canonical flag set — that each language's TCK implementation
packages and runs against real providers.

We have built a working proof of concept in Java, using flagd as the first provider under test. It
finds real bugs already (see [Findings](#findings-from-the-poc)). This issue is the **cross-language
tracking issue**; it is opened for discussion before anything is merged into the spec repo.

### What lives in this repository

Three artifacts, which must travel together:

| Artifact | What it is |
|---|---|
| `gherkin/*.feature` | The canonical test cases — the executable definition of the provider contract |
| `openapi/control-api.yaml` | What a backend under test must expose so a TCK can drive it |
| `flags/canonical-flags.json` | The flag set those test cases assume |

They belong here rather than in any single language's repository because:

1. **They are the contract.** If every language's TCK writes its own scenarios, we get N different
   definitions of conformance and are no better off than today.
2. **Backend vendors implement the control API in their own language.** flagd's testbed is Go. It
   should not have to read a Java repository to learn what `POST /restart` must do.
3. **They cannot be split.** A feature file that evaluates `boolean-flag` is meaningless without the
   flag definition; a disconnect scenario is meaningless without the endpoint that produces the
   disconnect. Splitting them across repositories guarantees drift.

Each language's TCK then vendors this repo (git submodule or equivalent) and packages the files into
its own distribution format, so adopting a TCK never requires a consumer to check out a submodule of
their own.

### What lives in each language repository

The runner: step definitions, an abstract base test class owning the container lifecycle and control
[...]

PR #423, opened 2026-08-24
feat: add Appendix F, provider conformance (TCK)

aepfli:
> **Implements the spec-repo half of [#417](https://github.com/open-feature/spec/issues/417)**, the cross-language tracking issue for the provider conformance suite.
>
> **Related:** [#425](https://github.com/open-feature/spec/pull/425) is stacked on this and adds the machine-readable report. [#424](https://github.com/open-feature/spec/issues/424) is the design discussion on collecting and trusting reports. [#430](https://github.com/open-feature/spec/issues/430) and [#433](https://github.com/open-feature/spec/issues/433) are specification gaps this work uncovered. None is needed to review this PR.

## What and why

Adds **Appendix F: Provider Conformance (TCK)** — a language-agnostic conformance suite for OpenFeature *providers*, plus the artifacts it rests on. Marked **experimental and non-normative**; the scenario set is a representative subset.

OpenFeature's central promise is that swapping providers does not change application behaviour, and nothing verifies it. Every provider tests itself differently, against its own backend, with its own harness — so "implements the provider contract" is an unverified claim, and a difference between two providers gets found by the application that trips over it.

**Why not extend Appendix B:** Appendix B tests the **SDK** with an in-memory provider as a stand-in; this tests the **provider** against a real backend. It also needs more than Gherkin — the scenarios, the flag set they assume and the control API that produces outages are co-dependent, and an OpenAPI document has no home under "Gherkin Suites". Both appendices now cross-reference each other.

## Contents

| path | what it is |
|---|---|
| `appendix-f-provider-conformance.md` | the appendix: artifacts, control-API contract, capability vocabulary, what each language owns, extension and run-integrity rules |
| `assets/provider-tck/gherkin/` | six feature files — evaluation, errors, events, lifecycle, reason, metadata |
| `assets/provider-tck/flags/canonical-flags.json` | the flag set every scenario assumes |
| `assets/provider-tck/openapi/control-api.yaml` | what a backend under test must expose |
| `assets/provider-tck/go.mod`, `embed.go` | makes the directory a nested Go module exposing the artifacts as an `embed.FS`, because a Go module zip carries a submodule only as a gitlink. Inert for the other languages; precedent `open-feature/flagd-schemas` |

`assets/provider-tck/` rather than beside the SDK Gherkin, because `assets/gherkin/evaluation.feature` already exists and means something else. `.gitattributes` normalises to LF: four languages consume these byte for byte and copy them into published artifacts, so a Windows checkout must not package differently from a Linux one.

Requirements are stated as **normative blockquotes** (58), not numbered requirements — following Appendix A, since `specification.json` is built from numbered sections and no appendix contributes to it. Whether any of this should *become* numbered is a TSC decision, recorded as an open question.

## Design decisions worth reviewing

Each is argued in the appendix; this is the index.

- **Capabilities.** Optional behaviour is tagged; a provider declares what it supports and the rest are **skipped with a reason**, never passed. Tags with no scenarios yet are reserved and must not be declarable.
- **`@lifecycle` and `@events` are separate**, because a provider can initialise against its backend without emitting events, and vice versa.
- **`@numeric-coercion` is borrowed, not specified.** The rule comes from a flagd ADR, not from the specification. An earlier draft of this PR claimed the opposite and called withholding the tag "an admission of a known bug" — wrong, and corrected. Both halves are tested, so rejecting every float cannot pass. Gap: [#430](https://github.com/open-feature/spec/issues/430).
- **`@string-typing` and `@fully-typed-values` are two tags because one hid a bug.** Three providers over one Flagsmith backend: two report `TYPE_MISMATCH` for a boolean and an integer asked as strings, one stringifies them, and all three stringify the float and the structure. The latter is the backend's shape, the former a provider defect; one tag reported the defect as a permitted absence. Gap: [#433](https://github.com/open-feature/spec/issues/433).
- **No container restarts.** Outages are simulated through the control API, because orchestrators reassign host ports on restart and every provider pointed at the old one then fails looking flaky.
- **In-process control is a narrow carve-out** for providers with no backend. A provider *with* a backend must use the control API, or the suite passes while proving nothing.
- **An extension point is required, not optional** (raised in review). An extension may add questions but never replace one: it cannot occupy a canonical path, cannot satisfy a canonical scenario, and a run that skipped part of the canonical set must fail. That last rule was found by accident — a selector matching one scenario name produced a green suite and a well-formed report covering one scenario.
- **Run-integrity checks**, because some failures are invisible in results. Three fail the run; the one that matters is a scenario carrying a tag the implementation does not know, which all four reference implementations ignored — an unknown tag gates nothing, so its scenarios stay *mandatory* and a suite that has not learned a capability silently keeps demanding the old behaviour. Two further rules: a check that cannot be performed must fail rather than skip, and the revision check must be in force where the scenarios execute. Both came from one adoption producing plausible numbers against the previous revision's scenarios.

## Open questions

- **Evaluation context passthrough** — the targeting key is covered, no other attribute is; needs an echo operation on the control API.
- **Accessor width** — `@large-integers` covers up to 2^53 − 1; 32- versus 64-bit accessors are not modelled. Part of [#430](https://github.com/open-feature/spec/issues/430).
- **Finer-grained flag manipulation** would need new control-API endpoints.
- **Flag metadata** has one scenario and is blocked on a testbed fixture, not on the suite.
- **Hooks and caching** are not covered; `@caching` is reserved.

Answers **#417's Q1** (directory layout, forced rather than chosen) and **Q4** (numeric coercion stays a capability). Q2, Q3, Q5 and Q6 stay open; Q7 is split into #424.

## Implementations

| | suite | flagd | OFREP | Flagsmith |
|---|---|---|---|---|
| Java (reference) | [#1830](https://github.com/open-feature/java-sdk-contrib/pull/1830) | [#1847](https://github.com/open-feature/java-sdk-contrib/pull/1847) | [#1840](https://github.com/open-feature/java-sdk-contrib/pull/1840) | [#1849](https://github.com/open-feature/java-sdk-contrib/pull/1849) |
| Go | [#940](https://github.com/open-feature/go-sdk-contrib/pull/940) | [#941](https://github.com/open-feature/go-sdk-contrib/pull/941) | [#942](https://github.com/open-feature/go-sdk-contrib/pull/942) | [#959](https://github.com/open-feature/go-sdk-contrib/pull/959) |
| JavaScript | [#1606](https://github.com/open-feature/js-sdk-contrib/pull/1606) | [#1608](https://github.com/open-feature/js-sdk-contrib/pull/1608) | [#1612](https://github.com/open-feature/js-sdk-contrib/pull/1612) | [#1623](https://github.com/open-feature/js-sdk-contrib/pull/1623) |
| Python | [#409](https://github.com/open-feature/python-sdk-contrib/pull/409) | [#411](https://github.com/open-feature/python-sdk-contrib/pull/411) | [#414](https://github.com/open-feature/python-sdk-contrib/pull/414) | — |

Go consumes the assets through the nested module; the others use a submodule, verified to round-trip byte-identical.

## What the suite has found

- **flagd silently narrows a float flag to an integer**, in Go and Java, both resolvers, no error code — [flagd#1996](https://github.com/open-feature/flagd/issues/1996).
- **The Java SDK's `MultiProvider` swallows child provider events** — extends `EventProvider` but never subscribes to its children, so no child event reaches the client. [java-sdk#1882](https://github.com/open-feature/java-sdk/issues/1882).
- **The Python flagd provider widens a boolean to a float** — [python-sdk-contrib#417](https://github.com/open-feature/python-sdk-contrib/issues/417).
- **The JavaScript Flagsmith provider stringifies unconditionally**, where its number, boolean and object paths correctly report a mismatch. Visible only because three languages ran the same scenarios against the same backend.
- **Two specification gaps**, raised rather than decided here: [#430](https://github.com/open-feature/spec/issues/430), [#433](https://github.com/open-feature/spec/issues/433).
- **Two testbed gaps** — unseeded canonical flags ([flagd-testbed#392](https://github.com/open-feature/flagd-testbed/pull/392)) and a launchpad readiness race ([flagd-testbed#394](https://github.com/open-feature/flagd-testbed/pull/394)).


specification/assets/provider-tck/flags/canonical-flags.json
```diff
@@ -0,0 +1,82 @@
+{
```
toddbaert, line 1: I wonder if the canonical flags should also be gherkin? They could be a very large "given", I guess.

specification/assets/provider-tck/gherkin/evaluation.feature
```diff
@@ -0,0 +1,59 @@
...
+    Then the resolved details value should be "10"
+    And the error-code should be ""
+    And no exception should have been thrown
+
+  Scenario: A float flag resolves as a float
+    Given a Float-flag with key "float-flag" and a default value "0.1"
+    When the flag was evaluated with details
+    Then the resolved details value should be "0.5"
+    And the error-code should be ""
+    And no exception should have been thrown
```
toddbaert, line 103: What about integer precision? I think a lot of our SDKs support resolutions above 32 bit (though some like JS are practically capped at 2⁵³... it might be worth testing at least up to 2⁵³ -1

specification/assets/provider-tck/gherkin/lifecycle.feature
```diff
@@ -0,0 +1,40 @@
...
+  # without a lifecycle passes the readiness scenario below without demonstrating anything --
+  # a NoOpProvider passes it identically. @lifecycle asserts that the provider actually reaches
+  # its backend during initialisation and that the outcome is observable; a provider that merely
+  # emits events does not necessarily do that.
+  #
+  # The failure case matters more than it looks. A provider that blocks forever, or throws out
+  # of provider registration, takes the host application down with it — so the requirement is
+  # not merely that initialisation fails, but that it fails observably and promptly.
+
+  Scenario: A provider reaching its backend becomes ready
```
toddbaert:
```suggestion
  Scenario: A provider that successfully initializes becomes ready
```

I think I would keep this closer to the provider interface language. Some providers might not even need to reach a backend, but they might have some init that could fail.

Of course, I think most things tested by this suite would have some connection, but I think it's better to use this general language.

toddbaert:
I have to say, the Java impl is remarkably small and clean in Java. I think there's a lot of value here.

toddbaert:
I really like the idea, and the impl is clean (though TBH the issue and PR description is very wordy and maybe could be compressed and made less verbose) it took me a long time to get through :sweat_smile: .

I have 2 things I think I'd want to understand additionally:

- extension: I'd want as an explicit Appendix F requirement: the single-init lifecycle should be open for authors to add their own scenarios/steps that run in the same "phase" (one backend start/teardown), so vendors like flagd **extend** the TCK rather than maintaining a parallel harness for their provider-specific features (`fractional`, for example). I think it'd be good to make this outcome-normative and per-language, since it's nearly free in godog/pytest-bdd/jest but the current Java annotation-driven suite can't add features/glue without redeclaring the whole set; so if it's not required, we will end up with suites that can't be easily extended. WDYT?

- A report like you describe in https://github.com/open-feature/spec/issues/424... I made a comment [there](https://github.com/open-feature/spec/issues/424#issuecomment-5610029549), but I think we should consider bundling these (not sure).

specification/assets/provider-tck/flags/canonical-flags.json
aepfli: i can see the motivition behind it. but i am not sure it will be easier to read or understand or parse. Especially a json is easily transformed into another JSON for eg. CLI's to prefill databases etc. Not so sure about the gherkin for this purpose.

specification/assets/provider-tck/gherkin/evaluation.feature
aepfli: fully agree, but for now, i would love to focus on the method, rather than the tests. We should agree and finalize the basic process. Afterwards we can add more tests to ensure compatibility

aepfli:
I am in total favor for extensibility. This would even help our providers. We could try to define defaults in java, like a default resource path and a default step path, add those to the test, and have a simply way for the beginning. But also I am not sure, if this would be something which we should delay to the next version. For now, there is the idea, but do we need it immediately to ensure conformance. It sounds like a nice to have feature. A good iteration on the tck, i feel like it could bloat the current efforts. wdyt?

<details>
<summary>I measured what the Java classpath actually does here — the defaults idea works, with one catch</summary>

Four measurements against the real module:

| | |
| --- | --- |
| absent `@SelectClasspathResource` | hard discovery error |
| same directory name, two classpath roots | additive — both scanned |
| same directory **and filename** | **one silently wins** (`test-classes` beats the jar) |
| directory holding only a README | resolves fine |

The third is the catch. If the default resource path is just `features/`, a vendor adding `features/errors.feature` **silently replaces** the canonical file and the suite goes green having run theirs instead. So the default path has to be a separate one:

```
features/          canonical, TCK jar only
tck-extensions/    vendor features; ships with a README so the selector resolves
```

plus one extra glue package on the base suite (`openfeature.tck.extensions`) — an absent package is tolerated, so it costs nothing for anyone not using it.

Net cost: two lines in `AbstractProviderTckTest` and a README in the jar. A vendor then drops in a `.feature` file and a step class and writes **zero annotations**.

If it does land, I'd word it in Appendix F as an outcome rather than a mechanism — *an adopter can add scenarios and steps that run in the same backend phase without redeclaring the canonical set* — since it's pure convention in Java and Python, while Go and JS have no runtime scanning and need one registration callback.

Worth separating from the above either way: the suite should verify its canonical scenarios actually ran. That is not about extensibility — the same footgun already exists as `cucumber.filter.tags`, `-k` or `testPathIgnorePatterns`, and the file-shadowing result above makes it a concrete way to run fewer scenarios while reporting success.

</details>

// edit: flagd specifically is actually not a good reference example, as we are testing the evaluation engine seperately now. with really good tests. not sure if we want to have all of those tests executed during TCK evaluation. I see the benefit and the ease of running. But the more i think about, the TCK might be best to kept seperate as a verifiable unit, with whom nobody can mingle. Still torn, by both sides

aepfli:
Description updated to match the branch: `@numeric-coercion` is now described as genuinely optional, the Go module and the extension rules are listed, and #425 / #424 are linked.

specification/assets/provider-tck/gherkin/evaluation.feature
aepfli: added more cases on the side

Issue #424, opened 2026-08-24
Publishing provider conformance: a machine-readable report, and how to collect one from providers we do not host


aepfli:
> **Follow-up to [#417](https://github.com/open-feature/spec/issues/417)**, which asks in its open questions: *"Who owns conformance claims? Is there a badge, a report format, a listing on openfeature.dev? Or is passing simply its own reward for now?"* This issue is that question, split out so it can progress on its own timeline.
>
> **This is a design discussion, not a proposal to implement.**

## Status, 2026-09-11

The report *format* is no longer the substance of this issue. Following @toddbaert's comment below, [#425](https://github.com/open-feature/spec/pull/425) now defines only a small **envelope** and carries the results as **Cucumber Messages** rather than a format of our own. The sections below have been trimmed to what that leaves open: **collection, trust and expiry.** The minimized comments record how the format got there.

## The opportunity

The provider conformance suite ([Appendix F](https://github.com/open-feature/spec/pull/423)) has implementations in four languages, each of which already computes the data a public conformance report needs: the capability set a provider declares, and a per-scenario outcome. What is missing is a way to publish and collect reports. That would let openfeature.dev show, for any provider, what it actually supports — replacing "implements the provider contract" as an unverified claim with something a user can look up before choosing a provider.

## The awkward part

Not every provider lives in an SDK-contrib repo. Many are in vendor organisations, on their own release cadence, and OpenFeature cannot run their suites — the backends and credentials are not ours. So any design has to work for providers we do not host and cannot execute.

## Settled in #425

- **Comparability.** A report records the TCK implementation, its version and the spec revision its assets came from. The revision is a property of the TCK artifact: every implementation consumes the assets from this repository, via git submodule or, for Go, via the nested Go module, so the pin is always known. For the feature files the revision is corroborated by the Messages stream, which carries the executed source.
- **Outcomes.** Per-scenario outcomes come from Cucumber Messages. The envelope adds the declaration, which is an input to reading them, and a `notApplicable` set for capabilities a language cannot express — `@numeric-coercion` in JavaScript, which has no integer type. Collapsing that into "undeclared" would show every JS provider as missing something no JS provider can have.
- **Known deviations.** A provider can say "failed, known bug, tracked here", which is distinct from "chose not to". Three implementations had each invented a local shape for this.
- **One provider, several reports.** `configuration` keys them: flagd's RPC and in-process resolvers differ in whether they emit `PROVIDER_STALE`.

## Never compute a conformance percentage

[...]

PR #425, opened 2026-08-24
feat: add a schema for machine-readable provider conformance reports

aepfli:
Adds a JSON Schema for the **envelope** of one provider conformance run: what was tested, what the provider claims, and where the executed results are. The results themselves are **Cucumber Messages**, not a format this specification defines.

**Stacked on #423** (Appendix F), which defines the suite this reports on. Review that one first; this PR is only the report envelope.

Phase 1 of #424, which is the design discussion. Collection and trust — how OpenFeature gets reports from providers it does not host, and how much a self-published report should be believed — stay open on the issue and are deliberately not addressed here.

## What changed since the first draft

The first draft defined its own per-scenario result list, a four-value outcome enum, and a field identifying which Scenario Outline row an entry came from. All three already exist in [Cucumber Messages](https://github.com/cucumber/messages), which is maintained, cross-language, JSON-schema'd, and emitted natively by cucumber-jvm and godog. Defining them again would have created a second format to maintain and version, and two places for the same fact to disagree. That was the point of @toddbaert's comment on #424, and this PR now follows it.

The outline-row field was the clearest case for dropping ours: it was added days ago, and four implementations each reinvented it independently — one by reverse-engineering how its runner maps a pickle back to a table row. Messages has carried that identity as AST node ids all along.

Messages is also stronger than what it replaces. It carries the executed feature source, so a consumer can verify which questions were asked rather than trusting a recorded revision.

## What the envelope still has to carry

**The declaration.** Which capabilities the provider claims is an *input* to reading the results, not a summary of them. A skipped scenario says the question was not put; only the declaration says whether that is because the provider declines the capability. Given the declaration and a scenario's tags, the reason for a skip follows, so it no longer needs transporting per scenario. A capability that no executed scenario carries **must not** be declared, because it cannot produce a skip and listing it invites a reader to believe it was verified.

**Why a capability is absent, carried on the skip itself.** An earlier draft had a `declaration.notApplicable` list beside `declared`, to separate "the provider declines this" from "this language cannot express it" — `@numeric-coercion` is unsatisfiable in JavaScript, which has no integer type, and reporting that as a choice would show every JavaScript provider as missing something none of them can have. The list was dropped: a Messages skip carries its own reason, so the distinction travels with the scenario that was not run rather than in a second place that can disagree with it. The declaration is now just `declared`.

**The tested subject and its provenance.** No standard results format has a slot for "the provider under test": Messages records the runtime and OS, not the subject. `configuration` matters because one provider can produce several non-interchangeable reports — flagd's RPC and in-process resolvers differ in whether they emit `PROVIDER_STALE`.

**`specRevision`**, corroborated rather than merely asserted for the feature files, because the Messages stream carries the executed source. For the flag set and the control API it stands alone. An earlier draft carried a git tree hash of the asset directory to close that gap; it was dropped because every implementation either reads the assets from the submodule directly or regenerates and diffs its copy in CI, so drift is caught where it happens.

**`knownDeviations`**, distinguishing "chose not to implement an optional feature" from "has a known bug, tracked here". Three of the four implementations needed it before it existed: flagd narrows floats in two languages, Python marks one scenario as an expected failure against an SDK bug, and the JavaScript multi-provider flattens child error codes to `GENERAL`.

**No aggregate score.** No percentage, no "27/29". A single number invites comparing providers that declared different capability sets, which is exactly the comparison the tag vocabulary exists to prevent.

## Results are referenced, not inlined

A Messages stream carries the feature sources and is far larger than this envelope, and a consumer deciding whether it cares about a report should not have to fetch a whole run to find out. The envelope carries `format`, `formatVersion`, `location` and an optional digest. Two lossier formats are admitted for runners that cannot emit Messages yet, with what each loses stated: `cucumber-json` identifies outline rows only by line number, and `junit-xml` carries no tags at all.

One thing a consumer of a Messages stream needs to know: `testCaseFinished` carries no status. A scenario's outcome is the most severe `testStepResult` among its steps, hooks included. That is what makes a capability skip visible — a gate aborting in a before-hook produces a `SKIPPED` hook result outranking the `PASSED` steps it prevented.

## What is not here

Emitters that produce this envelope; the existing ones in Go, Python and JavaScript wrote the earlier draft format and will be updated once this shape is agreed.

**Open, and worth deciding before merge:** `provider.version` is optional while `sdk.version`, `tck.version` and `tck.specRevision` are all required. @dgenio [made the case on #424](https://github.com/open-feature/spec/issues/424#issuecomment-5610029549) for separating *subject supersession* (the tested release is no longer current) from *contract supersession* (the assets moved on), and as the schema stands only the second is decidable from a report — the one thing a report is evidence about is the one identifier that is optional. Requiring `provider.version` is a one-line change and I think it should be made; raising it here rather than doing it silently.

Also still open on #424: whether a report should be published inside the provider's released artifact.

Closes nothing; #424 stays open for collection and trust.


aepfli:
Rebased onto the current #423 head and rewrote the description for the Cucumber Messages envelope, which the old text predated.

